PR #3, opened 2016-10-28
Support TX PDO remote request + some PDO related bugfixes

tuwuhs:


tuwuhs:
I use informations in [this book](https://books.google.co.id/books?id=CstN61COZe4C&pg=PA421&lpg=PA421&dq=In+order+to+change+the+current+mapping+of+a+PDO,+the+PDO+must+first+be+disabled+by+writing+zero+to+Subentry+00h.&source=bl&ots=UqXzpPKPUm&sig=o_BOjiN_9LTjSfFwRUbRT2lf7Jk&hl=en&sa=X&ved=0ahUKEwjo2NnGy4TQAhXBpI8KHdI4ARIQ6AEIHTAA#v=onepage&q&f=false) as reference: "In order to change the current mapping of a PDO, the PDO must first be disabled by writing zero to Subentry 00h."

I also confirmed the behavior with a Copley Stepnet and Accelnet drives. When I tried to write the PDO mapping without setting the number of mapped objects to zero first, the drives return SDO abort message 0x06010000 (Unsupported access to an object).

christiansandberg:
You are absolutely right. We have done it wrong in our products. :sweat: I've found the corresponding instructions in DS-301:

> Subindex 0 determines the valid number of objects that have been mapped. For changing the PDO
mapping first the PDO has to be deleted, the sub-index 0 must be set to 0 (mapping is deactivated).
Then the objects can be remapped. When a new object is mapped by wrinting a subindex between 1
and 64, the device may check whether the object specified by index / sub-index exists. If the object
does not exist or the object cannot be mapped, the SDO transfer must be aborted with the Abort SDO
Transfer Service with one of the abort codes 0602 0000h or 0604 0041h.
After all objects are mapped subindex 0 is set to the valid number of mapped objects. Finally the PDO
will be created by writing to its communication parameter COB-ID. When subindex 0 is set to a value
>0 the device may validate the new PDO mapping before transmitting the response of the SDO
service.

PR #6, opened 2016-12-22
Allow sdo DOMAIN access with 4 or less bytes

wodz:
Previously transfer type (expedited or segmented) was determined
purely on transfer size.

### DIFF
--- a/canopen/sdo.py
+++ b/canopen/sdo.py
@@ -93,17 +93,17 @@ class SdoClient(collections.Mapping):
         """
         return ReadableStream(self, index, subindex).read()
 
-    def download(self, index, subindex, data_type, data):
+    def download(self, index, subindex, data, data_type = None):
         """May be called to manually make a write operation.
 
         :param int index:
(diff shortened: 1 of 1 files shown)


christiansandberg:
Thank you for the pull request!

Does the device you're using not support DOMAIN download using expedited transfer? My understanding was that expedited vs segmented transfers were independent of data type. Anyway I don't have a problem with having a special case, a segmented transfer should always work but is more inefficient. But I would prefer if you moved the data_type argument last and having it default to None so that the API is backwards compatible.

wodz:
I hit this communicating with firmware on LPC11C24 with rom canopen drivers - expedited transfer to DOMAIN object is rejected in this case. My understanding of the docs is that domain objects always use segmented transfer as DOMAIN objects size is implementation specific (and can be variable) but it is not explicitly stated anywhere. Re data_type at the end - no problem. I just need to figure out how to upload new version to the same pull request.

PR #15, opened 2017-02-18
canopen/network.py: automatic add node upon discovery

luminize:
Convenience addition for adding a newly discovered node who was unknown.
So when a device is freshly plugged in it will be known to the network.

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -159,6 +159,9 @@ class Network(collections.MutableMapping):
         If a custom interface is used, this function must be called for each
         11-bit standard message read from the CAN bus.
 
+        when a node is discovered who isn't in the nodes list then
+        it will be automatically added to the network
+
         :param int can_id:
             CAN-ID of the message (always 11-bit)
         :param bytearray data:
@@ -166,6 +169,13 @@ class Network(collections.MutableMapping):
         :param float timestamp:
             Timestamp of the message, preferably as a Unix timestamp
         """
+
+        node_id = can_id & 0x7F
+        if node_id in self.nodes:
+            pass
+        else:
+            self.add_node(int(node_id))
+
         if can_id in self.subscribers:
             callback = self.subscribers[can_id]
             callback(can_id, data, timestamp)


luminize:
>  I think this is better suited in the NodeScanner class, mostly because right now your code will try to add nodes even when just a request is made, like when scanning for nodes by requesting an SDO from all node IDs. You should also filter out node ID 0.

I understand.

> One other potential issue is that when you add a node without specifying an Object Dictionary, we will try to load the EDS using SDO. 

I don't fully understand that. you mean that when I add a node to the network without an eds file like `network.add_node(3)` it tries to download the information from the registers using SDO requests from the physical node?

> But since this code is run in the same CAN thread that receives SDO responses I'm afraid that it will deadlock until it times out. It's probably best to add nodes with an empty canopen.objectdictionary.ObjectDictionary() to not cause weird side-effects.

like `network.add_node(3, None)'?

Another question. I looked a bit further into the receiving of the messages since I previously used a non blocking poll/select for initiating the receiving the can bus messages.
As I read the python-can and canopen docs the `MessageListener` and `NodeScanner` receive callbacks originating from the Notifier object. The reading is done by a separate thread (on Notifier instantiation) calling Notifier.rx_thread which returns callbacks to the `MessageListener` as well as `NodeScanner` instances.
Is that in summary how the synchronizing of the `network.nodes[]` with the hardware starts?

> I also strongly believe that this should be a feature that you must opt in to use. The default should be to only handle the nodes that are intended.

sure

christiansandberg:
Index 0x1021 can, if supported, contain the EDS file for that node. So we try to load it from there if no other object dictionary has been given. If it fails, we create an empty ObjectDictionary() and use that. You should instantiate that directly and give it as second argument to add_node().

I think you understood the concept of how messages are received in a separate thread through python-can's Notifier and Listener classes.

luminize:
thanks, I'll re-open or make another PR if i decide to follow up on this. On second thought it's better to add nodes manually, specifying the .eds file so I can take full advantage of the available functionality.

PR #48, opened 2017-10-30
PDO bit mapping

Muschasgrazias:
get_data method is now able to understand bit mapping

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -108,7 +108,7 @@ def build_variable(eds, section, index, subindex=0):
         # This optional object is then placed in the eds under the section [A0] (start point, iterates for more)
         # The eds.get function gives us 0x00A0 now convert to String without hex representation and upper case
         # The sub2 part is then the section where the type parameter stands
-        var.data_type = int(eds.get("%Xsub1" % var.data_type, "DataType"), 0)
+        var.data_type = int(eds.get("%Xsub1" % var.data_type, "DefaultValue"), 0)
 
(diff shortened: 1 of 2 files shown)


Muschasgrazias:
Hm somehow the "import struct" at the start of the file is missing. Can you add this please?

Muschasgrazias:
@christiansandberg I can't add you to the Reviewers or Assigness

christiansandberg:
I've been meaning to have a look at it. I want to write some unit test to verify it.

christiansandberg:
It still won't work correctly with signed values. For example if one byte is 0xFF and it contains a 4-bit signed value, the read value will be 15 instead of -1.

Anyway, supporting signed values is probably too difficult so we can skip that. But setting values should be supported so if you can implement set_data as well then I can merge it.

Muschasgrazias:
I did a few basic test on signed 8 and 16 bit and it was working both for set and get data

christiansandberg:
Great job! I just did some minor refactoring.

Muschasgrazias:
Hey,
Shifting might be not faster than multiplication for small numbers check:
https://stackoverflow.com/questions/37053379/times-two-faster-than-bit-shift
In short:
>So when multiplying two integers where each fits in a 30-bit digit, this is done as a direct multiplication by the CPython interpreter, instead of working with the integers as arrays

christiansandberg:
Interesting. Does that apply for power of as well?

Muschasgrazias:
Good point, power has a few checks before actually executing multiplications. I would have to do benchmark tests, but in the end, shift will be faster

Issue #49, opened 2017-11-06
PDO COB-ID DefaultValue issue


jra-skov-dk:
Hi.

Thanks for this great python canopen implementation.
I am using the EDS import function of python canopen to autogenerate C code for a canopen stack.

I have encountered a small issue in this regard.
```

[1401sub1]
ParameterName=COB ID
ObjectType=0x7
DataType=0x0007
AccessType=rw
DefaultValue=$NODEID+0x300
PDOMapping=0
LowLimit=0x00000001
HighLimit=0xFFFFFFFF
```

For all PDO configurations the COB-ID Default value looks like above `DefaultValue=$NODEID+0x300` (generated by VectorEds)
But it seems the line `var.default = int(eds.get(section, "DefaultValue"), 0)` in eds.py cannot handle this `$NODEID+` syntax and simply sets it to zero.

If I remove `$NODEID+` in the eds I get a warnings from VectorEds that this differs from standard value.

Any suggestions on how to solve this?

My suggestion:
1. Extract DefaultValue as a string
2. Evaluate for `$NODEID+` and remove it
3. Convert to int and use as it is done now.

What do you think? (I don't think the `$NODEID+` is used for anything in the master, It is just an indication of behavior shown in the EDS file)

jra-skov-dk:
Just did a test:

Changing from this:
`var.default = int(eds.get(section, "DefaultValue"), 0)`
to this:
`var.default = int(eds.get(section, "DefaultValue").replace('$NODEID+',''), 0)`

fixes everything for me, all other DefaultValues stay the same but COB-ID is loaded correctly now.
I will make a pull request and you can evaluate if this has unintended side effects.

christiansandberg:
This is a known limitation (see #1). $NODEID should be replaced with the actual node ID and then evaluated. So if the node ID is 6, $NODEID+0x300 should be 0x306. I prepared for this by feeding the node_id to import_eds(). Could you try to implement that instead?

jra-skov-dk:
Ok, I see what you mean.

I will implement that instead. (lets continue the talk in https://github.com/christiansandberg/canopen/pull/50)

PR #50, opened 2017-11-06
Fixed issue with loading eds DefaultValues from COB-ID's

jra-skov-dk:
DefaultValues for COB-ID's are often prefixed with '$NODEID+'
and tools like VectorEds give a warning if it is not.

eds.py could not handle this prefix and would convert the value to zero.

Solution for https://github.com/christiansandberg/canopen/issues/49

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -44,7 +44,7 @@ def import_eds(source, node_id):
             object_type = int(eds.get(section, "ObjectType"), 0)
 
             if object_type == VAR:
-                var = build_variable(eds, section, index)
+                var = build_variable(eds, section, node_id, index)
                 od.add_object(var)
             elif object_type == ARR and eds.has_option(section, "CompactSubObj"):
                 arr = objectdictionary.Array(name, index)
@@ -52,7 +52,7 @@ def import_eds(source, node_id):
                     "Number of entries", index, 0)
                 last_subindex.data_type = objectdictionary.UNSIGNED8
                 arr.add_member(last_subindex)
-                arr.add_member(build_variable(eds, section, index, 1))
+                arr.add_member(build_variable(eds, section, node_id, index, 1))
                 od.add_object(arr)
             elif object_type == ARR:
                 arr = objectdictionary.Array(name, index)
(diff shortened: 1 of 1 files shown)


jra-skov-dk:
Ok, my change to node.py caused Travis to fail.

I added a change so if node is set to 0 when calling add_node() and it doesn't find anything in the EDS it would actually use the node_id = 0.

I am aware that node_id = 0 is an invalid configuration if you need to connect to a node.

For my system where I use canopen python to generate C code based on EDS file I really just needed it to return the 0x280 value when the EDS was '$NODEID+0x280'

I'm not 100% sure what to do now.

canopen/objectdictionary/eds.py
```diff
@@ -122,7 +122,13 @@ def build_variable(eds, section, index, subindex=0):
             pass
     if eds.has_option(section, "DefaultValue"):
         try:
-            var.default = int(eds.get(section, "DefaultValue"), 0)
+            default_value = eds.get(section, "DefaultValue")
+            #COB-ID can have a suffix of '$NODEID+' so replace this with node_id before converting
+            if '$NODEID+' in default_value:
+               var.default = int(default_value.replace('$NODEID+',''), 0) + node_id
```
christiansandberg: Try something like this:

```python
if '$NODEID' in default_value and node_id is not None:
    var.default = eval(default_value.replace('$NODEID', str(node_id)))
```

Then you can revert node.py.

jra-skov-dk: Good idea and that would be a good solution.

But the problem is that when you set node ID to 0 and it doesn't locate a Node ID in the EDS file, python stops on the line below my change in node.py:
`self.sdo = SdoClient(0x600 + self.id, 0x580 + self.id, object_dictionary)`
That you cannot do '+' on an integer and a 'None' type.

And since you have an actual travis test for this specific case I thought this was expected behavior.
Otherwise there is no failure when the system cannot find a node ID in the EDS file.

But, I honestly think what my use case is a bit special so I will revert the change to node.py.
The suggestion with `and node_id is not None:` should not be needed since python will do a raise before this issue if node_id is None.

jra-skov-dk: It seems `and node_id is not None:` was needed :)

christiansandberg:
Thanks! I checked and it is not required to support complex expressions, so we can assume it is always `$NODEID+integer`.

jra-skov-dk:
Likewise! python canopen is a joy to use for CANopen slave auto testing :)

PR #54, opened 2018-01-11
Add inhibit_time to PDO configuration options

benkfra:
PDO configuration now supports the optional field inhibit_time.
If set, this time is a minimum interval for PDO transmission. 
The value is defined as multiple of 100µs.

### DIFF
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -150,7 +150,7 @@ class Map(object):
         self.rtr_allowed = True
         #: Transmission type (0-255)
         self.trans_type = None
-        #: Inhibit Time (optional) (in 100µs)
+        #: Inhibit Time (optional) (in 100us)
         self.inhibit_time = None
         #: Event timer (optional) (in ms)
         self.event_timer = None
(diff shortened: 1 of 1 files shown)


canopen/pdo.py
```diff
@@ -150,7 +150,9 @@ def __init__(self, pdo_node, com_record, map_array):
         self.rtr_allowed = True
         #: Transmission type (0-255)
         self.trans_type = None
-        #: Event timer (in ms)
+        #: Inhibit Time (optional) (in 100µs)
```
christiansandberg: Write "us" instead to avoid using unicode characters.

canopen/pdo.py
```diff
@@ -276,6 +285,9 @@ def save(self):
         if self.trans_type is not None:
             logger.info("Setting transmission type to %d", self.trans_type)
             self.com_record[2].raw = self.trans_type
+        if self.inhibit_time is not None:
+            logger.info("Setting inhibit time to %d µs", (self.inhibit_time * 100))
```
christiansandberg: Same here.

benkfra:
Replaced the unicode Chars with ASCII Equivalent as suggested.
Btw. thanks for this great module - it has saved me a lot of testing time already.

PR #56, opened 2018-01-24
Extension to local-node 

fredrikhoyer:
I have extended the canopen slave functionality (local-node). The added functionality is enough to make a working canopen slave that is accepted by a canopen master based on the Port stack. An example of a simple working canopen simulator using this extended functionality can be found here: https://github.com/fredrikhoyer/canslave_sim

The main extensions that I have added are:
- Added a NmtSlave class with functionality to handle Heartbeat service, reception of NMT commands and handling of NMT state
- Extended Local Node to use the new NmtSlave class
- Extend the Sdo Server so that it to some extent handle block upload/download and request aborted
- Extended the EDS parsing so that it handles some of the real world eds files that I have

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -1,6 +1,7 @@
 import collections
 import logging
 import threading
+import struct
 
 try:
     import can
@@ -42,6 +43,7 @@ class Network(collections.MutableMapping):
         self.notifier = None
         self.nodes = {}
         self.subscribers = {}
+        self.nmt_cmd_subscribers = {}
         self.send_lock = threading.Lock()
         self.sync = SyncProducer(self)
         self.time = TimeProducer(self)
@@ -68,6 +70,23 @@ class Network(collections.MutableMapping):
     def unsubscribe(self, can_id):
         """Stop listening for message."""
         del self.subscribers[can_id]
+
+    def subscribe_nmt_cmd(self, node_id, callback):
+        """Listen for nmt commands to a specific node.
+
+        Only one callback can be used per Node ID.
+
+        :param int node_id:
+            The Node ID to listen for.
+        :param callback:
+            Function to call when message is received.
+        """
+        self.nmt_cmd_subscribers[node_id] = callback
+
+    def unsubscribe_nmt_cmd(self, node_id):
+        """Stop listening for nmt commands."""
+        del self.nmt_cmd_subscribers[node_id]
+
 
     def connect(self, *args, **kwargs):
         """Connect to CAN bus using python-can.
@@ -197,10 +216,19 @@ class Network(collections.MutableMapping):
         :param float timestamp:
             Timestamp of the message, preferably as a Unix timestamp
         """
-        if can_id in self.subscribers:
-            callback = self.subscribers[can_id]
-            callback(can_id, data, timestamp)
-        self.scanner.on_message_received(can_id)
+
+        # NMT commands is sent out with can_id = 0
+        if can_id == 0:
+            (cmd, node_id) = struct.unpack_from("<BB", data)
+
+            if node_id in self.nmt_cmd_subscribers:
+                callback = self.nmt_cmd_subscribers[node_id]
+                callback(node_id, data, timestamp)
+        else:
+            if can_id in self.subscribers:
+                callback = self.subscribers[can_id]
+                callback(can_id, data, timestamp)
+            self.scanner.on_message_received(can_id)
 
     def __getitem__(self, node_id):
         return self.nodes[node_id]
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -219,11 +219,17 @@ class Network(collections.MutableMapping):
 
         # NMT commands is sent out with can_id = 0
         if can_id == 0:
-            (cmd, node_id) = struct.unpack_from("<BB", data)
+            (_, node_id) = struct.unpack_from("<BB", data)
 
-            if node_id in self.nmt_cmd_subscribers:
+            # Broadcast has node-id = 0
+            if node_id == 0:
+                for subscriber_id in self.nmt_cmd_subscribers:
+                    callback = self.nmt_cmd_subscribers[subscriber_id]
+                    callback(data, timestamp)
+
+            elif node_id in self.nmt_cmd_subscribers:
                 callback = self.nmt_cmd_subscribers[node_id]
-                callback(node_id, data, timestamp)
+                callback(data, timestamp)
         else:
             if can_id in self.subscribers:
                 callback = self.subscribers[can_id]
--- a/canopen/nmt.py
+++ b/canopen/nmt.py
@@ -151,6 +151,13 @@ class NmtSlave(object):
         self._heartbeat_time_ms = 0
         self._local_node = local_node
 
+    def on_command(self, can_id, data, timestamp):
+        (cmd, node_id) = struct.unpack_from("<BB", data)
+
+        if node_id == self._id:
+            logger.info("Received command %d", cmd)
+            self.state = NMT_STATES[COMMAND_TO_STATE[cmd]]
+
     @property
     def state(self):
         """Attribute to get or set node's state as a string.
@@ -177,7 +184,7 @@ class NmtSlave(object):
             new_nmt_state = COMMAND_TO_STATE[NMT_COMMANDS[new_state]]
 
             logger.info("New NMT state %s, old state %s",
-                    NMT_STATES[new_nmt_state], NMT_STATES[self._state])
+                        NMT_STATES[new_nmt_state], NMT_STATES[self._state])
 
             # The heartbeat service should start on the transition
             # between INITIALIZING and PRE-OPERATIONAL state
@@ -216,7 +223,7 @@ class NmtSlave(object):
         while not stop_event.is_set():
             stop_event.wait(self._heartbeat_time_ms/1000)
             logger.debug("Sending heartbeat, NMT state is  %s", NMT_STATES[self._state])
-            
+
             try:
                 self.network.send_message(1792 + self._id, [self._state])
             except CanError as e:
--- a/canopen/nmt.py
+++ b/canopen/nmt.py
@@ -149,7 +149,7 @@ class NmtSlave(object):
         self._thread_stop = None
         self._heartbeat_time_ms = 0
 
-    def start_hearbeat(self, heartbeat_time_ms):
+    def start_heartbeat(self, heartbeat_time_ms):
         """Start the hearbeat service.
 
         :param int hearbeat_time
--- a/canopen/nmt.py
+++ b/canopen/nmt.py
@@ -59,6 +59,7 @@ class NmtMaster(object):
         with self.state_update:
             self.timestamp = timestamp
             new_state, = struct.unpack("B", data)
+            logger.info("Received heartbeat can-id %d, state is %d", can_id, new_state)
             if new_state == 0:
                 # Boot-up, will go to PRE-OPERATIONAL automatically
                 self._state = 127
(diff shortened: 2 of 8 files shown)


canopen/nmt.py
```diff
@@ -135,3 +137,99 @@ def wait_for_bootup(self, timeout=10):
...
+    def state(self, new_state):
+        if new_state in NMT_COMMANDS:
+            new_nmt_state = COMMAND_TO_STATE[NMT_COMMANDS[new_state]]
+
+            logger.info("New NMT state %s, old state %s",
+                        NMT_STATES[new_nmt_state], NMT_STATES[self._state])
+
+            # The heartbeat service should start on the transition
+            # between INITIALIZING and PRE-OPERATIONAL state
+            if self._state is 0 and new_nmt_state is 127:
```
christiansandberg: Values should be compared using ==.

canopen/node/local.py
```diff
@@ -11,16 +13,35 @@ def __init__(self, node_id, object_dictionary):
...
         self.sdo = SdoServer(0x600 + self.id, 0x580 + self.id, self)
+        self.nmt = NmtSlave(self.id, self)
+        self.add_callback(self._producer_hearbeat_time_callback)
 
     def associate_network(self, network):
         self.network = network
         self.sdo.network = network
+        self.nmt.network = network
         network.subscribe(self.sdo.rx_cobid, self.sdo.on_request)
+        network.subscribe(0, self.nmt.on_command)
```
christiansandberg: I only added support for having one callback per CAN-ID, so this will only work for one node. You need to change the callback implementation to allow multiple callbacks.

canopen/node/local.py
```diff
@@ -11,16 +13,35 @@ def __init__(self, node_id, object_dictionary):
...
     def add_callback(self, callback):
         self.callbacks.append(callback)
+
+    def _producer_hearbeat_time_callback(self, **kwargs):
+        if kwargs["index"] == 0x1017:
+            if kwargs["data"] is None:
+                # This is a read callback. If we return none
+                # the data will be read from local storage
+                return None
+            elif kwargs["data"] == 0:
```
christiansandberg: Should be checked against b"\x00\x00" since data is a `bytes` instance.

canopen/node/local.py
```diff
@@ -11,16 +13,35 @@ def __init__(self, node_id, object_dictionary):
...
+                self.nmt.stop_heartbeat()
+            else:
+                (hearbeat_time, ) = struct.unpack_from("<H", kwargs["data"])
+                self.nmt.start_heartbeat(hearbeat_time)
+            return 0x0201
```
christiansandberg: Is this necessary?

canopen/objectdictionary/eds.py
```diff
@@ -40,7 +41,7 @@ def import_eds(source, node_id):
         match = re.match(r"^[0-9A-Fa-f]{4}$", section)
         if match is not None:
             index = int(section, 16)
-            name = eds.get(section, "ParameterName")
+            name = eds.get(section, "ParameterName", raw=True)
```
christiansandberg: Doesn't work in Python 2.7 (see [Travis](https://travis-ci.org/christiansandberg/canopen/jobs/332983799)).

canopen/objectdictionary/eds.py
```diff
@@ -97,7 +111,7 @@ def import_from_node(node_id, network):
 
 
 def build_variable(eds, section, index, subindex=0):
-    name = eds.get(section, "ParameterName")
+    name = eds.get(section, "ParameterName", raw=True)
```
christiansandberg: Doesn't work in Python 2.7 (see Travis).

canopen/objectdictionary/eds.py
```diff
@@ -120,7 +134,20 @@ def build_variable(eds, section, index, subindex=0):
             pass
     if eds.has_option(section, "DefaultValue"):
         try:
-            var.default = int(eds.get(section, "DefaultValue"), 0)
+            if var.data_type is objectdictionary.VISIBLE_STRING:
```
christiansandberg: Check if `data_type` is in `DATA_TYPES` instead.

canopen/objectdictionary/eds.py
```diff
@@ -120,7 +134,20 @@ def build_variable(eds, section, index, subindex=0):
...
+                var.default = eds.get(section, "DefaultValue")
+            elif var.data_type is objectdictionary.REAL32:
```
christiansandberg: Check if `data_type` is in `FLOAT_TYPES` instead.

test/test_local.py
```diff
@@ -30,10 +30,33 @@ def tearDownClass(cls):
...
+        data = b"TEST DEVICE"
+        with self.assertRaises(canopen.SdoAbortedError) as context:
+            self.remote_node.sdo[0x1008].open('wb',
+                                              size=len(data),
+                                              block_transfer=True)
+        self.assertEqual(context.exception.code, 0x05040001)
+
+    def test_expedited_upload_default_value_visible_string(self):
+        device_name = self.remote_node.sdo["Manufacturer device name"].raw
+        self.assertEqual(device_name, b"TEST DEVICE")
```
christiansandberg: `raw` returns decoded string so remote the `b`.

test/test_local.py
```diff
@@ -49,6 +72,42 @@ def test_segmented_download(self):
...
+        # Line below is just so that we are sure the client have received the abort
+        # before we do the check
+        self.remote_node.sdo["Manufacturer device name"].raw = "Another cool device"
+        self.assertEqual(self.local_node.sdo.last_received_error, 0x05040003)
+
+    def test_start_remote_node(self):
+        self.remote_node.nmt.state = 'OPERATIONAL'
+        # Line below is just so that we are sure the client have received the command
+        # before we do the check
+        self.remote_node.sdo["Manufacturer device name"].raw = "Another cool device"
```
christiansandberg: Maybe a sleep will do here?

test/test_local.py
```diff
@@ -30,10 +30,33 @@ def tearDownClass(cls):
...
+        self.local_node.sdo[0x1400][1].raw = 0x99
+        vendor_id = self.remote_node.sdo[0x1400][1].raw
         self.assertEqual(vendor_id, 0x99)
 
+    def test_block_upload_switch_to_expedite_upload(self):
+        with self.assertRaises(canopen.SdoCommunicationError) as context:
+            self.remote_node.sdo[0x1008].open('r', block_transfer=True)
+        # We get this since the sdo client don't support the switch
+        # from block upload to expedite upload
+        self.assertTrue("Unexpected response 0x41" in context.exception)
```
christiansandberg: Convert `context.exception` to `str` first (see [Travis](https://travis-ci.org/christiansandberg/canopen/jobs/332983801)).

canopen/objectdictionary/eds.py
fredrikhoyer: I'm struggling to see why this do not work. I have tested this on my Mac with python 2.7.10. The ConfigParser package is:

pip show configparser
---
Metadata-Version: 1.1
Name: configparser
Version: 3.5.0

Also according to the ConfigParser documentation it seems to indicate that the raw options should be supported:
https://docs.python.org/2/library/configparser.html

Any suggestions?

canopen/node/local.py
fredrikhoyer: I assume you mean the return value?

I'm not sure what to actually return. The server that calls the callback looks like this:

        for callback in self._callbacks:
            status = callback(index=index, subindex=subindex, od=obj, data=data)
            if status:
                break

So I guess the callback that handles the index, subindex should return something other than None to indicate that the download was handled.

Any suggestion ?

canopen/objectdictionary/eds.py
christiansandberg: For Python 2.7 I used RawConfigParser so it may just be to switch to ConfigParser.

canopen/node/local.py
christiansandberg: I think the idea was that the function could return `True` if no other callbacks should be called for this write operation. Not sure if this is actually necessary though.

canopen/objectdictionary/eds.py
fredrikhoyer: Fixed

canopen/objectdictionary/eds.py
fredrikhoyer: Fixed

canopen/nmt.py
fredrikhoyer: Fixed

canopen/node/local.py
fredrikhoyer: Fixed. Rewrote the logic a bit.

canopen/node/local.py
fredrikhoyer: Fixed. Return True to indicate that the callback has been handled.

Maybe a future improvements could be to have seperate upload and download callback. Then it wouldn't be necessary in the callback to deduce if the callback is called due to a upload or a download. 

For a download callback we could return True/False depending of the callback was handled. For the upload callback we could return the value or none.

test/test_local.py
fredrikhoyer: Fixed.

test/test_local.py
fredrikhoyer: Fixed

canopen/objectdictionary/eds.py
fredrikhoyer: That explained it. I did change to use the RawConfigParser independent of the module that is imported.

test/test_local.py
fredrikhoyer: Fixed

canopen/node/local.py
fredrikhoyer: I changed the implementation to use specific nmt command subscribe function. This should then work with multiple nodes.

This is the last change that you requested. I.e. this is my last update, if you don't request more changes.

canopen/network.py
```diff
@@ -197,10 +216,19 @@ def notify(self, can_id, data, timestamp):
...
-        if can_id in self.subscribers:
-            callback = self.subscribers[can_id]
-            callback(can_id, data, timestamp)
-        self.scanner.on_message_received(can_id)
+
+        # NMT commands is sent out with can_id = 0
+        if can_id == 0:
+            (cmd, node_id) = struct.unpack_from("<BB", data)
+
+            if node_id in self.nmt_cmd_subscribers:
```
christiansandberg: You should also add a case for when `node_id` is 0, then all NMT callbacks should be called as it is a broadcast message.

canopen/nmt.py
```diff
@@ -135,3 +137,98 @@ def wait_for_bootup(self, timeout=10):
...
+        self._state = 0
+        self._timer_thread = None
+        self._thread_stop = None
+        self._heartbeat_time_ms = 0
+        self._local_node = local_node
+
+    def on_command(self, can_id, data, timestamp):
+        (cmd, _) = struct.unpack_from("<BB", data)
+
+        logger.info("Node %d received command %d",can_id, cmd)
```
christiansandberg: I think `can_id` should be replaced with `self._id`.

fredrikhoyer: Fixed

canopen/network.py
fredrikhoyer: Updated with Broadcast support

TobiasPleyer:
Hi Christian, hi Fredrik

I am happy to see both of you working on the server support. A couple of months ago I also uttered my interest in this extension. Back then I didn't have green light for the open source contribution. This has now been settled and I would be happy to support you guys.
@fredrikhoyer Please let me know what is your course for the next changes and how I can support you with it. For the next 1-2 weeks I should have a nice time budget to contribute.

Thanks for your help
Tobias

PR #60, opened 2018-02-07
PDO Support

TobiasPleyer:
WORK IN PROGRESS

First draft for the PDO support in canopen.

The current implementation differentiates between:

-  Local and remote nodes (remote nodes are not implented yet)
- Transmit PDOs (TPDOs) and Receive PDOs (RPDOs)

Most of the implementation work has been put into supporting TPDOs.

The key about TPDOs is that their payload is fed from the current procedural data of the node. That means the payload has to change for some messages when the data of the node changes. For efficiency reasons the data is cached and recalculated when the data changes. Detection of data change is achieved via a callback system.


christiansandberg:
I think you’re on the right track so keep going.

TobiasPleyer:
With the current implementation status I am now able to send and receive PDOs on my CAN bus. I think that is a satisfying milestone. From that point onwards we can slowly add new features and resolve TODOs as necessary, while already having code that can be "online".

christiansandberg:
I see that you want to change the existing API too. I can understand that it may be a bit unorthodox but I wanted it to be short and concise. The definition of “raw” being the unscaled integer is something I took from other CAN tools so that it would be familiar. 

There are many implementations relying on the existing API so unless you have a really good reason to change it I would prefer to keep it. Adding alternative interfaces is OK.

TobiasPleyer:
I can understand your concerns. The main reason I changed the status quo was that I was concerned for handling the constant data changes in a unified way. The current solution with the **data_store** was not totally stisfying, because it doesn't handle bit mapping, data type verification, access permissions and default values. I thought because most of this information is stored in the **Variable** class in the objectdicitonary sub-package it is best to keep everything there. I understood the object dicitonary of a CANopen node as **the** central data structure. Everything iscontrolled by changing the values of this data structure.One of my goals was to reduce the implementation to only one class definition of **Variable**, **Record** and **Array**.

But I can understand that you need to preserve **backwards compatibility** and **API stability**. I will try to revert parts of my changes.

Can you give me a little bit more detailed information what you dislike. Is the user level API the only thing youdon't want to change? E.g. could you live with only one definition of **Variable**, **Record** and **Array** or would you say this is already too much of a change?

Please let me know.

Cheers

christiansandberg:
Everything concerning the slave functionality like the SDO server is new so you are free to change that. My idea was to have mostly the same API between local and remote nodes so that an application can use both without knowing if it is real or simulated. But it’s not required.

But basically the existing tests (except for the local nodes tests) and API documentation should stay the same to assure that it doesn’t break backwards compatibility. The underlying mechanisms can definitely be changed. I don’t like the mess with the different OD classes either so if you find a way to simplify that it would be awesome!

TobiasPleyer:
Hi Christian,

what do you think about the newest changes?

christiansandberg:
Thank you! I'll try to look at it next week.

canopen/sdo/client.py
```diff
@@ -10,15 +11,140 @@
...
 from .exceptions import *
+from ..objectdictionary import datatypes, Bits
+from ..objectdictionary.exceptions import OdIndexError, OdSubIndexError
 
 
 logger = logging.getLogger(__name__)
 
 
-class SdoClient(SdoBase):
+class DataProxy(object):
```
christiansandberg: We lose the ability to use Records and Arrays as dictionaries. It should for example be possible to iterate over them. Note that the length of an Array when using SDO must be read from subindex 0 in contrast to Arrays in the Object Dictionary. That's one of the reasons I had multiple implementations of those classes since they are similar but not the same.

It should also be possible to distinguish between variables, records and arrays in a straight forward way. This is an example of how it was done before:

```python
for item in node.sdo:
    if isinstance(item, canopen.sdo.Variable):
        print('%s = %d' % (item.od.name, item.raw))
    elif isinstance(item, canopen.sdo.Array):
        print('%s (array):' % item.od.name)
        for array_item in item:
            print('    %d: %d' % (array_item.od.subindex, array_item.raw))
    elif isinstance(item, canopen.sdo.Record):
        print('%s (record):' % item.od.name)
        for record_item in item:
            print('    %s: %d' % (record_item.od.name, record_item.raw))

File: canopen/sdo/client.py [hunk 2]
```diff
@@ -36,9 +162,25 @@ def __init__(self, rx_cobid, tx_cobid, od):
...
-        SdoBase.__init__(self, rx_cobid, tx_cobid, od)
+        self.rx_cobid = rx_cobid
+        self.tx_cobid = tx_cobid
+        self.network = None
+        self.node = node
+        self.od = node.object_dictionary
         self.responses = queue.Queue()
 
+    def __getitem__(self, index):
+        return DataProxy(self, self.node.object_dictionary[index])
```
christiansandberg: Can't you just use `self.od[index]` here and remove `self.node`?

canopen/objectdictionary/__init__.py
```diff
@@ -92,8 +94,92 @@ def add_object(self, obj):
...
+    def phys(self, values):
+        if len(values) != len(self.subindices):
+            logger.error("Length of list with new values does not match "
+                         "sub-index count")
+            return
+        for index in self:
+            self.subindices[index].phys = values.pop(0)
+
+
+class Record(collections.MutableMapping, ValueStore):
```
christiansandberg: The ValueStore class adds a lot of attributes that doesn't make much sense to records and arrays.

canopen/pdo.py
```diff
@@ -15,48 +12,97 @@
...
+            for pdo in pdos.values():
+                pdo.setup()
+
+    def cleanup(self):
+        for pdos in (self.rx, self.tx):
+            for pdo in pdos.values():
+                pdo.cleanup()
+
+
+class RemotePdoNode(RemotePdoBase):
```
christiansandberg: Should also implement the `stop` method.

canopen/pdo.py
```diff
@@ -15,48 +12,97 @@
...
         raise KeyError("%s was not found in any map" % key)
 
     def __len__(self):
-        count = 0
-        for pdo_maps in (self.rx, self.tx):
-            for pdo_map in pdo_maps.values():
-                count += len(pdo_map)
-        return count
-
-    def read(self):
```
christiansandberg: Keep `read` and `save` methods for backwards compatibility.

canopen/network.py
```diff
@@ -195,6 +198,8 @@ def send_message(self, can_id, data, remote=False):
         :raises can.CanError:
             When the message fails to be transmitted
         """
+        logger.info("Sending message with COB ID 0x%X" % can_id)
+        logger.info("Data: {}".format(data))
```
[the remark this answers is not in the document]
TobiasPleyer: Ok

canopen/objectdictionary/__init__.py
```diff
@@ -92,8 +94,92 @@ def add_object(self, obj):
...
+                raise OdSubIndexError("Object at index {} does not have a "
+                                      "subindex {}".format(index, subindex))
+            obj = obj[subindex]
+        return obj
+
+
+class ValueStore(object):
+    """Provides a common interface to set and get all values out of object
+    types that hold a collection of values."""
+    subindices = {}
```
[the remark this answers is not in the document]
TobiasPleyer: They don't, because all classes inheriting from this class (**Record** and **Array**) define their private subindices dicitionary in their own `__init__` method, which takes precedence over the class variable. This works in this specialized case, but I agree that it is not directly obvious. If you prefer to make is explicit I can add an `__init__` method.

canopen/objectdictionary/eds.py
```diff
@@ -8,14 +7,15 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-VAR = 7
-ARR = 8
-RECORD = 9
+od_main_index_regex = re.compile(r"^[0-9A-Fa-f]{4}$")
+od_main_index_with_name_regex = re.compile(r"^([0-9A-Fa-f]{4})Name")
+od_sub_index_regex = re.compile(r"^([0-9A-Fa-f]{4})sub([0-9A-Fa-f]+)$")
```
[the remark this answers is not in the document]
TobiasPleyer: Cool, I wasn't aware of that! Thanks for pointing that out.

canopen/pdo.py
```diff
@@ -67,295 +113,239 @@ def export(self, filename):
...
+                routing_hex = map_entry[map_entry_index].raw
+                index = (routing_hex >> 16) & 0xFFFF
+                subindex = (routing_hex >> 8) & 0xFF
+                length = int((routing_hex & 0xFF) / 8)
+                # Only if there is data to map, i.e. length > 0
+                if length > 0:
+                    mapping.append((index, subindex, length))
+        return mapping
+
+    def on_config_change(self, index, subindex, data):
```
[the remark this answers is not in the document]
TobiasPleyer: I certainly should...
I was still unsure about the best way to do it. For now I installed the brute force method, which will run the whole setup of the PDO if a configuration change is detected.

canopen/pdo.py
```diff
@@ -67,295 +113,239 @@ def export(self, filename):
...
+            trans_type_variable.raw)
+        #: Inhibit Time (optional) (in 100us)
+        if 3 in com_entry.subindices:
+            self.inhibit_time = com_entry.subindices[3]
+        else:
+            self.inhibit_time = None
+        #: Event timer (optional) (in ms)
+        if 5 in com_entry.subindices:
+            self.event_timer = com_entry.subindices[5]
+            self.trans_type |= self.TT_CYCLIC
```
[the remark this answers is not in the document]
TobiasPleyer: Correct. This most likely comes from my lack of understanding of how the trigger mechanism works. The canopen spec does not describe the transmission type as a bit mask, it has defined values. But I read between the lines that it is possible to send on SYNC **and** transmit the PDO periodically **and** on value change. This is the behavior I need for my application and I didn't know a better way to implement it. If you have a better insight please explain me the real logic :)

canopen/pdo.py
```diff
@@ -67,295 +113,239 @@ def export(self, filename):
...
+                obj = self.pdo_node.node.get_object(index, subindex)
+                if self.on_data_change not in obj.traps:
+                    obj.traps.append(self.on_data_change)
+
+    def on_sync(self, can_id, data, timestamp):
+        """This is the callback method for when this PDO receives a SYNC
+        message. The SYNC triggers the data of the PDO to be sent."""
+        logger.info("Node %d: Transmitting on SYNC request" %
+                    self.pdo_node.node.id)
+        self.transmit_once()
```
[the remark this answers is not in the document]
TobiasPleyer: Ok that already clears one of my misunderstandings...

canopen/pdo.py
```diff
@@ -67,295 +113,239 @@ def export(self, filename):
...
+
+    def on_data_change(self, index, subindex, value):
+        """This is the callback method for when the internal data of the node
+        has changed."""
+        logger.info("Data change detected")
+        logger.info("Old message content {}".format(self.data))
+        # Parts of our data changed, the details don't matter, rebuild the data
+        self.data = self._build_data()
+        logger.info("New message content {}".format(self.data))
+        if self._task is not None:
```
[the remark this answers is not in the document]
TobiasPleyer: Good point!

canopen/sdo/client.py
```diff
@@ -172,6 +316,14 @@ def open(self, index, subindex=0, mode="rb", encoding="ascii",
...
             A file like object.
         """
+        try:
+            entry = self.od.get_object(index, subindex)
+        except OdIndexError:
+            raise SdoAbortedError(0x06020000)
+        except OdSubIndexError:
+            raise SdoAbortedError(0x06090011)
+        index = entry.index
+        subindex = entry.subindex
```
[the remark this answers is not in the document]
TobiasPleyer: Can you please elaborate a bit more? What should I remove? The method? The trick is that before the try block index and subindex could be strings, but once the entry is found querying its index and subindex value are guaranteed to be integers.

test/test_pdo.py
```diff
@@ -9,27 +9,28 @@
 class TestPDO(unittest.TestCase):
 
     def test_bit_mapping(self):
```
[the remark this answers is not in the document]
TobiasPleyer: I agree. This is the only test that I short circuited for 2 reasons:
* I anyways want to write a more elaborate PDO test once you give me your general OK
* This test case was written for the remote node PDO, and I really don't know if this use case makes any sense. I think remote node PDOs should be read only, because sending them would make us a local node and for everything else we have SDOs

canopen/sdo/client.py
TobiasPleyer: Ok I have no quick answer to this. I need to think a bit...

canopen/sdo/client.py
TobiasPleyer: Absolutely

canopen/objectdictionary/__init__.py
TobiasPleyer: It allows to set and get all contained data at once. If this is not desired I can remove it.

canopen/pdo.py
TobiasPleyer: The question is what the implication of `stop` would be? In my conception a remote PDO node is a silent listener, which does no sending at all, thus nothing is to stop and the base case is taken.

canopen/pdo.py
TobiasPleyer: Ok. I will fix this in a follow up commit

TobiasPleyer:
Thanks for your detailed feedback. I can understand your concerns. I will try to meet your requests as best as I can. This is the first fix up wave. But in general maybe not everything can be saved, because I think overall the new changes bring a lot of new logic with them, which cannot be taken 1:1 from older revisions. E.g. the PDO indexing logic has changed, because the 1,2,3,4 index case is only valid for the default PDO mappings. If you have more than 4 PDOs per node, the object dictionary indices can be have totally different values and non-consecutive.

Best regards,
Tobi

TobiasPleyer:
Please let me know what you think of the newest additions. Maintaining 100% backwards compatibility is not easy and it makes parts of the implementation look weird. Some things I definitely wouldn't write like that if backwards compatibility wasn't an issue. I am personally convinced that the new API increases readability and simplifies the class hierarchy, but I currently doubt that I can combine that with 100% backwards compatibility.

canopen/objectdictionary/__init__.py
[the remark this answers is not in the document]
christiansandberg: Right, in that case you can keep it.

canopen/pdo.py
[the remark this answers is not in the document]
christiansandberg: Don't think you can use it as a bit field. These are the meaning of the values:

trans_type

canopen/pdo.py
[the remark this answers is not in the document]
christiansandberg: After I wrote that I changed the task class and fixed it on that level instead. But it doesn't hurt to  do it here also.

canopen/sdo/client.py
christiansandberg: I could accept the pull request as it is, but I might do some changes of my own afterwards.

canopen/objectdictionary/__init__.py
christiansandberg: I forgot about that. Then you can keep it for now.

canopen/pdo.py
christiansandberg: The RemotePdoNode sends RPDOs to the node. The `stop` method stops the RPDOs.

christiansandberg:
Why do PDO indexes need to start at 0x1500 and 0x1900?

christiansandberg:
I'm not convinced that the proxy classes makes the hierarchy much simpler. In general I'd rather sacrifice complexity inside a library to make the external interface easier to use. What that means is of course a matter of taste.

As I mentioned in a comment above I can accept your changes now but I will have to think about if I want to change some details later on.

christiansandberg:
The dictionary access to the `node.pdo` object is new and not very well documented, but the idea was to access variables that way, not the maps. For example `node.pdo['Some variable'].raw` to mirror the way SDOs works. It also doesn't require you to know which map it happens to be in. To access the maps you must use `node.pdo.tx` and `node.pdo.rx` objects, e.g. `node.pdo.tx[1]['Some variable'].raw`.

You are right that just reading or writing values doesn't justify the dictionary like API. But you can do other things that can be done with dictionaries and I believe it makes the code shorter and more intuitive to a Python developer. Here are some examples that would otherwise require more lines of code and harder to remember API methods to accomplish:

```python
# Set a bit
node.sdo['Some Bitfield'].raw |= 0x100

# Increment a value
node.sdo['Some Value'].phys += 0.1

# How many items is in the array right now?
len(node.sdo['Some Array'])

# Reading all values of an array
for item in node.sdo['Some Array'].values():
    print(item.raw)

# Read a value if it exists in the object dictionary, if not use 0 as default
node.sdo.get('Some Value', 0)

# Use a reference for event shorter code
command_speed = node.sdo['Command Speed']
for speed in range(0, 5000, 500):
    command_speed.phys = speed
    time.sleep(5)
```

Efficiency and memory usage is of lowest priority. After all, we're using a scripting language and the CAN bus is the bottle neck anyway. You can save references to the SDO objects and re-use them if you like.

I hope I managed to motivate my decision to model the SDO operation using a dictionary like API, even if it makes the internals more complicated.

TobiasPleyer:
Ok I'm convinced. But where do I go from here?

christiansandberg:
Sorry for the sparse communication. Currently I'm a bit short on time for taking on larger tasks. Since there are so many backwards compatible breaking changes here I want to make sure I have time to go through and test everything in detail before I merge it.

Some tests would help speeding this up, especially for PDOs since there are basically none today. One possibility is to do like the SDOs where a RemoteNode communicates with a LocalNode. Test that it is possible to configure the PDOs and that data can be transferred using RPDO and TPDO correctly.

TobiasPleyer:
I understand. I am also a bit short lately. I'll write a few tests for the new PDO extension. Until then maybe you are happy to hear that your library is successfully used in another project ;)

TobiasPleyer:
Hi Christian,

I added 9 PDO test cases to the test suite. There is a lot more to test, but I hope these tests provide a good starting point. Let me know what you think about it, I can add more test cases if desired.

TobiasPleyer:
Hi Christian,

what's your opinion about the test cases?

PR #73, opened 2018-06-20
Add a configurable pause when sending SDOs.

acolomb:
Some slave controllers cannot handle messages coming in too fast.
When mapping a PDO for example, python-canopen sends multiple SDO
queries in direct succession.

This adds an attribute SdoClient.pause_before_send, which can be set
on a node's sdo object.  If non-zero, time.sleep() is inserted before
each send_message() call in send_request().

This probably should go into the documentation somewhere. Any suggestions how to include it there, @christiansandberg?

### DIFF
--- a/canopen/sdo.py
+++ b/canopen/sdo.py
@@ -61,6 +61,9 @@ class SdoClient(collections.Mapping):
     #: Max number of request retries before raising error
     MAX_RETRIES = 1
 
+    #: Seconds to wait before sending a request, for rate limiting
+    PAUSE_BEFORE_SEND = 0.0
+
     def __init__(self, rx_cobid, tx_cobid, od):
         """
         :param int rx_cobid:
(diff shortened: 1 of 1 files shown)


christiansandberg:
Perhaps you could put it as a class variable, like `RESPONSE_TIMEOUT`. I regret a little making them upper-case since they are meant to be changed if needed, but once set they are usually constant.

acolomb:
Well, most conformant devices don't need it and therefore it should be set on a per-object basis.  We can move it to a class variable, though, and override that by setting the attribute on the individual objects.  `self.ATTRIBUTE_NAME` then choses the object attribute over the class attribute.  I've successfully used that pattern with `MAX_RETRIES` and `RESPONSE_TIMEOUT`.

Would any Python implementation optimize the check away if it sees that the class variable is initialized to zero?

christiansandberg:
There are no actual constants in Python so it will work like any variable.

acolomb:
Okay, pushed but not tested yet.

What about the documentation, will the class constant docstring suffice?

Issue #34, opened 2017-06-26
CAN data from server not datatype of eds, results in type mismatch (or wrong data).


TinusScheppers:
Hi,

It seems that our CAN server (Curtis motor driver) does not reply with bytes in the datatype defined in the eds.

eg: motor_rpm is defined as Int16 in the eds, but we receive b'\xAA\xFF\x00'. This results in "Mismatch between expected and actual data size". If I change to a bigger int we get the unsigned value instead of the signed one (as expected).

How would I go about fixing this given that we cannot change what the CAN server sends, only what is defined in the eds and this drivers code?

I have also noticed that unpack cannot handle extra zero bytes (\x00), but unpack_from (without offset) ignores the extra (\x00) bytes and displays the value correctly.

```python
up = struct.unpack('h', b'\xAA\xFF')
print(up[0]) #successful

up1 = struct.unpack('h', b'\xAA\xFF\x00\x00')
print(up1[0]) #struct.error: unpack requires a bytes object of length 2

up2 = struct.unpack_from('h', b'\xAA\xFF\x00')
print(up2[0]) #successful
```

I will test using unpack_from instead of unpack in the canopen driver next week but I thought I would ask you for your opinion in the meanwhile.

Thanks in advance!

Kind regards
Tinus

christiansandberg:
If it's a bug in the CAN server and it is only on a limited number of variables, I would probably read the data and unpack it manually.

```python
data = node.sdo['Motor RPM'].data
motor_rpm, = struct.unpack_from('<h', data)
```

TinusScheppers:
This seems to be the easiest in the short run, requiring a bit of custom code for each datatype.

However, is it a design choice to use unpack instead of unpack_from? I understand that using unpack_from hides datatype errors, but given that you can't (always) control what the server sends, it might be better to also have a case where you would use unpack_from to avoid this error?

What is your opinion on this?

christiansandberg:
It's more that you shouldn't get weird side effects that are difficult to debug. If for example the EDS specifies INT16 while the server sends INT32. As long as the values are within INT16 range it will be fine, but when the values are bigger you will just get garbage values. It could suffice with a warning though but also that could be missed.

Also if you try to download values the server will likely not accept it either.

TinusScheppers:
Indeed, I will manually decode the data, which should suffice.

Thanks for the explanation!

acolomb:
Hi!

[...]

PR #74, opened 2018-06-20
Ignore extra data bytes when parsing variables.

acolomb:
Some controllers send all PDOs with DLC=8 or specify wrong sizes in
SDO responses.  This is a catch-all solution to silently discard
unexpected extra bytes.

This is my hacky solution in part for #34, because the used Curtis Instruments motor controllers don't set the message length correctly in all cases.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -285,6 +285,9 @@ class Variable(object):
             return data.decode("utf_16_le")
         elif self.data_type in self.STRUCT_TYPES:
             try:
+                needed_bytes = self.STRUCT_TYPES[self.data_type].size
+                if needed_bytes < len(data):
+                    data = bytearray(data)[0:needed_bytes]
                 value, = self.STRUCT_TYPES[self.data_type].unpack(data)
                 return value
             except struct.error:
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -287,7 +287,7 @@ class Variable(object):
             try:
                 needed_bytes = self.STRUCT_TYPES[self.data_type].size
                 if needed_bytes < len(data):
-                    data = bytearray(data)[0:needed_bytes]
+                    data = data[0:needed_bytes]
                 value, = self.STRUCT_TYPES[self.data_type].unpack(data)
                 return value
             except struct.error:


canopen/objectdictionary/__init__.py
```diff
@@ -285,6 +285,9 @@ def decode_raw(self, data):
             return data.decode("utf_16_le")
         elif self.data_type in self.STRUCT_TYPES:
             try:
+                needed_bytes = self.STRUCT_TYPES[self.data_type].size
+                if needed_bytes < len(data):
+                    data = bytearray(data)[0:needed_bytes]
```
christiansandberg: Does it need to be a `bytearray`?

canopen/objectdictionary/__init__.py
```diff
@@ -285,6 +285,9 @@ def decode_raw(self, data):
             return data.decode("utf_16_le")
         elif self.data_type in self.STRUCT_TYPES:
             try:
+                needed_bytes = self.STRUCT_TYPES[self.data_type].size
+                if needed_bytes < len(data):
```
christiansandberg: A warning should be logged so if there is a mismatch between Object Dictionary and actual implementation and the user has issues with faulty values he/she might get an idea where to start looking.

acolomb: Should that be a `logger.debug()` then, as it would appear for every received PDO, possibly affecting performance?

canopen/objectdictionary/__init__.py
acolomb: Not quite sure why I couldn't use `data[0:needed_bytes]`, but I think it had something to do with mutable types?

Anyway, `encode_raw()` uses `bytearray`, and so does [python-can](https://python-can.readthedocs.io/en/stable/message.html#can.Message.data).  What other type did you have in mind?

canopen/objectdictionary/__init__.py
christiansandberg: I think I changed my mind. It’s probably better to make it configurable with default to enforce the correct size. An informative exception can be raised which explains how to configure the node to accept larger data packets. What do you think?

canopen/objectdictionary/__init__.py
christiansandberg: The data argument will be either bytes or bytearray which are both supported by the struct module so it should work. It doesn’t have to be mutable.

christiansandberg:
By the way, could you try with the master? I did a change in SdoClient.upload() that could fix this already.

canopen/objectdictionary/__init__.py
acolomb: Okay, removed the conversion and now slicing `data` directly.

canopen/objectdictionary/__init__.py
acolomb: The idea sounds nice, but where in the class stack should this be implemented?  Naming suggestions? Something with `strict` in the name probably...

acolomb:
I'll try with master when back at the machine tomorrow.  You're probably talking about 2e83a69d7e51866439e28b8c89c116751384f2c0.

acolomb:
Verified that commit 2e83a69d7e51866439e28b8c89c116751384f2c0 fixes the problem for the SDO case on my setup.

Since this PR approaches the problem on a lower layer, are we sure that no other cases (PDO, etc.) need to be made robust against such failures?

* `pdo.Map.get_data()` at least uses `unpack_from()`, so should be safe.
* `emcy.EmcyConsumer.on_emcy()` uses `unpack()`, but its struct length is 8.  Could be problematic if the manufacturer-specific error code field is sent with less than 5 bytes.
* The `sdo` submodule has some other uses of `unpack()` [here](https://github.com/christiansandberg/canopen/blob/4e51fc5ba28f20043cd87810f52033dc3a02c50a/canopen/sdo.py#L406) and [here](https://github.com/christiansandberg/canopen/blob/4e51fc5ba28f20043cd87810f52033dc3a02c50a/canopen/sdo.py#L545).  Is that safe enough?

christiansandberg:
The other uses of unpack should be according to the standard. Curtis probably also conforms to the standard regarding SDO, they just take the lazy path by not specifying the data size it sends.

christiansandberg:
I’ll close this as well for now. If you experience other problems we can reopen it.

PR #76, opened 2018-06-20
Provide a way to inhibit COB ID subscription for PDOs.

### LINKED ISSUE #75: Subscribers with the same COB ID are overwritten internally.
Let me first describe the scenario where the problem occurs.  I'm working on a system architecture that is not strictly a master-slave model.  Some slaves need to communicate among themselves, with the master (python-canopen) node just observing the communicated PDO data and setting up the appropriate mappings beforehand.

So the master configures node A's TxPDO1 mapping and sets a callback function to receive the data.  Then it configures node B's RxPDO1 mapping, overriding its COB ID record entry so that node B receives the PDO created by node A.  See the following example:

```python3
import canopen

network = canopen.Network()
network.connect(channel='vcan0', bustype='socketcan')

nodeA = canopen.Node(1, 'canopen/test/sample.eds')
nodeB = canopen.Node(2, 'canopen/test/sample.eds')
network.add_node(nodeA)
network.add_node(nodeB)

pdo = nodeA.pdo.tx[1]
pdo.add_variable('UNSIGNED8 value')
pdo.add_callback(lambda message: print(message))
pdo.enabled = True
pdo.save()

pdo = nodeB.pdo.rx[1]
pdo.cob_id = nodeA.pdo.tx[1].cob_id
pdo.add_variable('UNSIGNED8 value')
pdo.enabled = True
pdo.save()
```

After the last `pdo.save()`, node A's callback will never get called, because `canopen.Network.subscribe()` sets `nodeB.pdo.rx[1].on_message()` as the subscriber, overwriting nodeA's callback.  See also the [function's comment](https://github.com/christiansandberg/canopen/blob/b26721f2f46ddb06c24ebff6d52140badbab73a8/canopen/network.py#L59).

I see two possible solutions:

1. Allow `canopen.Network` to have multiple subscribers per COB ID.  This is not really needed in most cases, though, because the application does not need to get two callbacks for essentially the same received PDO.

2. Provide a way to inhibit the subscription in the second `pdo.save()` call, and error out if the same COB ID is subscribed to multiple times.

I have working code for both approaches and will open a PR shortly for the second method, which I would prefer.  Other opinions?

acolomb:
Add a boolean attribute canopen.pdo.Map.subscribe which can be set to
False to avoid overwriting any previous subscriber for the same COB
ID.

If subscription is intended (default), throw an error in read() or
save() if a subscriber is already registered for the same COB ID.

Using add_callback() implicitly enables subscription, because it is
obviously necessary.

Documentation and tests are not included.  Fixes #75.

### DIFF
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -165,6 +165,8 @@ class Map(object):
         self.period = None
         self.callbacks = []
         self.receive_condition = threading.Condition()
+        #: Should this object subscribe to the COB ID
+        self.subscribe = True
         self.is_received = False
         self._task = None
 
@@ -232,6 +234,8 @@ class Map(object):
             :class:`~canopen.pdo.Map`.
         """
         self.callbacks.append(callback)
+        # Ensure message handler is registered for our COB ID when saving
+        self.subscribe = True
 
     def read(self):
         """Read PDO configuration for this map using SDO."""
@@ -268,7 +272,8 @@ class Map(object):
             size = value & 0xFF
             self.add_variable(index, subindex, size)
 
-        if self.enabled:
+        if self.enabled and self.subscribe:
+            assert self.cob_id not in self.pdo_node.network.subscribers
             self.pdo_node.network.subscribe(self.cob_id, self.on_message)
 
     def save(self):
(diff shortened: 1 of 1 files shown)


christiansandberg:
Issue fixed in #79 instead.

PR #78, opened 2018-06-20
Respect number-of-entries in PDO communication record.

acolomb:
When read()ing a node's PDO configuration and the first subindex of
the communication record (number of entries) is set to its minimum,
skip reading the optional entries.

Also add some logging for the number of entries to be retrieved for
communication and mapping parameters.

### DIFF
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -235,6 +235,8 @@ class Map(object):
 
     def read(self):
         """Read PDO configuration for this map using SDO."""
+        nof_entries = self.com_record[0].raw
+        logger.info("Fetching %d communication record entries", nof_entries)
         cob_id = self.com_record[1].raw
         self.cob_id = cob_id & 0x7FF
         logger.info("COB-ID is 0x%X", self.cob_id)
@@ -244,7 +246,7 @@ class Map(object):
         logger.info("RTR is %s", "allowed" if self.rtr_allowed else "not allowed")
         self.trans_type = self.com_record[2].raw
         logger.info("Transmission type is %d", self.trans_type)
-        if self.trans_type >= 254:
+        if self.trans_type >= 254 and nof_entries > 2:
             try:
                 self.inhibit_time = self.com_record[3].raw
             except (KeyError, SdoAbortedError) as e:
@@ -261,6 +263,7 @@ class Map(object):
 
         self.clear()
         nof_entries = self.map_array[0].raw
+        logger.info("Fetching %d mapping entries", nof_entries)
         for subindex in range(1, nof_entries + 1):
             value = self.map_array[subindex].raw
             index = value >> 16


christiansandberg:
What's the argument for this? Can the Object Dictionary contain these entries but the node itself does not? Otherwise it seems a bit unnecessary to make an extra SDO read when the optional values will just result in a KeyError if the node does not support it.

acolomb:
I'm not quite sure anymore about the reason for this change.  I remember one device (Curtis 1222 steering controller) would not answer to SDO requests for unknown objects at all, instead of responding with an error code.  That might have also been caused by SDO requests in quick succession (see #73). Its OD lists these objects at least.

Will test this again tomorrow, if my application works without the change.

However, from a correctness point of view, the subindex 00 specifies how many valid entries are in the record.  Even though additional entries are listed in the EDS file, the device could still provide less valid entries without violating the standard.  The first entry is classified as `const`, so its value could change during NMT state Initialization.  As I see it, the CANopen protocols are all about being able to discover stuff at runtime, so that's the value I would use.  The object dictionary is an abstract concept, not actually tied to an EDS file in CiA 301.

Adding this one SDO request in advance may save three failing requests with `SdoAbortedError` later.

acolomb:
Just finished testing.  Both of my motor controller types do work correctly when using `pdo.Map.read()` without this change.

This must have been from a time when I had not yet implemented the SDO rate limiting #73, so SDO abort transfer replies sometimes got lost.  That's why I kept getting an `SdoCommunicationError` instead of `SdoAbortTransfer`, which would have been handled gracefully.

christiansandberg:
Good to hear. Then I’ll close this for now since #73 seems to solve the real issue with the node not responding. If it would be just about saving a few milliseconds for Curtis users while adding a few milliseconds for everyone else then it doesn’t seem to be motivated.

I just want to take a moment and thank you for your exemplary pull requests! I’ll do some merging next week when I have access to my computer.

acolomb:
You're welcome! :-)  These changes have been sitting in my working copy for several months now, but only now I had the time to untangle them into separate and documented changesets.

Thank **you** for this great library, saved me a lot of work in spite of the Curtis incompatibilites.  Digging up these fixes and workarounds was a good chance to learn Python by example.

Are you going to merge the other stuff through the GitHub interface or just pull and fast-forward through my commits?  I'd rather see the former, as it's easier to keep track using `git branch --no-merged`.

christiansandberg:
Normally I prefer merging short lived branches. Since your branches are based on a quite old master (from January) and many, they will be a bit difficult to follow in a graph view. That's why I think it's best to rebase them in this case. Unless you want to rebase them yourself?

acolomb:
It's a matter of taste, but I prefer seeing the merge commits and the time line where the patches originated (correct, from January).  Especially PRs with several commits should have their own merge commit IMHO, but in any case it's a good way to have a pointer to the relevant discussion in the history. 

I try to keep everything merging cleanly, but if you like I will rebase to current master.

PR #80, opened 2018-06-21
Optionally specify subindex override in PDO add_variable()

acolomb:
If the keyword argument "subindex" is given and non-zero, a mapping
entry is generated for the given subindex instead of the one in the
object dictionary.

Although such mappings should not be valid, for some
devices it is necessary to include them when sending PDO mappings.
The one this was tested with has a fixed number of eight entries,
corresponding to each PDO data byte.  Mapping e.g. 16-bit variables is
done by entering the object index twice, with subindex 00 and 01,
respectively.  So by forcing the subindex=01, it creates a virtual OD
entry corresponding to the high byte of the 16-bit integer object.

This is obviously not convered by the standard and only needed for
broken devices.  Tested with AC motor controllers from Curtis
Instruments, Inc.

### DIFF
--- a/canopen/common.py
+++ b/canopen/common.py
@@ -9,6 +9,10 @@ class Variable(object):
 
     def __init__(self, od):
         self.od = od
+        #: Holds a local, overridable copy of the Object Index
+        self.index = od.index
+        #: Holds a local, overridable copy of the Object Subindex
+        self.subindex = od.subindex
 
     def get_data(self):
         raise NotImplementedError("Variable is not readable")
@@ -63,8 +67,8 @@ class Variable(object):
         """
         value = self.od.decode_raw(self.data)
         text = "Value of %s (0x%X:%d) is %r" % (
-            self.od.name, self.od.index,
-            self.od.subindex, value)
+            self.od.name, self.index,
+            self.subindex, value)
         if value in self.od.value_descriptions:
             text += " (%s)" % self.od.value_descriptions[value]
         logger.debug(text)
@@ -73,8 +77,8 @@ class Variable(object):
     @raw.setter
     def raw(self, value):
         logger.debug("Writing %s (0x%X:%d) = %r",
-                     self.od.name, self.od.index,
-                     self.od.subindex, value)
+                     self.od.name, self.index,
+                     self.subindex, value)
         self.data = self.od.encode_raw(value)
 
     @property
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -311,8 +311,8 @@ class Map(object):
             subindex = 1
             for var in self.map:
                 logger.info("Writing %s (0x%X:%d, %d bits) to PDO map",
-                            var.name, var.od.index, var.subindex, var.length)
-                self.map_array[subindex].raw = (var.od.index << 16 |
+                            var.name, var.index, var.subindex, var.length)
+                self.map_array[subindex].raw = (var.index << 16 |
                                                 var.subindex << 8 |
                                                 var.length)
                 subindex += 1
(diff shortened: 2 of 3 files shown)


canopen/pdo.py
```diff
@@ -398,11 +401,23 @@ def __init__(self, od):
         self.offset = None
         self.name = od.name
         self.length = len(od)
+        self._override_subindex = None
```
christiansandberg: Perhaps it would be better to just have `self.index` and `self.subindex` as normal attributes which are defaulted to `od.index` and `od.subindex` but are allowed to be overridden.

acolomb:
Rebased to current master.

canopen/pdo.py
acolomb: After looking around I find that a bit inconsistent with how SDOs are handled.  Access to the index and subindex is always done through the `od` attribute of `common.Variable`.

For better encapsulation, we could wrap all those accesses into properties of `common.Variable` though, and just override the getter and setter in `pdo.Variable`.  That makes those `var.od.index` references a little easier to read and should be about equal performance-wise.

christiansandberg: Yes, it would be good to be able to access index and subindex directly in the common.Variable class. It seems overkill to use properties though.

acolomb:
Okay, I moved the implementation to `common.Variable` and used simple attributes holding an overridable copy instead of the property solution.

Should we do something similar about the `pdo.Variable.name` attribute so that SDO gets the same treatment?

These newest commits are not tested yet, will have a chance for that tomorrow afternoon.

christiansandberg:
The name is tricky for Records and Arrays since we need to squash the name of the index and subindex into one name for PDOs.

acolomb:
But isn't that also the case for SDO variables?  I just wonder why this is only implemented for `pdo.Variable`.

christiansandberg:
To access variables named "Some Variable" in different groups using SDO:
```python
node.sdo['Some Record 1']['Some Variable'].raw
node.sdo['Some Record 2']['Some Variable'].raw
```

Both could be mapped to the same PDO, so we need to have unique names.
```python
node.pdo['Some Record 1.Some Variable'].raw
node.pdo['Some Record 2.Some Variable'].raw
```

christiansandberg:
Uhm, I don't know why that matters though... 😕 You can add `name` to `common.Variable` too of course.

acolomb:
Yes, it should work because SDO code does not use the `name` attribute in `__getitem__()`, but looks up the subindex name / number in the dictionary instead.  PDO on the other hand uses its `pdo.Variable.name` to find the correct variable mapping.

So I think it should be okay moving it to `common.Variable`, so the displayed value for the attribute is consistent.  The second syntax you described would still not work on SDOs, though.  That's probably good because it avoids ambiguities and the first syntax is much cleaner.

I will implement this in a second PR, so feel free to merge this one.

canopen/common.py
```diff
@@ -63,8 +67,8 @@ def raw(self):
         """
         value = self.od.decode_raw(self.data)
         text = "Value of %s (0x%X:%d) is %r" % (
-            self.od.name, self.od.index,
```
christiansandberg: Could you change to use self.name here and below?

acolomb: Sorry, I missed that in #82.  Should I push it there or just include the change here?

christiansandberg: I merged the other one so you can do it here while merging with master.

acolomb: Made a new PR #84.  I don't know how I could fix it in this PR, because you will do the merging and the `common.Variable.name` changes are not included in (nor related to) this branch.

acolomb:
I'll just merge master into this to fix the conflicts.

acolomb:
The CI error seems totally unrelated to this pull request, or what am I missing?  Master itself passed at that point in the timeline.

acolomb:
Thanks for accepting all of these changes and sorry for any merging confusion I might have caused.  Really nice working with you @christiansandberg!

PR #83, opened 2018-06-26
Skip PDO mappings with zero length when looking up by variable name.

acolomb:
Follow-up for commit 062c097a09c0a6de8927fc3ce184088029e4c2e9.  When
looking up variables by name through the PdoNode object instead of an
individual pdo.Map object, the zero-length variables should be skipped
just the same.

Also simplify the nested loops a bit by looping through a list
concatenation instead.

This was missed in #81.

### DIFF
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -32,7 +32,7 @@ class PdoNode(collections.Mapping):
                     yield var.name
 
     def __getitem__(self, key):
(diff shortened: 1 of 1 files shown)


acolomb:
Hm... I just noticed the nested for loops are used in most `PdoNode` methods.  Should I clean them all up?

PR #85, opened 2018-06-26
Simplify nested loops through PDO maps.

acolomb:
Use a list concatenation instead of another nested loop.

As mentioned in #83, this cleans up the other loops for consistency.


acolomb:
I'm very sorry for not testing this change thoroughly, but it does not work on Python 3.  The `dict.values()` method returns a view object instead of a list, so we'll have to revert all of this.

Sorry for the noise, PR coming.

PR #89, opened 2018-07-05
Add workaround for Curtis' broken PDO mapping entries.

acolomb:
Differring from the standard, Curtis Instruments mixed up the order of
index, subindex and length fields in their implementation of the PDO
mapping parameter records.

Introduce an attribute "curtis_hack" on the Node class to enable
interpretation of the wrong ordered entries in pdo.Map.read() and
save().

This is needed at least for the 123xE series AC motor controllers from Curtis Instruments.  Tested on the 1232E with OS version 31.

### DIFF
--- a/canopen/node.py
+++ b/canopen/node.py
@@ -27,7 +27,7 @@ class Node(object):
 
         self.id = node_id or self.object_dictionary.node_id
 
-        #: Curtis HACK; Enable workarounds for broken implementation
+        #: Enable WORKAROUND for reversed PDO mapping entries
         self.curtis_hack = False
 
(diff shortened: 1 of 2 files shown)


acolomb:
Feel free to suggest a different name for the attribute.  Something like `workaround_reversed_pdo_mappings` would be more generic, but also a bit verbose...

christiansandberg:
I assume trying to write an entry the normal way will not result in an abort code?

acolomb:
Yes I think it does return an abort code, but not sure from the top of my head what it said.  Something like `0x06020000: "Object does not exist"` IIRC.  I would not take that to mean "retry with a wrong protocol implementation", though.

PR #97, opened 2018-07-25
Generate default PDO COB-ID values for the predefined connection set.

acolomb:
PDOs created purely via the Python API cannot be .start()ed unless
their cob_id attribute was set.  This can by done explicitly, or by
calling either .save() or .read() to retreive the device's value
online via SDO.

When just using objects from the Predefined Connection Set, the COB-ID
is already known without an SDO exchange.  Introduce a new attribute
pdo.Map.predefined_cob_id which holds this default COB-ID value so the
user does not have to calculate it.  It gets populated when generating
the available PDO mappings from the object dictionary.

To use it, just set the cob_id attribute to the value of
predefined_cob_id, avoiding the SDO exchange during save().

### DIFF
--- a/canopen/pdo.py
+++ b/canopen/pdo.py
@@ -22,8 +22,8 @@ class PdoNode(collections.Mapping):
     def __init__(self, node):
         self.network = None
         self.node = node
-        self.rx = Maps(0x1400, 0x1600, self)
-        self.tx = Maps(0x1800, 0x1A00, self)
+        self.rx = Maps(0x1400, 0x1600, self, 0x200)
+        self.tx = Maps(0x1800, 0x1A00, self, 0x180)
 
     def __iter__(self):
         for pdo_maps in (self.rx, self.tx):
@@ -116,14 +116,18 @@ class PdoNode(collections.Mapping):
 class Maps(collections.Mapping):
     """A collection of transmit or receive maps."""
 
-    def __init__(self, com_offset, map_offset, pdo_node):
+    def __init__(self, com_offset, map_offset, pdo_node, cob_base=None):
         self.maps = {}
         for map_no in range(128):
             if com_offset + map_no in pdo_node.node.object_dictionary:
-                self.maps[map_no + 1] = Map(
+                new_map = Map(
                     pdo_node,
                     pdo_node.node.sdo[com_offset + map_no],
                     pdo_node.node.sdo[map_offset + map_no])
+                # Generate default COB-IDs for predefined connection set
+                if cob_base is not None and map_no < 4:
+                    new_map.predefined_cob_id = cob_base + map_no * 0x100 + pdo_node.node.id
+                self.maps[map_no + 1] = new_map
 
     def __getitem__(self, key):
         return self.maps[key]
@@ -146,6 +150,8 @@ class Map(object):
         self.enabled = False
         #: COB-ID for this PDO
         self.cob_id = None
+        #: Default COB-ID if this PDO is part of the pre-defined connection set
+        self.predefined_cob_id = None
         #: Is the remote transmit request (RTR) allowed for this PDO
         self.rtr_allowed = True
         #: Transmission type (0-255)


canopen/pdo.py
```diff
@@ -116,14 +116,18 @@ def stop(self):
...
         for map_no in range(128):
             if com_offset + map_no in pdo_node.node.object_dictionary:
-                self.maps[map_no + 1] = Map(
+                new_map = Map(
                     pdo_node,
                     pdo_node.node.sdo[com_offset + map_no],
                     pdo_node.node.sdo[map_offset + map_no])
+                # Generate default COB-IDs for predefined connection set
+                if cob_base is not None and map_no < 4:
+                    new_map.predefined_cob_id = cob_base + map_no * 0x100 + pdo_node.node.id
```
christiansandberg: I think you can set `cob_id` directly instead of having a separate attribute.

acolomb: That would break the logic in `read()` and `save()`, which will use the device's configured COB-ID from the record only if the attribute was still `None`.

I intentionally left this as a conscious decision for the API user, because the nodes need not be configured to use the predefined connection set.  The PDO configuration record's COB-ID entry may be set to anything really, especially if some nodes should cross-communicate, setting one's RxPDO COB to the other's TxPDO COB for example.

acolomb:
I've had a look at your `local-nodes` changes regarding predefined COB-IDs.  Frankly I think my solution cleaner and it will do the trick there as well, just need to find out how to make the locally defined PDOs use the `predefined_cob_id` attribute by default.  But for other nodes, we should keep reading the COB-ID from their OD online and only provide this value as a fallback if one wants to avoid the SDO exchange.

Any idea why the Travis build has failed?  I don't see any reasonable error messages there...

christiansandberg:
Hm, yeah I think you’re right. One alternative is to make cob_id a property which returns self._cob_id if set, else self.predefined_cob_id.

acolomb:
Making it a property would not help with the usage in `read()` and `save()`, though.  We basically need to decide whether a PDO has a COB-ID by default or not.  If a default is set, we have no way to decide whether the node's OD record value should be used.  Keep in mind that only four of the 128 possible PDOs are included in the predefined connection set, so whatever other solution for setting the default we implement, it will not affect the majority of possible PDOs.

christiansandberg:
No I know, but internally we can use self._cob_id to determine if it has been explicitly set or not.

acolomb:
I think it's not worth the trouble to decide in all places where the current `cob_id` attribute is used, whether to access the internal `_cob_id` or keep accessing the new property.  There are just too many possible use cases for the `pdo.Map` object.  At least I don't see a sensible and clear default behavior.

That's why I just added it as an informational attribute.  One can externally set any COB-ID on the object really, including the predefined ones.  My approach just tries to encapsulate the calculation `0x180 + map_no * 0x100 if map_no <= 3` in a place where the `map_no` is actually available.  Hence the API user does not need to perform the calculation, but just write `o.cob_id = o.predefined_cob_id`, which quite clearly states "use the default COB-ID, and when I call `save()`, tell that to the node as well".  For PDOs outside of the default 4 TX + 4 RX range, it becomes a no-op and behaves as before (reading the COB-ID from the node's configuration record).

acolomb:
So for the `local-nodes` branch, the line I quoted above should just be included for the generated PDOs mappings.

Which raises the question anyhow, where would a local node read its PDO configuration records from?  I don't think an EDS file can currently specify values for the `0x1800.01` record entry...

acolomb:
Ping @christiansandberg, do you think this can be merged, or would you like a different approach?

Did I make my reasoning clear enough?

christiansandberg:
I can merge this for now. I don’t have a clear idea right now how to handle it in the long run so that it is easy to remember.

The object dictionary can specify everything so I’m thinking that it could be used somehow, probably by explicitly loading it if wanted.

PR #100, opened 2018-08-08
implemented the node guarding mechanism in the NMT object

af-silva:


af-silva:
change some code and improve code quality (I think)

PR #101, opened 2018-08-08
implemented the node guarding mechanism in the NMT object 

af-silva:

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -176,7 +176,7 @@ class Network(collections.MutableMapping):
             self.bus.send(msg)
         self.check()
 
-    def send_periodic(self, can_id, data, period):
+    def send_periodic(self, can_id, data, period, remote=False):
         """Start sending a message periodically.
 
         :param int can_id:
@@ -190,7 +190,7 @@ class Network(collections.MutableMapping):
             An task object with a ``.stop()`` method to stop the transmission
         :rtype: canopen.network.PeriodicMessageTask
         """
-        return PeriodicMessageTask(can_id, data, period, self.bus)
+        return PeriodicMessageTask(can_id, data, period, self.bus, remote)
 
     def notify(self, can_id, data, timestamp):
         """Feed incoming message to this library.
@@ -247,7 +247,7 @@ class PeriodicMessageTask(object):
     CyclicSendTask
     """
 
-    def __init__(self, can_id, data, period, bus):
+    def __init__(self, can_id, data, period, bus, extended=False, remote=False):
         """
         :param int can_id:
             CAN-ID of the message (always 11-bit)
@@ -260,9 +260,9 @@ class PeriodicMessageTask(object):
         """
         self.bus = bus
         self.period = period
-        self.msg = can.Message(extended_id=False,
+        self.msg = can.Message(extended_id=extended,
                                arbitration_id=can_id,
-                               data=data)
+                               data=data, is_remote_frame=remote)
         self._task = None
         self._start()
 
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -185,6 +185,8 @@ class Network(collections.MutableMapping):
             Data to be transmitted (anything that can be converted to bytes)
         :param float period:
             Seconds between each message
+        :param bool remote:
+            indicates if the message frame is a remote request to the slave node
 
         :return:
             An task object with a ``.stop()`` method to stop the transmission
@@ -247,7 +249,7 @@ class PeriodicMessageTask(object):
     CyclicSendTask
     """
 
-    def __init__(self, can_id, data, period, bus, extended=False, remote=False):
+    def __init__(self, can_id, data, period, bus, remote=False):
         """
         :param int can_id:
             CAN-ID of the message (always 11-bit)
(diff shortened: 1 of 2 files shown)


canopen/network.py
```diff
@@ -176,7 +176,7 @@ def send_message(self, can_id, data, remote=False):
             self.bus.send(msg)
         self.check()
 
-    def send_periodic(self, can_id, data, period):
+    def send_periodic(self, can_id, data, period, remote=False):
```
christiansandberg: Add parameter to documentation.

canopen/network.py
```diff
@@ -247,7 +247,7 @@ class PeriodicMessageTask(object):
     CyclicSendTask
     """
 
-    def __init__(self, can_id, data, period, bus):
+    def __init__(self, can_id, data, period, bus, extended=False, remote=False):
```
christiansandberg: Do you need to use extended frames? I wasn't sure how many actually used that so I have assumed standard frames until someone needs extended. If it is needed than there are a lot of additional changes needed so it should probably done in a new pull request.

canopen/nmt.py
```diff
@@ -132,6 +133,14 @@ def wait_for_bootup(self, timeout=10):
             if self._state_received == 0:
                 break
 
+    def start_node_guarding(self, period):
+        """Starts the node guarding mechanism."""
```
christiansandberg: Document the period parameter in the docstring.

canopen/network.py
```diff
@@ -190,7 +190,7 @@ def send_periodic(self, can_id, data, period):
             An task object with a ``.stop()`` method to stop the transmission
         :rtype: canopen.network.PeriodicMessageTask
         """
-        return PeriodicMessageTask(can_id, data, period, self.bus)
+        return PeriodicMessageTask(can_id, data, period, self.bus, remote)
```
christiansandberg: The remote parameter will be passed as the extended parameter since it is positional. You should write `remote=remote`.

canopen/network.py
af-silva: Ok, I have removed it for now. Like you said if a need extended a will make the changes and we go from their.

canopen/network.py
af-silva: Done ;)

canopen/network.py
af-silva: Fixed, my bad. Lack of attention

af-silva:
I will focus my attention in communication between your library and hardware that implements the CANOpen protocol DS301 and DS402. 
I'm fowling your discussion with TobiasPleyer regarding some changes to the PDO's and SDO's objects. I'm trying some implementation of my own with base of your work and some inputs from Tobias that I have check in the pullrequest e made. At the moment I intend to use this library as a Master node to manage slaves (implemented directly in the hardware) so the SDO Server does not make sense to me right now but since this is an API for the CANOpen protocol I agree that it should check every use case.

You can check my progress on my fork of your repository and when I have more suggestions I will make more pull requests. 

For now I will try to make as little as possible changes to the existing API and the way it works, I like a lot the way the PDO are mapped and accessed but I think it should be created specialization classes for the RPDO and TPDO exposing a more simplified API to receive (callbacks) and transmit data.

Also the .dcf or .edf files used to map the ObjectDicitionary could be used not only to know what objects are available in the remote nodes but also load pre-existing configuration to the remote nodes using the SDO's at "boot" time of the master. I search the API and didn't find this feature so I will also implement this. 

I hope I can help you improve this library, I thanks a lot for sharing your work, it has been a great starting point. I have already implemented simple ros_node based on the ros_canopen API but in the last version update they change a lot of things so I'm seeking a more stable and in a way stand alone library o integrate with the ros framework (I will provide also a ROS interface that uses this AP to communicate with CANOpen remote nodes and configure or control them)

christiansandberg:
I'm glad that you want to contribute! My plan is to merge the local-nodes branch to the master branch now. It's probably safest to manually pick Tobias changes regarding PDOs (like having separate TPDO and RPDO classes).

Reading configuration from Object Dictionary is also something that would be great (briefly mentioned [here](https://github.com/christiansandberg/canopen/pull/97#issuecomment-409972117)).

af-silva:
I will look into the Object Dictionary first then, since I need it and I can contribute. Regarding the PDOs I will take a look again to your implementation in the branch local-nodes and test with my hardware (in this case I will only look into the remote nodes).

In my opinion I agree in separating the local and remote nodes, but in the case of the PDOs and looking at @TobiasPleyer implementation it seems solid, a bit confusing with the classes RemotePdoNode and LocalPdoNode but I will try to use that and go from there. I also think that the PDO implementation should follow what you did with the SDO implementation. 
-- pdo (pkg) 
   -- __init__.py
   -- base.py
   -- tpdo.py
   -- rpdo.py
   -- exceptions.py

It would represent a cleaner structure and follow a great deal of documentation out there (at least the ones that a read separate the object in this way, like the SDO client/server mechanism)

What do you think? 
O don't know if I explained myself in the best way xD 

Thanks again, ;)

christiansandberg:
All sound like good ideas! I've merged to master now so you can continue to work from there.

TobiasPleyer:
Hi Andre (@af-silva) 

I am happy to hear that you want to contribute! I am very open for your input. I am running my forked version since several months now in a very communication heavy project and I can report that it is working very stable.

The ultimate goal however should be to bring as much of it as possible mainline. My use case is mostly to mimic real devices on a bus, thus my work on the PDOs. If you have questions I would be happy to help.

Cheers
Tobias

PR #106, opened 2018-08-28
LSS fastscan is supported

ymkim92:


doc/lss.rst
```diff
@@ -23,7 +23,7 @@ Finally, you can switch to LSS waiting state.
 
 
 .. note::
-    Fastscan is not supported.
+    Fastscan is supported as of August 2018.
```
christiansandberg: You could change this to v0.8.0.

ymkim92: No rush. I will just wait.. Is there anything I should do more?

christiansandberg: Sorry, I just meant that it is better to specify the version the functionality was (will be) introduced than the date it was added to master. Otherwise I don't have any comments. I'm not familiar with LSS so I leave it up to you that the implementation is correct. :)

ymkim92: I am sorry. I agree with your idea and I fixed it. Thank you.

PR #109, opened 2018-09-06
enabled a more agnostic way to access mapped objects in the PDO

af-silva:
Now the mapped objects in the PDO can be access by three ways e.g.:
node.tpdo['Statusword']
node.tpdo['0x6041']
node.tpdo[0x6041]

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -30,9 +30,9 @@ class PdoBase(collections.Mapping):
         return iter(self.map)
 
     def __getitem__(self, key):
-        if isinstance(key, int) and key in list(set(list(range(0x1A00, 0x1A80)) + # By TPDO ID
-                                                    list(range(0x1600, 0x1680)) + # By RPDO ID
-                                                    list(range(0, 80)))):         # By PDO Index
+        if isinstance(key, int) and (0x1A00 < key < 0x1A80 or   # By TPDO ID
+                                     0x1600 < key < 0x1680 or   # By RPDO ID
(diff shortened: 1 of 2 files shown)


af-silva:
@christiansandberg The travis is failing but I don't know way...

christiansandberg:
You’re trying to add range() together but range() is an iterator in Python 3.

af-silva:
Done! thanks! If you have a different solution please say so ;)

canopen/pdo/base.py
```diff
@@ -30,13 +30,18 @@ def __iter__(self):
         return iter(self.map)
 
     def __getitem__(self, key):
-        if isinstance(key, int):
+        if isinstance(key, int) and key in list(set(list(range(0x1A00, 0x1A80)) + # By TPDO ID
+                                                    list(range(0x1600, 0x1680)) + # By RPDO ID
+                                                    list(range(0, 80)))):         # By PDO Index
```
christiansandberg: Hmm, maybe it's better to just check value ranges. It feels a little bit unnecessary to create a large list of valid values each time.

```python
if isinstance(key, int):
    if 0x1A00 < key < 0x1A80 or 0x1600 < key < 0x1680 or 0 < key < 80:
```

af-silva: Done.

af-silva:
Ok everything check out now. If you can, verify the code and if you wish I think the API can be released as a version 1.0 or at least 0.9 (I don't know the working state of the local nodes implementation, so ....).

Thanks again!

PR #111, opened 2018-09-11
fix issue in the load_configuration() preventing one to set objects that were not records or arrays

af-silva:
The previous implementation only allow one to set composed objects, this is objects with sub-index's. Now we can set all the supported objects loaded by the EDS file

### DIFF
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -89,27 +89,46 @@ class RemoteNode(BaseNode):
         """
         self.sdo.download(0x1011, subindex, b"load")
 
+    def __load_configuration_helper(self, index, subindex, name, value):
+        """Helper function to send SDOs to the remote node
+        :param index: Object index
+        :param subindex: Object sub-index (if it does not exist e should be None)
+        :param name: Object name
+        :param value: Value to set in the object
+        """
+        try:
+            if subindex is not None:
+                logger.info(str('SDO [{index}][{subindex}]: {name}: {value}'.format(
+                    index=format(index, '#04x'),
+                    subindex=subindex,
+                    name=name,
+                    value=format(value, '#04x'))))
+                self.sdo[index][subindex].raw = value
+            else:
+                self.sdo[index].raw = value
+                logger.info(str('SDO [{index}]: {name}: {value}'.format(
+                    index=format(index, '#04x'),
+                    name=name,
+                    value=format(value, '#04x'))))
+        except canopen.SdoCommunicationError as e:
+            logger.info(str(e))
+        except canopen.SdoAbortedError as e:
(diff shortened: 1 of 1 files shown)


canopen/node/remote.py
```diff
@@ -89,27 +89,46 @@ def restore(self, subindex=1):
...
+        :param name: Object name
+        :param value: Value to set in the object
+        """
+        try:
+            if subindex is not None:
+                logger.info(str('SDO [{index}][{subindex}]: {name}: {value}'.format(
+                    index=format(index, '#04x'),
+                    subindex=subindex,
+                    name=name,
+                    value=format(value, '#04x'))))
```
christiansandberg: You could include the format of the hex in the first string: `logger.info('SDO [{index:#06x}][{subindex}]: {name}: {value:#06x}' ...`.

canopen/node/remote.py
```diff
@@ -89,27 +89,46 @@ def restore(self, subindex=1):
...
-                        except canopen.SdoAbortedError as e:
-                            # WORKAROUND for broken implementations: the SDO is set but the error
-                            # "Attempt to write a read-only object" is raised any way.
-                            if e.code != 0x06010002:
-                                # Abort codes other than "Attempt to write a read-only object"
-                                # should still be reported.
-                                print('[ERROR SETTING object {0}:{1}]  {2}'.format(subobj.index, subobj.subindex, str(e)))
-                                logger.info(str(e))
-                                raise
+                    if isinstance(subobj, Variable) and (subobj.access_type in ('rw', 'rww')) and (subobj.value is not None):
```
christiansandberg: I've added the `.writable` attribute to Variable as a higher level API than `.access_type`.

canopen/node/remote.py
```diff
@@ -89,27 +89,46 @@ def restore(self, subindex=1):
...
+        except canopen.SdoCommunicationError as e:
+            logger.info(str(e))
+        except canopen.SdoAbortedError as e:
+            # WORKAROUND for broken implementations: the SDO is set but the error
+            # "Attempt to write a read-only object" is raised any way.
+            if e.code != 0x06010002:
+                # Abort codes other than "Attempt to write a read-only object"
+                # should still be reported.
+                logger.warning('[ERROR SETTING object {0}:{1}]  {2}'.format(index, subindex, str(e)))
+                logger.info(str(e))
```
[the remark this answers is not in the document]
af-silva: You are right, removed!

canopen/node/remote.py
af-silva: Ok, done!

canopen/node/remote.py
af-silva: Ok, it's more clean this way. Done!

PR #112, opened 2018-09-14
Version 2 of the original implementation of the new DS402 Profile

af-silva:
Working: 
- Homing
- Fault Reset
- Profiled Position
- Profiled Velocity
- DS402 State Machine implementation 
- Auto config based on the configured PDOs (can be improved)

Working for the control of real world remote nodes.

### DIFF
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -377,9 +377,8 @@ class BaseNode402(RemoteNode):
         this map object is then used for changing the
         :param mapobject: :class: `canopen.objectdictionary.Variable`
         """
-
-        for map in mapobject:
-            self.tpdo_values[map.index] = map.raw
+        for obj in mapobject:
+            self.tpdo_values[obj.index] = obj.raw
 
     @property
     def statusword(self):
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -391,7 +391,6 @@ class BaseNode402(RemoteNode):
         except KeyError:
             raise KeyError('The object 0x6041 (Statusword) is not configured in this device.')
 
-    @classmethod
     @statusword.setter
     def statusword(self, value):
         raise RuntimeError('This property has no setter. The value {0} will be discarded'.format(value))
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -391,6 +391,7 @@ class BaseNode402(RemoteNode):
         except KeyError:
             raise KeyError('The object 0x6041 (Statusword) is not configured in this device.')
 
+    @classmethod
     @statusword.setter
     def statusword(self, value):
         raise RuntimeError('This property has no setter. The value {0} will be discarded'.format(value))
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -406,7 +406,7 @@ class BaseNode402(RemoteNode):
         """
         if 0x6040 in self.rpdo_pointers:
             self.rpdo_pointers[0x6040].raw = value
-            self.rpdo_pointers[0x6040].msg.transmit()
+            self.rpdo_pointers[0x6040].pdo_parent.transmit()
         else:
             self.sdo[0x6040].raw = value
 
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -232,7 +232,7 @@ class BaseNode402(RemoteNode):
             for obj in rpdo:
                 logger.debug('Configured RPDO: {0}'.format(obj.index))
                 if obj.index not in self.rpdo_pointers:
-                    self.rpdo_pointers[obj.index] = rpdo
+                    self.rpdo_pointers[obj.index] = obj
 
         # Check if the Controlword is configured
         if 0x6040 not in self.rpdo_pointers:
@@ -405,8 +405,8 @@ class BaseNode402(RemoteNode):
         :param int value: State value to send in the message
         """
         if 0x6040 in self.rpdo_pointers:
-            self.rpdo_pointers[0x6040][0x6040].raw = value
-            self.rpdo_pointers[0x6040].transmit()
+            self.rpdo_pointers[0x6040].raw = value
+            self.rpdo_pointers[0x6040].msg.transmit()
         else:
             self.sdo[0x6040].raw = value
 
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -393,7 +393,7 @@ class BaseNode402(RemoteNode):
 
     @statusword.setter
     def statusword(self, value):
-        raise RuntimeError('This property has no setter.')
+        raise RuntimeError('This property has no setter. The value {0} will be discarded'.format(value))
 
     @property
     def controlword(self):
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -86,6 +86,16 @@ class State402(object):
         ('FAULT', 'SWITCH ON DISABLED'):                  CW_SWITCH_ON_DISABLED,  # transition 15
     }
 
+    @staticmethod
+    def next_state_for_enabling(_from):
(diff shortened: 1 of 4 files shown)


af-silva:
Hi @christiansandberg I hold like to hear your opinion regarding the DS402 implementation. I have two versions done with some changes between them. I'm working and using the second to control 2 motors and everything is fine, I can handle everything from this API. I have a wrapper done in ROS around this implementation, but for now I can't release the code. Thanks for all your work. Best regards! ;)

christiansandberg:
Do you mean we should close version 1? Fix the unittests (on Python 3) and PyLint first. I don’t know much about DS402 so I can’t comment much about the actual implementation.

af-silva:
> Do you mean we should close version 1? 

I think so, the API is very solid for real applications. In my opinion for now we could improve the Documentation and add more examples to help developers quickstart their development, when a have a little time I can help with this.

> Fix the unittests (on Python 3) and PyLint first

Regarding the base implementation ? I think I have solved every one, or do you have more suggestions. 
The DS402 implementation I think is solid, can be improved (but everything can be improved), and I'm using it to control with success two motors and until now everything works as expected and far better then a previous implementation that I have in c++ based on the ROSCanopen API. Some functionality are "missing" from the base implementation because it can be very specific depending on the application (e.g., the handling of EMCY messages). Another thing that I did (it can be a wrong approach but it works very nice), I only use one callback (since they are run synchronously) to grab all the configured TPDOs (I assume only the default 4, this is one think it can be improved) and saved the values into dictionary's were the keys are the mapped objects. This is done for the TPDOs and RPDOs, were the RPDOs I save the PDO object enabling me to call the transmite. 

I'm open to suggestions in improving this implementation or add more features, for now I have implemented what I needed for my application with the time frame that I have.

Thanks.

af-silva:
@christiansandberg Two things I would appreciate your input:
**1st**, if you cloud take a look at the ``p402.py`` file, more specific at the code block of the _Controlword_:
```
if 0x6040 in self.rpdo_pointers:
            self.rpdo_pointers[0x6040][0x6040].raw = value
            self.rpdo_pointers[0x6040].transmit()
        else:
            self.sdo[0x6040].raw = value
```
I stored a "pointer" to the PDO object that holds the Controlword so I could call the transmit method after set the value. The "problem" here is that I have to store the PDO instead of the var, what do you think of adding a **"parent"** variable on the ``Variable`` instance that holds the parent PDO of that variable (passed as attribute when creating the object) ?? This way I could access the PDO trough the ``Variable`` (parent) instance.

**2nd**, is the usage of the callback ```on_TPDOs_update_callback()``` correct? In your example you use a ``@staticmethod`` declaration, and for a time I used too, but now I use a class function and everything works the same... what it's the advantage of using the static method instead of a class one? What is the best approach? (or I'm doing something wrong the things work all the same xD )

Thanks

af-silva:
Hi @luminize, if you are still working with this I would like your opinion too. 
Thanks in advance!

christiansandberg:
Sure, you can add a parent attribute member. There is a `.msg` attribute, maybe it is what you're looking for?

You can use class methods as callbacks.

af-silva:
Ok, you were right, what I needed was the .msg attribute but the name is not very intuitive... what do you think in changing the attribute ``.msg`` to ``.parent``?? Even looking at the code, the usage is not very clear and no documentation is provided in the class (at least I didn't find it).

The code is working as expected and is more simplified, it's working.

Thanks.

christiansandberg:
Sure, as you say it wasn’t documented and .msg is not a very good name.

canopen/profiles/p402.py
```diff
@@ -45,59 +201,218 @@ class Node402(RemoteNode):
...
+        """Returns the last read value of the Statusword (0x6041) from the device.
+        :raise ValueError: The Object 0x6041 (Statusword) is not configured in this device.
+        """
+        try:
+            return self.tpdo_values[0x6041]
+        except KeyError:
+            raise KeyError('The object 0x6041 (Statusword) is not configured in this device.')
+
+    @statusword.setter
+    def statusword(self, value):
```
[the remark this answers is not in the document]
af-silva: This error is annoying but i'm not seeing a way to remove it, the setter can't have the  ``@classmethod`` decorator so... 
My intent is to throw the exception if setting a statusword. I set the statusword as a property because I think is a good design choice.

Everything works.

@christiansandberg if you want, you can cherry pick the commit for the renamed ``.msg`` to ``.pdo_parent`` or if you agree this is a good approach for the ds402 profile you can accept the PR. If anyone have suggestion (you included) or bugs to report I'm glad to solve them.

Thanks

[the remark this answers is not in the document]
christiansandberg: I think you can remove the methods that don’t do anything. Then the user should simply get an exception that the attribute does not exist or is not settable.

canopen/profiles/p402.py
```diff
@@ -45,59 +201,214 @@ class Node402(RemoteNode):
...
+        #
+        for rpdo in self.rpdo.values():
+            for obj in rpdo:
+                logger.debug('Configured RPDO: {0}'.format(obj.index))
+                if obj.index not in self.rpdo_pointers:
+                    self.rpdo_pointers[obj.index] = obj
+
+        # Check if the Controlword is configured
+        if 0x6040 not in self.rpdo_pointers:
+            raise ValueError('Controlword not configured in the PDOs of this node, using SDOs to set Controlword')
```
comkieffer: Should this really be an error ? Wouldn't it be better to simply log a warning? 

The `ValueError` implies that the function failed but the setter for the control word checks whether or not the control word is in the `rpdo_pointers` which seems to indicate that this should not be an error.

canopen/profiles/p402.py
```diff
@@ -45,59 +201,214 @@ class Node402(RemoteNode):
...
 
+    def reset_from_fault(self):
+        """Reset node from fault and set it to Operation Enable state
+        """
+        if self.state == 'FAULT':
+            # particular case, it resets the Fault Reset bit (rising edge 0 -> 1)
+            self.controlword = State402.CW_DISABLE_VOLTAGE
+            timeout = time.time() + 0.4  # 400 milliseconds
+            # Check if the Fault Reset bit is still = 1
+            while self.statusword & State402.SW_MASK['FAULT'][0] == State402.SW_MASK['FAULT'][1]:
```
comkieffer: Parens would definitely make the condition more understandable here

canopen/profiles/p402.py
```diff
@@ -1,39 +1,195 @@
...
+        'VELOCITY'                    : 0x2,
+        'PROFILED VELOCITY'           : 0x4,
+        'PROFILED TORQUE'             : 0x8,
+        'HOMING'                      : 0x20,
+        'INTERPOLATED POSITION'       : 0x40,
+        'CYCLIC SYNCHRONOUS POSITION' : 0x80,
+        'CYCLIC SYNCHRONOUS VELOCITY' : 0x100,
+        'CYCLIC SYNCHRONOUS TORQUE'   : 0x200,
+        'OPEN LOOP SCALAR MODE'       : 0x10000,
+        'OPEN LOOP VECTOR MODE'       : 0x20000
```
comkieffer: I don't think that non-standard modes should be here. Every driver has their own stack of non standard nodes which aren't compatible with each other and this is only going to cause problems down the line. There should be a mechanism to extend the supported modes with vendor specific values.

canopen/profiles/p402.py
af-silva: I think you are right, we could change that to an error. Also if the the Controlword (0x6040) is not configured in the PDO it will try to use the SDO's (more slowly) to set the Controlword. I will make a push request to change this.

canopen/profiles/p402.py
af-silva: I also agree. At the moment I put the modes, but to enable the correct usage of this library we should only support generic modes, and all drive specific modes should be enable through extending the class.

PR #149, opened 2019-03-14
Fix for handling actual bits (length=1) in objectdictionary STRUCT_TYPES

Sandyman:
The BOOLEAN value was handled by using `struct.Struct("?")`. Unfortunately, this treats the data as an actual boolean, which breaks the functionality in `pdo/base.py`. By using "B" in stead of "?", we treat the data as an unsigned char (a byte), and the logic in `pdo/base.py` works without any further changes. 

I've added a few lines to the `sample.eds` and `test_pdo.py` files for testing purposes.


christiansandberg:
I kind of like having BOOLEANs represented as bool if possible. I tried to fix it in f6a35e33d2e765e450b06e2bfaf581d25655f392 if you could try it out.

PR #158, opened 2019-06-13
Added on_command method in NmtSlave which calls update_heartbeat 

rasmuspeders1:
This is to properly reflect NmtSlave state when receiving commands from NMT Master.

I was testing the LocalNode functionality and couldn't figure out why, when sending command to to slave to go to operation from NMT Master, the slave state was not correctly picked up on the Master side.

I believe this pull request could be the solution, but I might have misunderstood how this is supposed to work.


rasmuspeders1:
I understand the the slave functionality is in early stages. Let me know if you need anything else for this to be merged.

PR #160, opened 2019-08-27
Fix SDO communication error printout

Qbicz:
Before this change, raising an exception would fail with
TypeError: not enough arguments for format string


christiansandberg:
Thanks a lot!

PR #162, opened 2019-09-24
Show bit fields of Variables that are added to a Map.

Sandyman:
I encountered a situation where it was useful to see which bits are used inside PDOs. Added a few lines of code to do just that.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -415,8 +415,13 @@ class Map(object):
             if length is not None:
                 # Custom bit length
                 var.length = length
-            logger.info("Adding %s (0x%X:%d, %d bits) to PDO map",
-                        var.name, var.index, var.subindex, var.length)
+            # We want to see the bit fields within the PDO
+            offset = var.offset
+            byte_offset, bit_offset = divmod(offset, 8)
+            sb = byte_offset * 8 + bit_offset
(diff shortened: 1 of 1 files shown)


canopen/pdo/base.py
```diff
@@ -415,8 +415,13 @@ def add_variable(self, index, subindex=0, length=None):
...
                 # Custom bit length
                 var.length = length
-            logger.info("Adding %s (0x%X:%d, %d bits) to PDO map",
-                        var.name, var.index, var.subindex, var.length)
+            # We want to see the bit fields within the PDO
+            offset = var.offset
+            byte_offset, bit_offset = divmod(offset, 8)
+            sb = byte_offset * 8 + bit_offset
+            eb = sb + var.length - 1
+            logger.info("Adding %s (0x%X:%d, %d bits, %d:%d) to PDO map",
```
christiansandberg: Maybe it could be clearer what the values mean. If we log start and end bits then we don't need to have length also.

christiansandberg: Maybe `"Adding %s (0x%X:%d) at bits %d - %d to PDO map"`?

Sandyman: Sounds good.

Sandyman: Done. :)

Issue #169, opened 2019-11-14
DS402 State Transition 'QUICK STOP ACTIVE' to 'OPERATION ENABLED'


gsulc:
The DS 402 specification indicates the state transition 16, 'QUICK STOP ACTIVE' to 'OPERATION ENABLED', is optional. Some devices do not support it. This software library is the perfect place to handle this: try the transition and if it doesn't work, fall through to the next-best thing: go to 'SWITCH ON DISABLED' and then 'OPERATION ENABLED'. Bonus points: remember this for the node and don't try it again the next time. I can take this one.

I monkey-patched the method below, and it works fine:
```python
def enable_operation(self):
    state = self.state

    if state == 'OPERATION ENABLED':
        return
    elif state == 'QUICK STOP ACTIVE':
        self.state = 'SWITCH ON DISABLED'
        self.state = 'OPERATION ENABLED'
    else:
        self.state = 'OPERATION ENABLED'

canopen.BaseNode402.enable_operation = enable_operation
```

af-silva:
Hi @gsulc nice to see you hacking the code,
This issue is related to #167, could you please check the code snippet I posted there and try it please, and validate?

Thanks!

gsulc:
This did not work. As it is a different case/transition. I've had limited time to try things out, but I'm working on something. I should have something at leas viable tomorrow.

gsulc:
Okay, here is the first pass that works:
```python
    __init__(blahblahbblah):
        #...
        self.use_state_transition_16 = True
        #...

    @state.setter
    def state(self, new_state):
        """ Defines the state for the DS402 state machine
        :param string new_state: Target state
        :param int timeout:
        :raise RuntimeError: Occurs when the time defined to change the state is reached
        :raise TypeError: Occurs when trying to execute a ilegal transition in the sate machine
        """
        t_to_new_state = time.time() + 8  # 800 milliseconds tiemout
        while self.state != new_state:
            try:
                if new_state == 'OPERATION ENABLED':
                    if not self.use_state_transition_16 and self.state == 'QUICK STOP ACTIVE':
                        next_state = 'SWITCH ON DISABLED'
                    else:
                        next_state = State402.next_state_for_enabling(self.state)
                else:
                    next_state = new_state
                
                self.controlword = State402.TRANSITIONTABLE[
                        (self.state, next_state)]
                
                # timeout of 400 milliseconds to try set the next state
                t_to_next_state = time.time() + 0.4
                while self.state != next_state:
[...]

Issue #167, opened 2019-11-07
BaseNode402 Transition from 'OPERATION ENABLED' to 'QUICK STOP ACTIVE'


JulianS-Uni:
Hi there,

I have a problem when I do a transition form the state 'OPERATION ENABLED' to 'QUICK STOP ACTIVE'. I always get an `ValueError: 'Illegal transition from SWITCH ON DISABLED  to QUICK STOP ACTIVE'`. Other transitions are working fine, and besides the error, the motor is doing the transition just fine. I came to the conclusion, that the problem may be, that the transition to 'QUICK STOP ACTIVE' immediately triggers a transition to 'SWITCH ON DISABLED' so that the motor never really is in state 'QUICK STOP ACTIVE' so that your routine `while self.state != new_state:` never gets true and ends up comparing the current state which is 'SWITCH ON DISABLED' with 'QUICK STOP ACTIVE' which then is an illegal transition.

Do you have an idea how to fix this?

edit: This is my code.

```
import canopen

Net = canopen.Network()
    Net.connect(bustype='socketcan', channel='can0', bitrate=1000000)
    Node = canopen.BaseNode402(1, '/path_to_my.eds')
    Net.add_node(Node)
    Node.load_configuration()

    Node.setup_402_state_machine()
    Node.state = 'OPERATION ENABLED' 
    print(Node.state)            ## prints 'OPERATION ENABLED'
    Node.state = 'QUICK STOP ACTIVE'
    print(Node.state)            ## prints 'SWITCH ON DISABLED'
```

af-silva:
Hi @JulianS-Uni,

That is a strange behavior. I will look into it, but in the meanwhile at what time `ValueError: 'Illegal transition from 'SWITCH ON ENABLED' to 'QUICK STOP ACTIVE'.` this appears ??? 

In the file [p402.py](https://github.com/christiansandberg/canopen/blob/master/canopen/profiles/p402.py) you can find the state transition table and there is no **_"SWITCH ON ENABLED"_** command or state defined. 

Here is the CiA 402 Power State Machine:
![image](https://user-images.githubusercontent.com/11874980/68411293-19e47400-0182-11ea-9c0d-c1341a1bd088.png)
(image from [Nanotec](https://en.nanotec.com/products/manual/PD2C_CAN_EN/general%252Fds402_power_state_machine.html/))

Can you please confirm what the StatusWord holds when this error happens.

Thanks.

JulianS-Uni:
[...]

PR #172, opened 2019-11-26
small 402 cleanup

gsulc:
Some small changes to p402.py to make it a little easier to understand in preparation for solving issues #167 and #169 . I also improved some log statements.

I have two open questions about setup_402_state_machine(): why is it necessary to enter the "Pre-Operational" NMT state to read the pdo configurations. And why change state to "Switch on Disabled"? Shouldn't the present state be kept?

### DIFF
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -6,10 +6,8 @@ from ..sdo import SdoCommunicationError
 
 logger = logging.getLogger(__name__)
 
-
 class State402(object):
-
-    # Control word 0x6040 commands
+    # Controlword (0x6040) commands
     CW_OPERATION_ENABLED = 0x0F
     CW_SHUTDOWN = 0x06
     CW_SWITCH_ON = 0x07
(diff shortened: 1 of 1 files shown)


af-silva:
Sorry @gsulc, I'm with some work right now... I will try to review your changes by the end of the week. 
But a quick peek at the changes seems everything is ok, I just notice a small thing in the home function, the way you set the return of the function (return right the value instead of at the end of execution) bypass the finally statement of the try-catch when returning true...

But during the weekend I will review and discuss with you.

gsulc:
I understand. Get to it when you can.

gsulc:
I have solutions for those tickets. If you want, I can resolve the conflicts and merge, then submit those changes. I don't want to step on people's toes though.

af-silva:
> The return statement will execute after the finally clause. From the Python documentation:

You are absolutely right, my bad.
Today I will review and merge your changes with the code base. Keep in mind that I'm trusting the fixes are not breaking anything, at the moment I can't validate this with real hardware. 

> I have solutions for those tickets

What thickets are you referring to?

> I don't want to step on people's toes though

You are not stepping in anyone toes @gsulc, I'm happy to see people contributing code to this project.

PR #176, opened 2020-01-28
EDS : Add loading of DummyUsage definitions

stan-dalencon:
In EDS parsing, add load of [DummyUsage] section in order to allow PDO Dummy Mapping as defined in §4.6.2 - "Mapping of dummy entries" of CiA 306 (v1.3)

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -37,11 +37,11 @@ def import_eds(source, node_id):
 
     for section in eds.sections():
         # Match dummy definitions
-        match = re.match(r"^[D|d]ummy[U|u]sage$", section)
+        match = re.match(r"^[Dd]ummy[Uu]sage$", section)
         if match is not None:
-            for i in range(1,8):
+            for i in range(1, 8):
                 key = "Dummy%04d" % i
-                if eds.getint(section,key) == 1:
+                if eds.getint(section, key) == 1:
                     var = objectdictionary.Variable(key, i, 0)
                     var.data_type = i
                     var.access_type = "const"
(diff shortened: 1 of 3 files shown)


canopen/objectdictionary/eds.py
```diff
@@ -36,6 +36,17 @@ def import_eds(source, node_id):
         od.node_id = int(eds.get("DeviceComissioning", "NodeID"))
 
     for section in eds.sections():
+        # Match dummy definitions
+        match = re.match(r"^[D|d]ummy[U|u]sage$", section)
```
christiansandberg: You don’t need the | here. All characters inside [] will be matched.

canopen/objectdictionary/eds.py
```diff
@@ -36,6 +36,17 @@ def import_eds(source, node_id):
...
+        if match is not None:
+            for i in range(1,8):
+                key = "Dummy%04d" % i
+                if eds.getint(section,key) == 1:
```
christiansandberg: Nitpicking a little but add a space after comma here and in the range call.

christiansandberg:
Could you also add a simple test? Modify the sample.eds file and add a check in test_eds.py.

canopen/objectdictionary/eds.py
stan-dalencon: Right, will remove the "pipes"

canopen/objectdictionary/eds.py
stan-dalencon: No problem.

stan-dalencon:
> Could you also add a simple test? Modify the sample.eds file and add a check in test_eds.py.

Sure, I add two tests:
- one with defined dummy, to check var properties
- one with an undefined dummy, to check that KeyError exception is thrown

stan-dalencon:
@christiansandberg : Don't know why CI build failed especially in a part I didn't modify, any idea ?

christiansandberg:
Thanks! The failing test was due to a timing issue.

stan-dalencon:
> Thanks! The failing test was due to a timing issue.

Ok, thanks a lot.

PR #177, opened 2020-01-31
Fix #167 and #169 tricky state transitions

### LINKED ISSUE #167: BaseNode402 Transition from 'OPERATION ENABLED' to 'QUICK STOP ACTIVE'

gsulc:
I haven't been able to test this, but this change should fix these issues. I should be able to test them at some point next week, but it would be good to start the review process and get any early feedback.

### DIFF
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -185,6 +185,7 @@ class BaseNode402(RemoteNode):
         super(BaseNode402, self).__init__(node_id, object_dictionary)
         self.tpdo_values = dict() # { index: TPDO_value }
         self.rpdo_pointers = dict() # { index: RPDO_pointer }
+        self.use_state_transition_16 = True
 
     def setup_402_state_machine(self):
         """Configure the state machine by searching for a TPDO that has the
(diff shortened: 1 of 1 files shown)


gsulc:
Gah. Sorry. Next week. I have hardware at work, but other things are taking priority even though I'm utilizing this project.

yyjpeter:
canopen

PR #190, opened 2020-05-30
Add unit test for SYNC producer

paoloteti:
Test one-shot SYNC transmission with and without counter value.


christiansandberg:
Thanks a lot for your contribution!

PR #204, opened 2020-10-13
callback possibility on block transfer after each segment

arne123:
To update application information, e.g. a progress bar, during huge block transfers like firmware updates, it makes sense to get information during an ongoing block transfer. 
This is an implementation for a block download with a callback after each block end.


christiansandberg:
Thanks you for your contribution! One reason for using a file like interface for large transfers is so that you split it up in smaller chunks and do other things inbetween, like updating a progress bar for example. You could for example do something like this:

```python
FIRMWARE_PATH = '/path/to/firmware.bin'
FILESIZE = os.path.getsize(FIRMWARE_PATH)
infile = open(FIRMWARE_PATH, 'rb')
outfile = node.sdo['Firmware'].open('wb', size=FILESIZE, block_transfer=True)

# Iteratively transfer data without having to read all into memory
while True:
    # Read appropriately large chunks
    data = infile.read(128)
    if not data:
        break
    outfile.write(data)
    progress = outfile.tell() / FILESIZE
    # Update GUI with progress
infile.close()
outfile.close()
```

Do you think that would fulfill your needs?

arne123:
Thanks for your input, this looks like an approach, 
I am not sure yet, we had an issue with chunking before, but the approach was different.
I will that give it a try and let you know.

christiansandberg:
You may also need to adjust the [`buffering` argument](https://canopen.readthedocs.io/en/latest/sdo.html#canopen.sdo.SdoClient.open) (probably to a value <= your chunk size) to get a smooth progress update.

christiansandberg:
@arne123 did you manage to get it working?

arne123:
Yes, this works quite well so far, there is also no measurable effect on timing.
I didn't test it on a bigger scale, but I think we can discard and close the pull request.

PR #206, opened 2020-11-05
[StorageLocation] Add Support for StorageLocation of EDS

rubenverhoef:
Hi!

I'm using the StorageLocation functionality from the eds. This pull request makes it possible to use it.

For me this is working, I don't know if it is according to your coding style and if I miss something?

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -116,14 +116,14 @@ class Record(MutableMapping):
     #: Description for the whole record
     description = ""
 
-    def __init__(self, name, index, StorageLocation):
+    def __init__(self, name, index, storage_location):
         #: The :class:`~canopen.ObjectDictionary` owning the record.
         self.parent = None
         #: 16-bit address of the record
         self.index = index
         #: Name of record
         self.name = name
-        self.StorageLocation = StorageLocation
+        self.storage_location = storage_location
         self.subindices = {}
         self.names = {}
 
@@ -171,14 +171,14 @@ class Array(Mapping):
     #: Description for the whole array
     description = ""
 
-    def __init__(self, name, index, StorageLocation):
+    def __init__(self, name, index, storage_location):
         #: The :class:`~canopen.ObjectDictionary` owning the record.
         self.parent = None
         #: 16-bit address of the array
         self.index = index
         #: Name of array
         self.name = name
-        self.StorageLocation = StorageLocation
+        self.storage_location = storage_location
         self.subindices = {}
         self.names = {}
 
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -116,13 +116,14 @@ class Record(MutableMapping):
     #: Description for the whole record
     description = ""
 
-    def __init__(self, name, index):
+    def __init__(self, name, index, StorageLocation):
         #: The :class:`~canopen.ObjectDictionary` owning the record.
         self.parent = None
         #: 16-bit address of the record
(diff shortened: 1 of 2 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -116,13 +116,14 @@ class Record(MutableMapping):
     #: Description for the whole record
     description = ""
 
-    def __init__(self, name, index):
+    def __init__(self, name, index, StorageLocation):
```
christiansandberg, line 119: The way I added extra information to the Variable object is to write those attributes after it has been created. This could definitely be debated but it keeps constructor clean and backwards compatible. So I would prefer if you removed it from here and just write the value later.

canopen/objectdictionary/__init__.py
```diff
@@ -116,13 +116,14 @@ class Record(MutableMapping):
...
         #: The :class:`~canopen.ObjectDictionary` owning the record.
         self.parent = None
         #: 16-bit address of the record
         self.index = index
         #: Name of record
         self.name = name
+        self.StorageLocation = StorageLocation
```
christiansandberg, line 126: Use a default value of None here so it can be easily identified that it was not present in the EDS.

Add documentation about the attribute by using ```#: Some text```.

canopen/objectdictionary/eds.py
```diff
@@ -59,23 +59,27 @@ def import_eds(source, node_id):
                 # If the keyword ObjectType is missing, this is regarded as
                 # "ObjectType=0x7" (=VAR).
                 object_type = VAR
+            try:
+                StorageLocation = eds.get(section, "StorageLocation")
+            except NoOptionError:
+                StorageLocation = "Unknown"
```
christiansandberg, line 65: Would prefer None if it was not present.

christiansandberg (changes requested):
Thank you for your contribution! The standard used is PEP-8, so it should be named `storage_location` instead.

canopen/objectdictionary/__init__.py
rubenverhoef: How can i add extra information later for an `Array` or `Record` ? Do i need to make a funciton my own or are there functions to do that inside the code?

christiansandberg:
```python
arr = objectdictionary.Array(name, index)
arr.storage_location = storage_location
```

Similar to how it's done on Variable.

canopen/objectdictionary/eds.py
```diff
@@ -154,6 +158,10 @@ def build_variable(eds, section, node_id, index, subindex=0):
     """
     name = eds.get(section, "ParameterName")
     var = objectdictionary.Variable(name, index, subindex)
+    try:
+        var.storage_location = eds.get(section, "StorageLocation")
+    except NoOptionError:
+        var.storage_location = None
```
christiansandberg, line 164: Also add this member to the Variable class so it is documented.

rubenverhoef:
Thanks for your information, i have implemented your feedback.

How about this section:
https://github.com/christiansandberg/canopen/blob/bf3b19430ec40bd6c3330e60ea5d1e5cc7df2534/canopen/objectdictionary/__init__.py#L194-L196

Should the `storage_location` also be listed there?
Otherwise it think this was it?

christiansandberg:
Yes you're right about that. Otherwise it looks good. 👍

rubenverhoef:
I think it is now ready for merge, can you look into it?

Thanks!

christiansandberg (approved):

PR #218, opened 2021-01-17
Added script to create debian package

unsanded:
I needed it. Thought someone else might like it to. 

Tested in archlinux and python docker image. 
Package maintainer is set to whoever made the last commit. So, if you merge it will be you :)


christiansandberg:
Thank you!

Issue #224, opened 2021-02-18
Exception from Threading on multiple stop_node_guarding()


Bundesber:
If you stop twice the Node Guarding on one Node, you will receive some Exception from the python Threading
https://github.com/christiansandberg/canopen/blob/bb6b528e49aa888ee22354f2264bc7d0ada2be34/canopen/nmt.py#L186

setting the self._node_guarding_producer to None after stopping helps
`self._node_guarding_producer = None`

https://github.com/christiansandberg/canopen/blob/bb6b528e49aa888ee22354f2264bc7d0ada2be34/canopen/nmt.py#L181

Also a check for None before starting, would be a good condig style

`if self._node_guarding_producer:
    self.stop_node_guarding()`

af-silva:
PR merge with the proposed fix.

PR #225, opened 2021-02-18
fix for exception on multipe stop

Bundesber:
#224 
Request to this issue


af-silva (approved):
Seems ok, accepting the code submission.

PR #239, opened 2021-06-03
added SDO Abort if a timeout occurs when sending a SDO request

jnu144:
If during a SDO Request (e.g. block transfer) a timeout occurs, the SDO request will stop but the SDO Server will still be in the receiving state until an SDO Abort is sent. With this fix an SDO Abort is sent when a timeout occurs and with this the SDO Server state machine will be reset.


christiansandberg (approved):
Thanks! I don't see a reason this change would cause any issues but I'll keep it open for a couple of days in case someone else would have an opinion.

PR #240, opened 2021-06-07
added argument for block transfer to enable/disable CRC Support

jnu144:
We had some problems with domain transfers when the SDO Server does not support the CRC_SUPPORTED Flag for the Block Transfer. With this you can disable (optional) the CRC_SUPPORTED Flag when calling open().


christiansandberg (changes requested):
Thanks! Could you also add the new argument to the docstrings?

jnu144:
Ah yes, sorry I forgot

christiansandberg (approved):

PR #245, opened 2021-07-27
DS402 code cleanup

acolomb:
These are some hopefully uncontroversial changes to clean up the `p402` module code.  Mainly style issues without actual change in behavior, some reported by the `flake8` tool.

Some missing DS402 operation mode constants were added.  Timeout handling uses the `time.monotonic()` clock now and all durations are configurable through class constants (respectively object attributes if overwritten).

Reviewing each commit individually is probably easiest, and the commit messages usually contain some rationale.


af-silva:
@acolomb Nice Work, Everything seems ok, I will accept your PR.

PR #252, opened 2021-08-02
DS402: Increase delay to check status on homing start.

acolomb:
The Statusword is examined immediately after setting the Controlword command to start homing.  That is very likely to fail because of the round-trip time until the Statusword is actually updated from a TPDO.  To work around that, the delay between each check of the Statusword should be moved before the actual comparison, and its default value increased.

Introduce a new constant TIMEOUT_CHECK_HOMING to configure that with a default value of 100 ms.  This replaces the previously used INTERVAL_CHECK_STATE which is only 10 ms by default.

An even better solution would be to wait for the Statusword to be updated by a received PDO, but that would be much more complex.


acolomb:
Note that #263 includes these changes, but goes one step further in eliminating the delays altogether when a periodic TPDO is configured (also for other cases).

PR #253, opened 2021-08-06
PDO: Allow subscription without calling read() or save()

acolomb:
Normally one needs to call read() or save() on a single pdo (or the node's whole PDO collection) in order to receive any such objects from the network.  That however requires quite a few SDO exchanges to make sure the node's PDO configuration matches the parameters configured in the object.

For applications where the PDO configuration is stored persistently in the node (e.g. device EEPROM), doing this SDO exchange can be skipped entirely if the application programmer takes care to mirror the same configuration in the python-canopen objects.  Another use case is reconnecting to a node for which the same python-canopen script previously ran and the PDO configuration is still known to be valid.

Factor out a new method subscribe() from read() and save() to offer doing only that last part via the public API.  Adapt the log message and make sure it is logged in read() as well.

### DIFF
--- a/canopen/pdo/__init__.py
+++ b/canopen/pdo/__init__.py
@@ -8,7 +8,8 @@ logger = logging.getLogger(__name__)
 
 
 class PDO(PdoBase):
-    """PDO Class for backwards compatibility
+    """PDO Class for backwards compatibility.
+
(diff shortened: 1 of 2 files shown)


christiansandberg (approved):

acolomb:
Thanks :-)

PR #254, opened 2021-08-09
Export Eds or Dcf

unsanded:
I needed this for work, so i wrote it. Also added a unit test. 

## supported
 - Write eds
 - Write Dcf with extra device commisioning info
 - DeviceInfo section
 - cache FileInfo
 
## not supported
 - Eds comments 
 - Write `compactSubObj` sections
 - Features i didn't know about?

Ps. I improved the deb script slightly. (My previous pull request) So i just included it.
I could rebase it out if we want to be all proper like that :)


canopen/objectdictionary/__init__.py
```diff
@@ -13,6 +13,20 @@
 logger = logging.getLogger(__name__)
 
 
+def export_od(od, dest=None, docType="eds"):
```
christiansandberg, line 16: Maybe the doc type could be inferred from the file extension if left out?

canopen/objectdictionary/__init__.py
```diff
@@ -13,6 +13,20 @@
...
+    """ Export :class: ObjectDictionary to a file.
+    :param docType: type of document to export
+    :rtype: str or None
+    """
```
christiansandberg, line 20: The docstring doesn't correlate with the actual interface here.

christiansandberg (changes requested):
This will be a nice addition!

We should however stick to PEP-8 coding conventions as much as possible. Variables and attributes should be snake_case and whitespace used consistently.

https://www.python.org/dev/peps/pep-0008/

canopen/objectdictionary/__init__.py
unsanded: Yeah that makes sense. Though i originally thought that dest should only be a file descriptor. Now i will also accept a filename.

canopen/objectdictionary/__init__.py
unsanded: What is says is, i think, true but i admit that is says far too little.  I will expand it a little.

unsanded:
Ok, I think that is it. I think  i processed all of the feedback :)  Thanks for the quick review btw :+1:

unsanded:
Thanks @acolomb I looked before starting, but i didn't see it.

unsanded:
I changed it to a WIP, because i am using it in a project now, and i keep finding missing features. So i'll just add those, and put it back

christiansandberg:
Sounds good! There are some camelCase style code left also.

PR #264, opened 2021-08-12
DS402: Fix logic errors in the power state machine and generalize it.

acolomb:
Extend the DS402 State Machine implementation to reach more states automatically by going through several transitions.  So far that was only possible for the OPERATION ENABLED state.  Add a testing script to verify how the logic works and fix some bugs found using the test script.

See individual commit messages for detailed explanations.

### DIFF
--- a/canopen/profiles/p402.py
+++ b/canopen/profiles/p402.py
@@ -502,10 +502,11 @@ class BaseNode402(RemoteNode):
                             'FAULT'):
             raise ValueError(
                 'Target state {} cannot be entered programmatically'.format(target_state))
-        if target_state == 'OPERATION ENABLED':
-            return State402.next_state_for_enabling(self.state)
-        else:
+        from_state = self.state
+        if (from_state, target_state) in State402.TRANSITIONTABLE:
             return target_state
+        else:
+            return State402.next_state_for_enabling(from_state)
 
     def _change_state(self, target_state):
         try:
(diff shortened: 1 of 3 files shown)


af-silva (approved):

af-silva (approved):

af-silva (changes requested):
Maybe it would be better to move the test script into the test folder at the root or create a new folder for tests that target the DS402 profile (e.g., canopen/test/ds402).

acolomb:
@af-silva did you have a chance to test yet or is the approval only related to reviewing the code?  It's running fine in my application so from my side, I think it's ready to merge.

acolomb:
Sorry your reply was showing up delayed... Okay I will try to move it to a subdirectory. Not sure how to import the package then, as relative imports can only be used from within the package.

acolomb:
Never mind about the import question.

I see the existing test files are somewhat different, they are unit tests.  The new script added here is rather meant to be used as a manual diagnostics tool.  Not sure how that should be integrated into the unit test suite, or if it should just live within the same directory structure but not really be hooked up to the test suite...

af-silva:
> @af-silva did you have a chance to test yet or is the approval only related to reviewing the code?  It's running fine in my application so from my side, I think it's ready to merge.

It's just from reviewing the code. Here I have to trust you for validation on actual hardware, since I don't have access to it right now. Everything seems ok, you spot the wrong transitions that I have introduce and fix them, nice work!

af-silva:
> Never mind about the import question.
> 
> I see the existing test files are somewhat different, they are unit tests.  The new script added here is rather meant to be used as a manual diagnostics tool.  Not sure how that should be integrated into the unit test suite, or if it should just live within the same directory structure but not really be hooked up to the test suite...

hum ok, maybe then crate a test or tools inside the ds402 profile, to better separation and even allowing the creation/extension of other tolls? I now the file is called "test_" but I would put it inside a folder for tests or tools...

af-silva (approved):
Ok, thanks.

PR #267, opened 2021-09-06
Bug fixes, send/receive debug logs and  Unsigned 24 support 

TzviRonen:
Changes:
* Added test if callback exist before removing it from network subscribers.
* Added to sdo client and sdo server, send and receive debug logs.
* Added pause time before retrying send_request in sdo client request_response.
* Fixed the var len calculation in pdo base get_data. This one took me hours to debug..
* Added support for Unsigned 24 data types.


canopen/sdo/client.py
```diff
@@ -42,6 +45,7 @@ def __init__(self, rx_cobid, tx_cobid, od):
         self.responses = queue.Queue()
 
     def on_response(self, can_id, data, timestamp):
+        logger.debug(f"received response in {can_id} data {data}.")
```
christiansandberg, line 48: As I understand it, it is preferable to use old-style string interpolation for logging because then the final message will only be constructed if needed.

https://docs.python.org/3/howto/logging.html#logging-variable-data

canopen/objectdictionary/__init__.py
```diff
@@ -232,6 +243,7 @@ class Variable(object):
         UNSIGNED8: struct.Struct("B"),
         UNSIGNED16: struct.Struct("<H"),
         UNSIGNED32: struct.Struct("<L"),
+        UNSIGNED24: Unsigned24(),
```
acolomb, line 246: Should be ordered by size?

.gitignore
```diff
@@ -4,6 +4,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+.idea/
```
acolomb, line 7: There is a separate section for IDE stuff down below, if such things must be collected here.

This change seems unrelated though, so let's keep it in a separate commit / PR.

acolomb (changes requested):
I wish there was more explanation / better commit messages for these changes.  Some cleanup required IMHO and better to keep things separated for different purposes, so I would recommend against squash-merging this as is.

acolomb:
Regarding the additional log messages, I think that's too much verbosity.  I sometimes rely on the `DEBUG` log level setting, but this would really spam the log without much additional gain.  Is there some timing issue you're trying to debug?  Then maybe having these debug messages locally is better than adding them here for every library user.

canopen/sdo/client.py
```diff
@@ -29,6 +29,9 @@ class SdoClient(SdoBase):
     #: Seconds to wait before sending a request, for rate limiting
     PAUSE_BEFORE_SEND = 0.0
 
+    # Seconds to wait before next read attempt for response in queue. For delayed responses.
+    PAUSE_BEFORE_READ = 0.2
```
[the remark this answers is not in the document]
TzviRonen: @acolomb If it fails to respond to the send a request that may be a sign that the device isn't ready and not necessarily a communication error.
Why not use RESPONSE_TIMEOUT? I want to know about it in the log's as soon as it happens and yet, wait more time before the next read_response().
I used it only for debugging. I agree with you it should be optional by defaulting it to zero.
And I agree with @christiansandberg about the naming. I changed the name to RETRY_DELAY.

canopen/objectdictionary/__init__.py
```diff
@@ -219,6 +219,17 @@ def add_member(self, variable):
         self.subindices[variable.subindex] = variable
         self.names[variable.name] = variable
 
+class Unsigned24(struct.Struct):
+    def __init__(self, *args, **kwargs):
+        super(Unsigned24, self).__init__("<I", *args, **kwargs)
```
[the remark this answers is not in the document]
TzviRonen: Fixed the super thing.
This isn't an attempt to convert the uint24 to a regular 32-bit. The only reason I'm converting it to an ```I``` is because that's the closest data types ```Struct``` supports. 
Then I'm using ```Struct.unpack("<I)``` to make that 32-bit int to python int.

canopen/objectdictionary/__init__.py
TzviRonen: fixed

canopen/pdo/base.py
```diff
@@ -562,7 +562,7 @@ def get_data(self):
                 data = data | (~((1 << self.length) - 1))
             data = od_struct.pack(data)
         else:
-            data = self.pdo_parent.data[byte_offset:byte_offset + len(self.od) // 8]
+            data = self.pdo_parent.data[byte_offset:byte_offset + self.length // 8]
```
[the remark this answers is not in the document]
TzviRonen: I couldn't think of any reason len (self. OD) is used here.
Somehow len ```(self.od)``` changed during the program execution and wasn't equal to ```self.length``` when ```get_data``` was called. It made the returned data to be in the wrong size. Changing it to ```self.length``` fixed the problem.

TzviRonen (commented):
> Regarding the additional log messages, I think that's too much verbosity. I sometimes rely on the `DEBUG` log level setting, but this would really spam the log without much additional gain. Is there some timing issue you're trying to debug? Then maybe having these debug messages locally is better than adding them here for every library user.

This log very helped me debug as they are at the very core of the library. 
But you probably know better about the library logging level so I'll remove them to not spam the log's. 
They indeed lead to a huge log file..

canopen/sdo/client.py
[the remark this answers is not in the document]
acolomb: It's a very slight notion, but `time.sleep(0)` will allow switching to a different thread AFAIK, while not calling it at all does not.  I suppose that doesn't make much difference, but maybe better check the constant first and skip sleep if it's zero?

canopen/pdo/base.py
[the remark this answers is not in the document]
acolomb: Well, `len(self.od)` returns the number of bits (rounded up to full bytes) that the object consists of according to the node's object dictionary, which is constant.  While `self.length` starts off with the same value, it might be overridden when adding the `Variable` to a PDO mapping, where you can e.g. specify to map only (the lower) 16 bits of a 32 bit unsigned object.  That difference is adjusted in `pdo.Map.add_variable()`.

I agree there are some corner cases in the code that don't deal well with mapping only some bits of an object.  But to fix that properly, each use within `get_data()` and `set_data()` should be reconsidered individually.  Just changing one instance might fix your problem, but cause issues for other users.

By the way, my (hacky) solution so far was to adjust the `ObjectDictionary` instance to artificially overwrite the `data_type` member when using partial objects in PDO mappings.  That breaks SDO access though, and doesn't cope well with signed numbers (two's complement requires changing the higher bytes as well when crossing zero).

canopen/objectdictionary/__init__.py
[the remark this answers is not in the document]
acolomb: I had to look it up in the Python library docs, but as I see it, the `<L` and `<I` formats are equivalent, both using four bytes for the conversion.  So your code seems correct, just wondering why not use the same letter when it doesn't really make a difference?

canopen/objectdictionary/__init__.py
```diff
@@ -219,6 +219,17 @@ def add_member(self, variable):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__("<I", *args, **kwargs)
+
+    def unpack(self, data, *args, **kwargs):
+        if isinstance(data, bytearray):
+            while len(data) < 4:
+                data += b'\x00'
+        else:
+            logger.error(f"Unsigned24.unpack received wrong type - {type(data)}")
+        return super(Unsigned24, self).unpack(data, *args, **kwargs)
```
acolomb, line 232: Another `super()` thing.

canopen/objectdictionary/__init__.py
```diff
@@ -219,6 +219,17 @@ def add_member(self, variable):
         self.subindices[variable.subindex] = variable
         self.names[variable.name] = variable
 
+class Unsigned24(struct.Struct):
+    def __init__(self, *args, **kwargs):
+        super().__init__("<I", *args, **kwargs)
```
acolomb, line 224: As previously noted, use `"<L"` for consistency with the UNSIGNED32 type.

canopen/sdo/client.py
```diff
@@ -89,6 +92,7 @@ def request_response(self, sdo_request):
                     self.abort(0x5040000) 
                     raise
                 logger.warning(str(e))
+                time.sleep(self.RETRY_DELAY)
```
acolomb, line 95:
```suggestion
                if self.RETRY_DELAY:
                    time.sleep(self.RETRY_DELAY)
```

canopen/sdo/server.py
```diff
@@ -172,6 +172,7 @@ def segmented_download(self, command, request):
         self.send_response(response)
 
     def send_response(self, response):
+        logger.debug(f"Sending to {self.tx_cobid} data {response}.")
```
acolomb, line 175: Another overly verbose log message.

acolomb (changes requested):
Could you please still consider breaking up the changes into several commits / PRs with only related changes?  I think pinpointing any possible issues will be easier when each commit holds only changes related to the same topic.

canopen/objectdictionary/__init__.py
```diff
@@ -219,6 +219,17 @@ def add_member(self, variable):
...
+
+    def unpack(self, data, *args, **kwargs):
+        if isinstance(data, bytearray):
+            while len(data) < 4:
+                data += b'\x00'
+        else:
+            logger.error(f"Unsigned24.unpack received wrong type - {type(data)}")
```
acolomb, line 231: Please use `%`-style format string.

acolomb (changes requested):

sveinse:
Isn't this PR outdated? PR #429, PR #435 implements delay on SDO send and retries. PR #406 implements 24-bit support. PR #440 will implement the remaining datatypes. Perhaps the PR can be closed?

acolomb:
I've skimmed over the changes and converstion again.  Not quite sure whether all problems addressed here have been dealt with.  One of the main issues was that the PR contains many unrelated changes all-in-one.  That now makes it harder to dissect what has already been fixed elsewhere, and where issues remain.

@sveinse if you're motivated to check closely, extracting the relevant problem descriptions into separate issues would help.  Pointing back at this PR for tentative fixes.  I might get to it sometime, but don't hold your breath...

canopen/pdo/base.py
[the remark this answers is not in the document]
sveinse: @acolomb This one

sveinse (commented):
@acolomb I reviewed the patch and I think everything is covered already with one exception which I have commented.

acolomb:
@sveinse That remaining change should definitely go in.  I even found it in one of my older branches, just never got around to making a PR.

sveinse:
@acolomb Should I embed it into PR #440 ?

acolomb:
No, that one is unrelated.  I'm currently checking what other fixes I had started to work on about 1.5 years ago and will make a separate PR.

acolomb:
The rework of PDO variable decoding (especially for mapped lengths differing from the OD size) is done in #454.  Merging that will make the remaining bit of this PR obsolete I suppose.

PR #268, opened 2021-09-10
Do not raise an exception in "_check_statusword_configured" just log …

Lauszus:
…it as a warning


acolomb (commented):
Looks reasonable, as there is already a check in the statusword property falling back to SDO.

acolomb (approved):

PR #279, opened 2021-11-09
eds: Fix exception in comment parsing with hexadecimal line count.

acolomb:
This fixes a bug introduced in #254, which immediately broke parsing of my EDS files.

Inside an EDS file's [Comments] section, the Lines attribute may be expressed as a hexadecimal number prefixed with 0x.  The ConfigParser.getint() method however does not support such a format.

Switch to the default int() constructor with a base=0 argument to enable auto-detection of this condition.  The same is done for basically every other numeric value in the EDS importer.

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -41,7 +41,7 @@ def import_eds(source, node_id):
         }
 
     if eds.has_section("Comments"):
-        linecount = eds.getint("Comments", "Lines")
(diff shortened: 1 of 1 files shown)


acolomb:
Please have a look @unsanded.

unsanded:
Oh good one. I think i missed that one. I started to use getInt everywhere, but i found out that int(...,0) is the right way.

af-silva (approved):

PR #280, opened 2021-11-09
Document how to run the Python unit tests

henrikbrixandersen:
Add support for Python unit test autodiscovery and document how to run them.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,6 +51,10 @@ it in `develop mode`_::
     $ cd canopen
     $ pip install -e .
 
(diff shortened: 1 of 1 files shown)


christiansandberg:
I think moving to pytest is a better move. We already execute the tests using pytest in CI and we might start writing new tests in pytest later.

henrikbrixandersen:
> I think moving to pytest is a better move. We already execute the tests using pytest in CI and we might start writing new tests in pytest later.

Good point. I have updated the instructions to use `pytest` instead.

PR #281, opened 2021-11-09
Add 'relative' property for COB-ID vars relative to the node-ID

henrikbrixandersen:
Add 'relative' property for COB-ID vars relative to the node-ID.

I have used the term "relative" only since I could not come up with something better (and CiA306-1 doesn't really help here). Other suggestions are most welcome.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -304,6 +304,8 @@ class Variable(object):
         self.max: Optional[int] = None
         #: Default value at start-up
         self.default: Optional[int] = None
+        #: Is the default value relative to the node-ID (only applies to COB-IDs)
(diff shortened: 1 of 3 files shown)


henrikbrixandersen:
@christiansandberg Any objections to adding this? I would like to use Python canopen to auto-generate C-code from EDS files for use in a native CANopen stack for the Zephyr RTOS, but I need to be able to determine if a given object entry contains a COB-ID relative to the node-ID.

We are already using Python canopen in Zephyr for performing program download (CiA302-3) and all the Zephyr CANopen examples contain Python canopen snippets for testing device functionality.

christiansandberg:
Sorry for the late feedback. I guess it wouldn't hurt to add another property if it will helpful.

PR #287, opened 2021-12-09
Fix canmatrix export

walmis:
Exporting PDO mappings seems to be broken with canmatrix >= 0.9 (haven't checked exactly where it broke).
This PR should fix this.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -114,7 +114,7 @@ class PdoBase(Mapping):
                     signal.addValues(value, desc)
                 frame.add_signal(signal)
             frame.calc_dlc()
-            db.frames.append(frame)
+            db.add_frame(frame)
         formats.dumpp({"": db}, filename)
         return db
 
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -88,8 +88,7 @@ class PdoBase(Mapping):
             if pdo_map.cob_id is None:
                 continue
             frame = canmatrix.Frame(pdo_map.name,
-                                    Id=pdo_map.cob_id,
-                                    extended=0)
+                                    arbitration_id=pdo_map.cob_id)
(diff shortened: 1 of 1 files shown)


canopen/pdo/base.py
```diff
@@ -113,9 +112,9 @@ def export(self, filename):
                                           unit=var.od.unit)
                 for value, desc in var.od.value_descriptions.items():
                     signal.addValues(value, desc)
-                frame.addSignal(signal)
-            frame.calcDLC()
-            db.frames.addFrame(frame)
+                frame.add_signal(signal)
+            frame.calc_dlc()
+            db.frames.append(frame)
```
acolomb, line 117: I don't see a `frames` attribute documented in the `CanMatrix` API reference.  It might work, but there seem to be several `add_frame_*()` methods which look more official.

Would you please check if any of those would be suitable here?

acolomb (changes requested):
Looks good from the code, I haven't tested anything here.

walmis:
Updated to use add_frame method

acolomb:
> This PR should fix this.

Have you done any testing?

Sorry I have never used such functionality and don't see when I'll have time to dig deeper into testing it.  I agree that it was broken before, so we can't make it much worse.  Just would like to hear that the changes have actually been tested on a real world scenario.

walmis:
> > This PR should fix this.
> 
> Have you done any testing?
> 
> Sorry I have never used such functionality and don't see when I'll have time to dig deeper into testing it. I agree that it was broken before, so we can't make it much worse. Just would like to hear that the changes have actually been tested on a real world scenario.

Yes, it exported the mapping as expected. It works as far as it doesn't throw a unknown method exception.

acolomb (approved):

acolomb:
Merging as it cannot realistically make things worse than not working at all.

PR #298, opened 2022-03-08
classic LSS scan

LBekel:
added return value to send_identify_remote_slave function to perform classic LSS scan, add flag to remove delay from __send_lss_address function

### DIFF
--- a/canopen/lss.py
+++ b/canopen/lss.py
@@ -62,6 +62,7 @@ ListMessageNeedResponse = [
     CS_INQUIRE_REVISION_NUMBER,
     CS_INQUIRE_SERIAL_NUMBER,
     CS_INQUIRE_NODE_ID,
+    CS_IDENTIFY_REMOTE_SLAVE_SERIAL_NUMBER_HIGH,
 ]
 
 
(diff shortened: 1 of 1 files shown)


bizfsc:
Thanks for this PR — the basic idea is correct and follows the existing pattern from `send_switch_state_selective`. However, there are a few issues that need to be addressed before we can merge this:

**1. No tests.**
The newly implemented `True`/`False` return behaviour of `send_identify_remote_slave` is not covered by any test. Please add tests following the pattern already established in `test_lss.py`.

**2. Parameter naming.**
`nodelay=False` is a double negative — `delay=True` (or similar) would read more naturally at call sites.

**3. Delay removal may break slow devices.**
The existing code comment explicitly states *"some device needs these delays between messages because it can't handle messages arriving with no delay."* Removing all inter-message delays may silently break compatibility with such devices. A short delay (e.g. 1–5 ms) between the first five messages would be a safer compromise while still keeping the scan reasonably fast.

**4. Silent handling of unexpected responses.**
When the response CS is neither `CS_IDENTIFY_SLAVE` nor causes a timeout, the function silently returns `False`. Other methods in this file raise `LssError` on an unexpected CS — please be consistent.

We would be happy to merge this once these points are addressed. If there is no update within the next few weeks, we will close the PR to keep the backlog manageable.

PR #306, opened 2022-04-03
Fix unhandled exception in pdo read

NicoPowers:
This fixes the unhandled exception SdoCommunicationError that is raised in the SdoClient when trying to read PDO parameters

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -8,7 +8,7 @@ except ImportError:
 import logging
 import binascii
 
-from ..sdo import SdoAbortedError
+from ..sdo import SdoAbortedError, SdoCommunicationError
 from .. import objectdictionary
 from .. import variable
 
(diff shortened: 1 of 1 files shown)


acolomb:
Could you explain a little more please?  What kind of error are you seeing?  What's the cause?  What kind of CANopen node are you talking to?

NicoPowers:
> Could you explain a little more please? What kind of error are you seeing? What's the cause? What kind of CANopen node are you talking to?

Yes sorry, I am interfacing with a chinese servo motor and it doesn't receive an SDO response for event timers or SYNC starts, and the function that looks for these parameters in the SdoClient will raise an SdoCommunicationError; which is currently not handled by the rpdo and tpdo read functions.

acolomb:
What does the device send then? An SDO abort code? Nothing? Just silencing arbitrary error classes will hide problems that other people may need to notice, drill into, and fix properly at the source. That's why I'm asking for more details, so we can make a targeted fix or make it opt-in for devices that don't adhere to the spec completely, instead of making error diagnosis harder for everyone.

acolomb:
I don't agree with the approach taken here, and no more information was provided to justify it.  Closing because of timed out feedback.  Feel free to reopen with additional details.

PR #324, opened 2022-08-06
Adding retransmit functionnality to sdo block download

samsamfire:
I am submitting a pull request to add retransmission functionnality for sdo block download.
I faced a hardware issue (silicon errata) on a DSP where in some very rare cases the CAN rx fifo would swap the reception of two or more CAN frames causing the block transfer to fail, this was a nice solution to overcome the problem.

### DIFF
--- a/canopen/sdo/client.py
+++ b/canopen/sdo/client.py
@@ -629,7 +629,7 @@ class BlockDownloadStream(io.RawIOBase):
         self._crc = sdo_client.crc_cls()
         self._last_bytes_sent = 0
         self._current_block = []
-        self._retransmit_counter = 0
+        self._retransmitting = False
         command = REQUEST_BLOCK_DOWNLOAD | INITIATE_BLOCK_TRANSFER
         if request_crc_support:
             command |= CRC_SUPPORTED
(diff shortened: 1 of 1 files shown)


canopen/sdo/client.py
```diff
@@ -731,14 +736,38 @@ def _block_ack(self):
             raise SdoCommunicationError("Server did not respond with a "
                                         "block download response")
         if ackseq != self._blksize:
+            # Sequence error, try to retransmit
             self.sdo_client.abort(0x05040003)
```
christiansandberg, line 740: I don't quite remember the spec, is this abort code still needed?

christiansandberg (commented):
Thanks for this PR!

canopen/sdo/client.py
samsamfire: image
This is taken from the spec and apparently an SDO abort is only necessary in case of a fatal error, so  I will remove it.
However it still worked with this abort message, so we probably have a bug on our side.

canopen/sdo/client.py
```diff
@@ -731,14 +736,38 @@ def _block_ack(self):
...
         self._blksize = blksize
         self._seqno = 0
+        
+    def _retransmit(self, ackseq, blksize):
+        """Retransmit the failed block"""
+        logger.info(("%d of %d sequences were received. "
+                 "Will start retransmission") % (ackseq, self._blksize))
+        # Sub blocks betwen ackseq and end of corrupted block need to be resent
+        # Copy block to resend, and clear it so that multiple retransmits can be done
+        block = self._current_block[ackseq:].copy()
```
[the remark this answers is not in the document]
samsamfire: Yes you are absolutely right !

canopen/sdo/client.py
christiansandberg: No it's not very common for clients to send abort codes to the server, so some servers may just ignore them.

samsamfire: Ok, good to know.

PR #338, opened 2022-12-08
Check data length sdo server

samsamfire:
Currently, if writing a value with the wrong size inside object dictionary, no errors are raised.
This can cause all sorts of problem : for example when sending PDO frames, the frame could be greater than 8 bytes, etc.
I added a a length check when the expected datatype is of NUMBER_TYPE.


christiansandberg (approved):
Thanks, looks reasonable.

acolomb (approved):

acolomb:
I guess the automated checks need some attention, though nothing related to the changes in this PR.

PR #341, opened 2023-01-03
Reworking PDOs for Local Nodes

samsamfire:
Hello,

After several projects were I had to implement local node PDO emission and reception I've had a look at how to implement it properly inside the library.

This PR is more of a discussion right now (maybe there is a better way of doing this ?) as I am not really sure of the way to go.

Right now my main goal is the following : 

- Synchronous TPDOs should automatically be sent on sync reception
- If internal object dictionary is updated (via SDO or application) then the updated data should be propagated inside PDO maps accordingly
- The same thing goes for RPDOs : on reception, internal OD data should be updated.
- I'd also like to implement some of the NMT behavior such as PDO sending only whilst the device is in OPERATIONAL

I'm leaving aside dynamic PDO configuration as I don't really have use for this right now.

Right now, I don't really like the fact that Map values are decoupled from the object dictionary and we have to use callbacks for copying data between each other, I feel like they should point to the same data.

Anyway if you feel like sharing ideas it would be greatly appreciated ! :)

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -33,7 +33,7 @@ Callback = Callable[[int, bytearray, float], None]
 class Network(MutableMapping):
     """Representation of one CAN bus containing one or more nodes."""
 
-    def __init__(self, bus=None):
+    def __init__(self, bus: can.BusABC | None = None):
         """
         :param can.BusABC bus:
             A python-can bus instance to re-use.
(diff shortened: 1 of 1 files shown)


acolomb:
Just wanted to leave a note, this PR has not been overlooked.  I currently don't have much time available to work with this library, so probably won't get a chance to review and test soon :-(

PR #342, opened 2023-01-05
 Remove old callbacks if re-adding a node with the same id

samsamfire:
Adding a node with the same ID that was previously added on the network would not update the can message callbacks (SDO, NMT, etc..) to the new node object, keeping the old node object callbacks.
This fixes the issue by first checking if the id already exists and removing the callbacks if any.


acolomb (approved):
Looks sensible.

PR #343, opened 2023-01-15
Add custom eds storage format & block transfer mechanism

samsamfire:
Hi,

CiA defines objects 0x1021 for storing EDS and object 0x1022 for storing the format used (which suppose to be mandatory according to spec btw).
I've added two optional parameters to add_node : 
- block_transfer to enable or disable block transfer 
- eds_format_handler, callable that takes sdo io and returns an eds file handle. The point of this is so that anyone with a specific format can handle their own logic.
I've also added some documentation , an example and description for parameters.

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -172,7 +172,7 @@ def import_from_node(node_id, network,block_transfer =False,eds_format_handler =
     :param int node_id: Identifier of the node
     :param network: network object
     :param block_transfer: use block transfer
-    :param eds_format_handler: Callable with custom logic that should retunr the EDS
+    :param eds_format_handler: Callable with custom logic that should return the EDS
     """
     # Create temporary SDO client
     sdo_client = SdoClient(0x600 + node_id, 0x580 + node_id, objectdictionary.ObjectDictionary())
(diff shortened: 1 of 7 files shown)


af-silva:
During the weekend will try to give it a look.

samsamfire:
Bump have you had a chance to look at it ?

canopen/objectdictionary/eds.py
```diff
@@ -167,21 +167,39 @@ def import_eds(source, node_id):
...
+            # Custom format handler must return a fp or string to extracted EDS file
+            eds_fp = eds_format_handler(eds_raw_fp,storage_format)
+        else:
+            if storage_format != 0:
+                logger.warning("Trying to read EDS with ASCII format (0) even though storage format is not 0")
+            eds_fp = sdo_client.open(0x1021, 0, "rt", block_transfer=block_transfer)
+        od = import_eds(eds_fp,node_id)
+
     except Exception as e:
+        print(e)
```
af-silva, line 202: remove the print

af-silva (commented):
Sorry for the late late response, real live took the best of me. Just some small comments.
The remaining added functionality seems ok. In terms of use I can't test it, so I can't validate. Have you test it using real hardware and everything behaves as it should?

PR #345, opened 2023-02-22
Fixed HighLimit and LowLimit for SIGNED values in EDS

friederschueler:
These values have been interpreted as unsigned values and therefore valid values raised warnings.

### DIFF
--- a/test/test_eds.py
+++ b/test/test_eds.py
@@ -4,6 +4,7 @@ import canopen
 
 EDS_PATH = os.path.join(os.path.dirname(__file__), 'sample.eds')
 
+
 class TestEDS(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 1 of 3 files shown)


af-silva (commented):
Hi, thanks for the PR.
I know this library now supports python 3.x+, but I would drop the type hints for sake of uniformization with the rest of the files, since the new functions are the only ones using it.

test/test_eds.py
```diff
@@ -47,6 +48,20 @@ def test_record(self):
...
 
+    def test_record_with_limits(self):
+        int8 = self.od[0x3020]
+        self.assertEqual(int8.min, 0)
+        self.assertEqual(int8.max, 127)
+        uint8 = self.od[0x3021]
+        self.assertEqual(uint8.min, 2)
+        self.assertEqual(uint8.max, 10)
+        int32 = self.od[0x3030]
+        self.assertEqual(int32.min, -2147483648)
```
lamuguo, line 59: int32.min should be -1

friederschueler: You are correct. I don't know why I didn't see it. This should be the correct function:

`def _signed_int_from_hex(hex_str, bit_length):
    number = int(hex_str, 0)
    max_value = (1 << (bit_length - 1)) - 1
    
    if number > max_value:
        return number - (1 << bit_length)
    else:
        return number`

I will commit a full fix with correct tests in the evening.

PR #347, opened 2023-02-22
Fix typo in for Baudrate in DeviceComissioning section

thesummer:
According to to CiA 306 version 1.3.0 Section _5.3.3.3 Network variable values_ the field name is `Baudrate` (with small **r**)

### DIFF
--- a/doc/profiles.rst
+++ b/doc/profiles.rst
@@ -66,7 +66,7 @@ class :attr:`.state` attribute can be read and set (command) by a string::
     # command a state (an SDO message will be called)
     some_node.state = 'SWITCHED ON'
     # read the current state
-    some_node.state = 'SWITCHED ON'
+    some_node.state
 
 Available states:
 


acolomb:
Looks good in principle, however we should think about a migration plan or something.  Users relying on the previous spelling might have existing EDS files stop working, no?  Or is the whole parsing case-insensitive?

thesummer:
> Looks good in principle, however we should think about a migration plan or something. Users relying on the previous spelling might have existing EDS files stop working, no? 

EDS files should be fine. This only affects DCF files. The current behavior is that a standard conformant DCF file will cause a runtime error.

> Or is the whole parsing case-insensitive?

That I don't know. This library at least is case sensitive.

thesummer:
Just checked the standard:

> The keyname is not case sensitive.

acolomb:
Then could you maybe check the code and make sure it is handled case-insensitive here as well, as part of the PR?

thesummer:
@acolomb I had a short look at it. What I don't understand is why this error happens at all. Accoridng to the docu of `configparser` the keys should be read case insensitive, but here: https://github.com/christiansandberg/canopen/blob/master/canopen/objectdictionary/eds.py#L90 the case of the key seems to matter.

Do you have any suggestion for this. I am not that familiar with the configparser library.

dirtcrusher:
I found why the error is happening @thesummer.

By setting the `optionxform` method of the ConfigParser to the `str` method [here](https://github.com/christiansandberg/canopen/blob/v2.1.0/canopen/objectdictionary/eds.py#L23), we explicitly change the ConfigParser to be case sensitive.

(This is even mentioned as an usage example of the `optionxform` method [in the docs](https://docs.python.org/3.10/library/configparser.html#configparser.ConfigParser.optionxform).

PR #350, opened 2023-03-14
objectdictionary: Use node_id from DCF if not provided

thesummer:
Hi,

When creating a RemoteNode from the objectdictionary, like
```
node = canopen.RemoteNode(None, object_dictionary=config_file)
```
then the variables relative to the `NODEID` are not resolved properly, because the `None` value in `node_id` is not updated before `build_variable` is called.

This should fix that and still keep the user provided node_id, if provided.

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -87,8 +87,9 @@ def import_eds(source, node_id):
                 pass
 
     if eds.has_section("DeviceComissioning"):
-        od.bitrate = int(eds.get("DeviceComissioning", "BaudRate")) * 1000
+        od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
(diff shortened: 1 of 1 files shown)


thesummer:
Any chance to merge this into the official sources or is there something missing?

acolomb (commented):
The case change here seems to be related to your PR #347, but the discussion there wasn't finished IIRC.  If the parser really is case-sensitive, this has the potential to break people's code.  So can we implement a backwards-compatible solution?  Given that the EDS spec seems to be case-insensitive anyway.

The node ID change seems fine.

thesummer:
My bad. I only wanted to fix the node_id thing. Must have messed up the commit. I will update it in the coming days.

acolomb:
@thesummer Are you still going to work on this? Seems like a worthwhile improvement, just got hung up on minor details.

acolomb (approved):

PR #352, opened 2023-03-15
Add missing aborts and typehints for sdoclient

samsamfire:
Hi,
I found some missing aborts whilst doing some tests:
in particular, an sdo abort should be sent when a timeout occurs (0x05040000)
an abort should also be sent if the expected command value is not correct (0x05040001).

I also added typehints were possible.

Also, have any of you considered to add a static type checker (mypy,...) and a linting tool ? Could be good in the future

### DIFF
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -3,16 +3,16 @@ from typing import Dict, Union
 
 from .base import BaseNode
 from ..sdo import SdoServer, SdoAbortedError
-from ..pdo import PDO, TPDO, RPDO
+from ..pdo import PDO, TPDO, RPDO, Map
 from ..nmt import NmtSlave
 from ..emcy import EmcyProducer
+from ..sync import SyncProducer
 from .. import objectdictionary
 
 logger = logging.getLogger(__name__)
 
 
 class LocalNode(BaseNode):
-
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


samsamfire:
reverting the PDO change that has nothing to do with this PR

acolomb:
@samsamfire Sorry for not responding for such a long time.  I've tried to grasp your changes here in order to see if there's merit in merging it for the next release.  However, it is quite convoluted between actual changes and formatting.  Would you mind separating those into several smaller PRs or at least split into independent commits?

acolomb:
Keeping this open because the functional change regarding SDO aborts could be worthwhile.  But as is, this is too convoluted to act upon.  Someone should pick up the gist of it and redo another, focused PR.

sveinse:
I think this makes sense. Once the formatting and type hints are shaved off, the changes aren't really that big. Getting them stripped down to pure functional changes should be done, yes.

bizfsc:
Had another look at it and vote close now. The other changes are all related to #652 and the dynamically typed value pipeline.

acolomb:
Actually some EDS limit value checking and a small doc fix is still worthwhile in this pile of changes.  So I'm still leaving it open for future reference until these are incorporated.

PR #356, opened 2023-03-25
Use with contexts for open

sveinse:
This PR adds `with` contexts to the usage of open. A context handles closing automatically, making the call to close superfluous. 

Ths PR also fixes issue with close in objectdictionary/__init__.py and eds.py. `export_od()` and `import_eds()` should not close file handles that is not opened within these functions.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -48,6 +48,10 @@ def export_od(od, dest:Union[str,TextIO,None]=None, doc_type:Optional[str]=None)
         from . import eds
         return eds.export_dcf(od, dest)
 
+    # If dest is opened in this fn, it should be closed
+    if type(dest) is str:
+        dest.close()
+
 
 def import_od(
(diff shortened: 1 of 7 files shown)


acolomb (commented):
Looks good overall from a quick glance. I guess if people relied on these functions closing their file handle for them, that's rather weird, but was it documented? I can't imagine many cases where it would matter, but let's hear some more thoughts on whether this behavior change needs a bit of a migration path to avoid breaking existing code on upgrade.

christiansandberg (approved):

canopen/objectdictionary/__init__.py
sveinse, line 1: A small snag: Since this PR was created before the `isinstance()` fix was created, this statement should ideally be changed to (in canopen/objectdictionary/__init__.py)

```py
    # If dest is opened in this fn, it should be closed
    if isinstance(dest, str):
        dest.close()
```

Issue #355, opened 2023-03-25
Deprecate attribute based IO operations, like `v = var.raw`


sveinse:
I'd like to propose to deprecate the uses of attribute based operations, e.g.  `value = var.raw` or `var.raw = 32` throughout canopen. The problem with the existing implementation is that some of these attr getters and setters end up in IO operations towards the CAN system. This makes it very opaque if the attr ops are "local", i.e. fetching data from the class, vs. doing remote operations. A more explicit API will make it easier to understand when IO happens and it makes the code more portable.

I have created a PR-ready proposal here:

https://github.com/sveinse/canopen-asyncio/tree/feature-deprecate-attr-ops

All attribute getters and setters in canopen are replaced with explicit methods, .e.g. `var.get_raw()` and `var.set_raw()`. The proposal doesn't alter or break the existing attr-based API, but it prints a deprecation warning if used.

I am well aware and understand that this proposal might be controversial. There are advantages to keeping the attr based operations. Its simple and it allows good duck-typing for e.g. local and remote nodes and it has been this way in canopen for a long time. However, this methodology is very implicit. It is difficult to spot when an attr operation results in blocking IO. This type of change will become necessary when porting to async, since async getters and setters is not a thing.

Should I proceed with creating a PR for this change, or is there need to discuss the topic?

acolomb:
I'm strongly against creating methods named get_something() and set_something() when we do have proper properties in Python. The only understandable exception to that rule I see is when wrapping a library API to keep the names compatible, although a good wrapper will even then provide properties to shorten the code calls.

I do understand your concern, but would like to keep the properties for the most simple, readable way to express what is being done, still in a synchronous fashion. The fact that it involves blocking for a response is something that I suppose most users actually welcome, such as used in a script. The performance oriented use cases where thought has been put into how concurrent operations cooperate is the advanced, less common way where I would rather see any form of methods with longer names.

sveinse:
That's fine for me, I understand that the proposal meddles with principle architectural decisions.

If canopen decides to include asyncio support, the regular synchronous mechanism of using attr getters/setters cannot be used. Hence, the effect of this choice will be that the regular sync use (e.g. `v = param.phy`) will have be changed to other methods for async use (e.g. `v = await param.aget_phy()`). Other dual async/regular libraries tries to keep consistency between the variants and this was my suggestion to bridge this gap. That's all.

Issue #284, opened 2021-11-13
Unprotected callbacks from CAN RX thread


sveinse:
In my work of porting canopen to asyncio, I stubled across that canopen contains a lot of unprotected thread crossings. By and large all callbacks coming from can through `Network.notify()` run in another thread. This means that all subscribed callbacks must protect any class members access, otherwise interesting results may arise. Many callback handlers have synchronization mechanisms, but there a still quite a few that does not.

```
emcy.py:EcmyConsumer.on_emcy()  Callbacks are not protected
network.py:NodeScanner.on_message_received()  self.nodes is not protected
nmt.py:NmtBase.on_command()  self._state is update and not protected
nmt.py:NmtSlave.on_command()  self.update_heartbeat() is called which alters state
profiles/p402.py:BaseNode402.on_TPDOs_update_callback()  self.tpdo_values[] updated from thread
sdo/server.py:SdoServer:on_request()  Modifies lots of variables
```

I haven't investigated how many of these have an implicit safe design. This issue is mostly for information, but in general race conditions due to missing synchronization protection is sporadic and hard to debug.

acolomb:
Can you give an example of what problems may arise of this?

AFAIK many operations in Python (e.g. dict modifications) are inherently thread-safe.  Each access sees a consistent state either from before another thread's modification, or with it applied atomically.  So what do we need to "protect" from?

And what measures would you suggest?  Adding mutex locks in many places increases the complexity, potential for deadlocks, and possibly hurts performance.  So such a thing must be well thought through.

sveinse:
As mentioned, I haven't investigated how many of these thread crossings are implicitly safe by design, e.g. write-only from RX thread, read-only from main thread and so on. I found approx twice the amount of thread crossings in total than listed above, so the majority were inherently safe. E.g. using `threading.Condition` or `queue.Queue` and similar.

Python itself with the GIL make a lot of the py operations atomic, which is great. However, in general, one typical pitfall might be if one reads a class member attribute into a local variable, mutates it and writes it back. This execution might be interrupted mid flight and the RX thread might get the GIL and write that class attribute. The first code resumes controls writes back the locally cached data, overwriting the new data from the RX thread.

One example might be `nmt.py:NmtBase.on_command()` and `.state()`, where both read `self._state` one or more times and then later write back the result. Each of these two can run in two different threads and as such race to set `self._state`. 

[...]

PR #359, opened 2023-03-27
Add asyncio support

### LINKED ISSUE #272: Add async support to canopen
I have a use case where I have a need to control many independent CAN nodes with separate controls. In order to do that one could either do regular CAN callbacks and state machines, or do them concurrently with an async framework in Python. After investigating a bit, the `python-can` package does support async operations and I think that `canopen` should get async support. The biggest upside to concurrent async programming is that the async user code is linearized as if using regular blocking functions, and it makes coding and maintenance much simpler.

I have started on the concept work in my own private fork using `asyncio`, linked here https://github.com/sveinse/canopen-asyncio/tree/feature-asyncio. For a peek I have a simple example code at https://github.com/sveinse/canopen-asyncio/blob/feature-asyncio/examples/canopen_async.py. This specific example won't work out of box, but a modified variant of the same script with real PDOs and SDOs is tested against physical CANopen HW devices.

I believe it should be rather feasible to add `asyncio` support alongside the existing API, which is what I've started on in my branch. I'd argue strongly for releasing it as one package/repo due to the constant risk of getting any asyncio fork out of sync with the official upstream canopen package.

This work is currently WIP. I hope this work could be considered for the official upstream `canopen`. I had hoped we could use this issue as a place to discuss async support. A warning is in order: My motivation in this is a specific use case for a very specific CAN system diagnostics tool and I plan to take it as far as that task requires. The system doesn't implement all of canopen and thus there are many functions in the canopen package I will never encounter. Hopefully this work may become a start point that can inspire others to continue developing it.

sveinse:
This PR adds support of asyncio to canopen. The overall goals is to make canopen able to be used in _either_ with asyncio or regular synchronous mode (but not at the same time) from the same code base.

Note that this work is still work in progress. This PR was created to discuss the specific solutions for async and non-async as mentioned in #272. This PR closes #272.

Current status until feature complete:

- [x] ~Implement `ABlockUploadStream`, `ABlockDownloadStream` and `ATextIOWrapper` for async in `SdoClient`.~ Not needed
- [x] Implement `EcmyConsumer.wait()` for async
- [x] ~Async implementation of `LssMaster`~ Only `fast_scan`
- [x] ~Async implementation of `BaseNode402`~Omitted for now
- [x] Implement async variant of `Network.add_node()`
- [x] Update unittests for async
- [x] Update examples
- [ ] Update documentation

### Merging

This PR won't be merged directly as it is very big. Instead the asyncio feature will be split up into multiple smaller PRs. This PR will be kept up-to-date with the changes, so its possible to use this to see where this is going.

List of sub-PRs:

- [ ] #676


sveinse:
What is the best way to deal with `Variable` attributes? The feedback from #355 indicates that it is desirable to keep the attr based get/set mechanisms. However, this scheme cannot be used for async as there is no await mechanism built into `@data.setter`.

```python
    var = param.data   # Non-async use
    var = await param.aget_data()   # Async use
```

My goal has been to keep the async and non-async uses of canopen as equal as possible, but this is an area where users will see a difference.

acolomb:
There is an implementation for `.read()` and `.write()` on the underlying `canopen.variable.Variable` type. Could that be used in an async wrapper, or something similar added for async?

sveinse:
@acolomb I'm not precisely sure what you mean, so let me guess: You need separate `async` methods for read and right and they should only call their respective async setters/getters.

acolomb:
Sorry, I was trying to answer your question in the previous comment about synchronous getters / setters used in the properties. What I meant was that instead we already do have methods to access a variable remotely. Those might be a better fit to mirror to the async world:

```python3
value = sdo_var.raw  # This is the usual, terse style recommended in the docs
value = sdo_var.read(fmt='raw')  # This also works already
value = await sdo_var.aread(fmt='raw')  # Feels like a natural enhancement for async
```

Note that the `fmt='raw'` argument can of course be left out, it's only needed for `phys` or `desc` variants.  This would give us an easy way to support async without changing the existing synchronous property-based access.

sveinse:
> Sorry, I was trying to answer your question in the previous comment about synchronous getters / setters used in the properties. What I meant was that instead we already do have methods to access a variable remotely. Those might be a better fit to mirror to the async world:
> 
> ```python
> value = sdo_var.raw  # This is the usual, terse style recommended in the docs
> value = sdo_var.read(fmt='raw')  # This also works already
> value = await sdo_var.aread(fmt='raw')  # Feels like a natural enhancement for async
> ```
> 
> Note that the `fmt='raw'` argument can of course be left out, it's only needed for `phys` or `desc` variants. This would give us an easy way to support async without changing the existing synchronous property-based access.

Sorry for late reply here. All of these methods are supported in the async port. In addition to `await sdo_var.aget_raw()` as mentioned.  What if all the getter and setters-like functions and only rely on `aread()` and `awrite()` for async? Interesting. I need to contemplate that.

The areas of challenge for async currently are:
* `Variable`: `.data`, `.raw`, `.phys`, `.desc`, `.bits`
* `BaseNode402`: `.op_mode`, `.statusword`, `.controlword`, `.state`
* `PdoMap.read()` - Because it contains logic among blocking IO. I'm not sure if my dual sync-async solution is elegant or if its smelly.
* `SdoClient` - is an ongoing challenge. The `*Stream()` classes contains considerable logic and relies on `RawIOBase` mixin, all synchronous code. I'd rather not need to replicate all those methods and functions

sveinse:
This branch is ready for review and integration into the project. Its not fully complete (I suppose it never will). The biggest lack is that there is no documentation for async.

The branch's README (https://github.com/sveinse/canopen-asyncio?tab=readme-ov-file#difference-between-async-and-non-async-version) contains an overview over the current differences between this branch and the main project.

The port is quite "naive" async-wise as it relies on `asyncio.to_thread()`  on many async operations, which not ideal. The back-end of the canopen library relies on synchronous callbacks which doesn't work too well from an async perspective. To be able to support both non-async (regular) and async use, this approach have been chosen. In time I hope we can refactor the inner working of the library such that we can do it independent of the communication (e.g. vis sansio).

canopen/network.py
```diff
@@ -28,20 +30,24 @@ class Network(MutableMapping):
     NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration.
     NOTIFIER_SHUTDOWN_TIMEOUT: float = 5.0  #: Maximum waiting time to stop notifiers.
 
-    def __init__(self, bus: Optional[can.BusABC] = None):
+    # NOTE: Function arguments changed to provide notifier, see #556
```
acolomb: As discussed in #556, this can simply be set between `__init__()` and `connect()`, with the latter only creating one if not already set (as you did).

canopen/network.py
```diff
@@ -28,20 +30,24 @@ class Network(MutableMapping):
...
+                 loop: Optional[asyncio.AbstractEventLoop] = None):
         """
         :param can.BusABC bus:
             A python-can bus instance to re-use.
         """
         #: A python-can :class:`can.BusABC` instance which is set after
         #: :meth:`canopen.Network.connect` is called
-        self.bus = bus
+        self.bus: Optional[can.BusABC] = bus
+        self.loop: Optional[asyncio.AbstractEventLoop] = loop
```
acolomb: The type hints here are inferred from the arguments, better skip repeating them.

canopen/network.py
```diff
@@ -52,6 +58,12 @@ def __init__(self, bus: Optional[can.BusABC] = None):
 
         self.lss = LssMaster()
         self.lss.network = self
+
+        # Register this function as the means to check if canopen is run in
+        # async mode. This enables the @ensure_not_async() decorator to
+        # work. See async_guard.py
+        set_async_sentinel(self.is_async())
```
acolomb: This is not "registering a function", but simply flipping a switch concerning the current thread ID, right?  The comment should be adapted, maybe leftover from a previous design?

canopen/network.py
```diff
@@ -260,6 +344,10 @@ def check(self) -> None:
                 logger.error("An error has caused receiving of messages to stop")
                 raise exc
 
+    def is_async(self) -> bool:
```
acolomb: Should be a `@property`.

canopen/network.py
```diff
@@ -243,11 +294,44 @@ def notify(self, can_id: int, data: bytearray, timestamp: float) -> None:
...
+            self.dispatch_callbacks(self.subscribers[can_id], can_id, data, timestamp)
         self.scanner.on_message_received(can_id)
 
+    def on_error(self, exc: BaseException) -> None:
+        """This method is called to handle any exception in the callbacks."""
+
+        # Exceptions in any callbaks should not affect CAN processing
+        logger.exception("Exception in callback: %s", exc_info=exc)
+
+    def dispatch_callbacks(self, callbacks: List[Callback], *args) -> None:
```
acolomb: This is one of the central elements it seems, making it easy to influence what exactly "invoking a callback" means and does.  The approach is alright and the Network seems like a good place to handle such a definition centrally.

However, I'm unsure whether this is the best / only solution regarding callbacks.  One alternative that springs to mind is to wrap each given callback function when it is registered somewhere, replacing it with a callable that creates the corresponding task instead of executing the given callback directly.

canopen/network.py
```diff
@@ -243,11 +294,44 @@ def notify(self, can_id: int, data: bytearray, timestamp: float) -> None:
...
+            try:
+                if (exc := task.exception()) is not None:
+                    self.on_error(exc)
+            except (asyncio.CancelledError, asyncio.InvalidStateError) as exc:
+                # Handle cancelled tasks and unfinished tasks gracefully
+                self.on_error(exc)
+
+        # Run the callbacks
+        for callback in callbacks:
+            result = callback(*args)
```
acolomb: This will result in any non-coroutine callbacks to be handled immediately, regardless of how long they will run, right?  Given all the effort you're putting into failing early of some action would block the main loop, this seems like a pretty big hole in that protection, no?

canopen/network.py
```diff
@@ -335,6 +423,7 @@ def stop(self):
         """Stop transmission"""
         self._task.stop()
 
+    # @callback  # NOTE: Indirectly called from another thread via other callbacks
```
acolomb: Is this supposed to become an actual decorator later?

The comment is not really clear to me.  How can we know what thread or other context this will be called from?

canopen/network.py
```diff
@@ -243,11 +294,44 @@ def notify(self, can_id: int, data: bytearray, timestamp: float) -> None:
             Timestamp of the message, preferably as a Unix timestamp
         """
         if can_id in self.subscribers:
-            callbacks = self.subscribers[can_id]
-            for callback in callbacks:
-                callback(can_id, data, timestamp)
+            self.dispatch_callbacks(self.subscribers[can_id], can_id, data, timestamp)
         self.scanner.on_message_received(can_id)
 
+    def on_error(self, exc: BaseException) -> None:
```
acolomb: I think that python-can already handles exceptions in the receiver thread gracefully, doesn't it?  Anyway, this is probably a sane mechanism to build.

canopen/nmt.py
```diff
@@ -119,15 +123,17 @@ def __init__(self, node_id: int):
         self.state_update = threading.Condition()
         self._callbacks: List[Callable[[int], None]] = []
 
+    # @callback  # NOTE: called from another thread
+    @ensure_not_async  # NOTE: Safeguard for accidental async use
```
acolomb: I'd prefer not repeating this comment on every usage.  If that's what the function does, it's best to keep that in the docstring and rely on IDE features to show the function's short description as needed.  Otherwise, the function / decorator name should speak for itself.

canopen/nmt.py
```diff
@@ -147,28 +156,42 @@ def send_command(self, code: int):
...
         self.network.send_message(0, [code, self.id])
 
+    @ensure_not_async  # NOTE: Safeguard for accidental async use
     def wait_for_heartbeat(self, timeout: float = 10):
         """Wait until a heartbeat message is received."""
+        # NOTE: Blocking lock
         with self.state_update:
             self._state_received = None
+            # NOTE: Blocking call
             self.state_update.wait(timeout)
```
acolomb, line 159: Am I understanding correctly that this is the central point here where an asyncio-native awaitable should be used instead?

canopen/nmt.py
```diff
@@ -147,28 +156,42 @@ def send_command(self, code: int):
...
         if self._state_received is None:
             raise NmtError("No boot-up or heartbeat received")
         return self.state
 
+    async def await_for_heartbeat(self, timeout: float = 10):
```
acolomb, line 164: Isn't there a way to wrap one function in order to generate the other?  Repeating arguments, default values, type hints, and docstrings seems very error-prone.

I understand that an async function can never be made by wrapping a regular (blocking) function.  But the other way round seems doable, basically moving the functionality to an async implementation, then running that inside its own event loop in the synchronous API function?

Otherwise, if punting off to a background thread is acceptable behavior, why not simply let the caller do it?  Why would we need to duplicate many API methods if the same result can be achieved by just calling `await asyncio.to_thread(node.nmt.wait_for_heartbeat, timeout)` from the application code?

canopen/nmt.py
```diff
@@ -228,7 +252,12 @@ def send_command(self, code: int) -> None:
         # The heartbeat service should start on the transition
         # between INITIALIZING and PRE-OPERATIONAL state
         if old_state == 0 and self._state == 127:
-            heartbeat_time_ms = self._local_node.sdo[0x1017].raw
+            # FIXME: Document why this was fixed
```
acolomb: Please do.

canopen/variable.py
```diff
@@ -33,9 +34,15 @@ def __repr__(self) -> str:
     def get_data(self) -> bytes:
         raise NotImplementedError("Variable is not readable")
 
+    async def aget_data(self) -> bytes:
```
acolomb, line 38: May just as well be me overlooking something major.  But why do we need to add this on the generic `Variable` implementation level?  Isn't the only protocol which might block on variable data transfer the SDO client implementation?  `PdoVariable` updates the message content in-place.  The `SdoServer` class is not even using the `SdoVariable`, but accesses its own data store.  Since the property based access is a bad fit for an async-aware API, there seems to be little value in forcing all derived classes to implement coroutine methods to access the data.  Since these methods just serve as a common base for the properties implemented in the generic `Variable` class.

canopen/variable.py
```diff
@@ -75,7 +82,14 @@ def raw(self) -> Union[int, bool, float, str, bytes]:
         Data types that this library does not handle yet must be read and
         written as :class:`bytes`.
         """
-        value = self.od.decode_raw(self.data)
+        return self._get_raw(self.get_data())
+
+    async def aget_raw(self) -> Union[int, bool, float, str, bytes]:
```
acolomb: Just a roadside thought...  We had discussed how properties are not a good fit for async API design, as they cannot be declared `async`.  But how about a property getter returning a coroutine?  Wouldn't that work just as well?

```python3
@property
def raw_async(self):
    async def _get_raw():
        ...
    return _get_raw

raw_value = await var.raw_async
```

canopen/variable.py
```diff
@@ -143,6 +185,16 @@ def read(self, fmt: str = "raw") -> Union[int, bool, float, str, bytes]:
         elif fmt == "desc":
             return self.desc
 
+    async def aread(self, fmt: str = "raw") -> Union[int, bool, float, str, bytes]:
+        """Alternative way of reading using a function instead of attributes. Async variant."""
```
acolomb, line 189: Here I really don't understand why the docstring on `.read()` says "May be useful for asynchronous reading", yet this method plainly reimplements the switching logic without ever using the former method.  The main difference is that this method is a callable, while the properties are not.  But that alone is not very useful for asynchronous usage, as you are proving by the mirrored implementation.

canopen/variable.py
```diff
@@ -163,11 +215,24 @@ def write(
...
+        elif fmt == "desc":
+            await self.aset_desc(value)  # type: ignore[arg-type]
+
 
 class Bits(Mapping):
 
+    @ensure_not_async  # NOTE: Safeguard for accidental async use
     def __init__(self, variable: Variable):
         self.variable = variable
+        # FIXME: This is not compatible with async
```
acolomb: Not compatible because the constructor cannot be a coroutine, right?  We basically need another adapter class `BitsAsync` or a function that returns a coroutine / Future which resolves to the `Bits` object here.

I agree that the architecture here is very async-unfriendly.  We basically have a proxy object which wraps some data conversion around the actual data accessing `Variable` object.  Neat, but hard to port into async.  This seems to be one of the core points of difficulty, thus I'd ask what the API consumer side would look like in an application designed around asyncio from the ground up?

examples/canopen_async.py
```diff
@@ -0,0 +1,69 @@
...
+    while True:
+        i += 1
+
+        # Wait for PDO
+        t = await node.tpdo[1].await_for_reception(1)
+        if not t:
+            continue
+
+        # Get TPDO value
+        # PDO values are accessed non-synchronously using attributes
```
acolomb: You mean non-blocking?

examples/canopen_async.py
```diff
@@ -0,0 +1,69 @@
...
+        state = node.tpdo[1]['state'].raw
+
+        # If state send RPDO to remote
+        if state == 5:
+
+            await asyncio.sleep(0.2)
+
+            # Set RPDO and transmit
+            node.rpdo[1]['count'].phys = i
+            node.rpdo[1].transmit()
```
acolomb, line 45: Where will this transmission actually be handled?  I think it's a queue of messages to transmit, handled by a background thread in python-can?  Or is there actually some mechanism on that level which tries to integrate the sending / buffering retry loop into the asyncio event loop?  That would seem like a worthwhile use of asyncio.

examples/canopen_async.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+async def amain():
+
+    # Create the canopen network and connect it to the CAN bus
+    loop = asyncio.get_running_loop()
+    async with canopen.Network(loop=loop).connect(
+        interface='virtual', bitrate=1000000, recieve_own_messages=True
```
acolomb:
```suggestion
        interface='virtual', bitrate=1000000, receive_own_messages=True
```

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+either async or non-async environment, with suitable API for both.
+
+To minimize the impact of the async changes, this port is designed to use the
+existing synchronous backend of the library. This means that the library
+uses :code:`asyncio.to_thread()` for many asynchronous operations.
+
+This port remains compatible with using it in a regular non-asyncio
+environment. This is selected with the `loop` parameter in the
+:code:`Network` constructor. If you pass a valid asyncio event loop, the
+library will run in async mode. If you pass `loop=None`, it will run in
```
acolomb, line 30: Looking at the python-can code, passing a loop is mainly interesting for the `Notifier` object.  Which seems to be an elegant way of making sure callbacks triggered by message reception are run as tasks, instead of rolling our own dispatcher solution.  Especially since an arbitrary notifier can be passed in now in latest upstream.  On the `Listener` side, the `AsyncBufferedReader` implementation actually warns about a loop parameter being passed.  This just makes me wonder whether the API of passing a loop to the `Network` constructor is going in the right direction.

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+regular blocking mode. It cannot be used in both modes at the same time.
+
+
+Difference between async and non-async version
+----------------------------------------------
+
+This port have some differences with the upstream non-async version of canopen.
+
+* Minimum python version is 3.9, while the upstream version supports 3.8.
```
acolomb: No longer - we will even bump to 3.10 soonish.  Which is good, for less difference between fork and upstream.

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+
+* The :code:`Network` accepts additional parameters than upstream. It accepts
+  :code:`loop` which selects the mode of operation. If :code:`None` it will
+  run in blocking mode, otherwise it will run in async mode. It supports
+  providing a custom CAN :code:`notifier` if the CAN bus will be shared by
```
acolomb: Also available in upstream by now, see #621.

canopen/nmt.py
```diff
@@ -119,15 +123,17 @@ def __init__(self, node_id: int):
         self.state_update = threading.Condition()
         self._callbacks: List[Callable[[int], None]] = []
 
+    # @callback  # NOTE: called from another thread
+    @ensure_not_async  # NOTE: Safeguard for accidental async use
     def on_heartbeat(self, can_id, data, timestamp):
+        new_state, = struct.unpack_from("B", data)
```
acolomb, line 125: I think this reordering has mostly been adopted in #601.

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+  multiple protocols.
+
+* The :code:`Network` class can be (and should be) used in an async context
+  manager. This will ensure the network will be automatically disconnected when
+  exiting the context. See the example below.
```
acolomb: Wouldn't this be a good chance to distinguish whether we are supposed to be running in an asyncio based execution model?  I'd like that a lot more than passing an explicit loop parameter.  Basic idea: Get a reference to the current asyncio loop in `__aenter__()`, which may influence some execution paths later.  Like creating a `Notifier` instance with the loop referenced explicitly during `.connect()` (unless a notifier was already set earlier of course).

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+
+* Most async functions follow an "a" prefix naming scheme.
```
acolomb, line 46: I find that difficult in some places.  `await_for_bootup()` looks really weird for example.  We may need to decide case-by-case what's an appropriate name, not applying the scheme blindly.  But those are minor details.

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+  E.g. the async variant for :code:`SdoClient.download()` is available
+  as :code:`SdoClient.adownload()`.
+
+* Variables in the regular canopen library uses properties for getting and
+  setting. This is replaced with awaitable methods in the async version.
+
+      var = sdo['Variable'].raw  # synchronous
+      sdo['Variable'].raw = 12   # synchronous
+
+      var = await sdo['Variable'].get_raw()  # async
```
acolomb: To minimize API duplication, this could be as simple as:

    var = await sdo['Variable'].aread()
    desc = await sdo['Variable'].aread("desc")

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+
+* Installed :code:`ensure_not_async()` sentinel guard in functions which
+  prevents calling blocking functions in async context. It will raise the
+  exception :code:`RuntimeError` "Calling a blocking function" when this
+  happen. If this is encountered, it is likely that the code is not using the
+  async variants of the library.
+
+* The mechanism for CAN bus callbacks have been changed. Callbacks might be
+  async, which means they cannot be called immediately. This affects how
+  error handling is done in the library.
```
acolomb: This sounds a lot like duplicating what python-can already provides.  Can we embrace that lower-level concept more instead of building our own dispatcher?

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+
+* The callbacks to the message handlers have been changed to be handled by
+  :code:`Network.dispatch_callbacks()`. They are no longer called with any
+  locks held, as this would not work with async. This affects:
+    * :code:`PdoMaps.on_message`
+    * :code:`EmcyConsumer.on_emcy`
+    * :code:`NtmMaster.on_heartbaet`
+
+* SDO block upload and download is not yet supported in async mode.
```
acolomb, line 76: Which would be a really good fit for a coroutine I suppose, if done right :-)

README.rst
```diff
@@ -8,6 +8,84 @@ automation tasks rather than a standard compliant master implementation.
...
+
+* :code:`ODVariable.__len__()` returns 64 bits instead of 8 bits to support
+  truncated 24-bits integers, see #436
+
+* :code:`BaseNode402` does not work with async
```
acolomb, line 78: I can imagine how a homing procedure for example would benefit from being called with `await`.

canopen/sdo/server.py
```diff
@@ -205,6 +209,22 @@ def upload(self, index: int, subindex: int) -> bytes:
         """
         return self._node.get_data(index, subindex)
 
+    async def aupload(self, index: int, subindex: int) -> bytes:
```
acolomb, line 209: Since there is no blocking access, we don't really need the async method variants on the `SdoServer` (nor `SdoBase` for that matter).  Unless `LocalNode.get_data()` is also allowed to be a blocking access pattern, implemented as a coroutine itself, there is no value in wrapping it here like this.  The only expected blocking I/O on the library side is the `SdoClient`, thus the async methods can be introduced only there.

canopen/pdo/base.py
```diff
@@ -331,58 +349,48 @@ def add_callback(self, callback: Callable[[PdoMap], None]) -> None:
...
-    def read(self, from_od=False) -> None:
-        """Read PDO configuration for this map.
-        
-        :param from_od:
-            Read using SDO if False, read from object dictionary if True.
-            When reading from object dictionary, if DCF populated a value, the
-            DCF value will be used, otherwise the EDS default will be used instead.
-        """
+    def read_generator(self):
+        """Generator to run through steps for reading the PDO configuration
```
acolomb, line 367: I like this solution.  Maybe a similar pattern can be applied in other places?

acolomb (commented):
Phew, that's a lot of material and I hope you don't mind my elaborate review comments.  But first, let's take a step back and see if we can better define what the PR does and what the goals are.

I'm not really experienced with asyncio based programming in Python, but I do know pretty well what an event loop and coroutines are.  So my main mental challenge with this is trying to imagine what an application would really look like if using asyncio to handle its functionality elegantly.  From own experience, I'd say moving long-blocking background activities from threads to a (cooperatively scheduled) task is one of the main motivations -- such as waiting for a drive homing procedure to finish.

My impression on what this focuses on, with some overall remarks about the implementation:

1. *Finding and marking all API functions potentially blocking on I/O.*  This is a useful thing in itself, but I don't think we want the NOTE comments in the code indefinitely.  It makes sense to gather findings like you did now, but provides little extra value to the library consumer and will bit-rot really fast I suppose.

2. *Detecting improper use of blocking functions in async context, which would block the main loop.*  This is useful but totally optional IMHO, so could better be moved to its own PR to focus on the actual async capabilities first.  An approach with a single `assert canopen.async_guard.ensure_not_async` line (or similar) could be just as good as the decorator approach, but clearer.

3. *Arranging user callbacks to be scheduled in the event loop instead of called immediately.*  I see several alternatives to the current dispatcher solution, so let's clarify what situation this applies to:  A CAN message arrives on the network and the application needs to passively react to this external trigger.  The reaction may be an internal callback (as in p402.py) or an arbitrarily long-running external application function.  This uncertainty must be contained to not block the main event loop.  Traditionally, it is executed inside the RX thread of python-can's `Notifier` anyway, blocking that at worst.  There is already functionality in there to schedule callbacks as tasks if they are coroutines, execute directly otherwise.  So the easiest would be for the application to provide a coroutine directly as callback -- this just needs to be passed through the canopen library, so that the listener is also an async / await wrapper around the application callback.  But shouldn't it be an application responsibility to decide whether the callback happens inside the RX background thread, or as a task on the event loop?

4. *Provide async (coroutine) variants of common blocking operation methods.*  This is the core of the "async integration" API-wise.  But it's actually much less of a concern than the callback handling.  Because the caller can always just do as you did in many places, calling `await asyncio.to_thread()` on the existing synchronous function.  Some of the really interesting functions are not converted to async variants yet.  I suppose that `to_thread()` approach is a stop-gap solution before implementing real, awaitable internal methods.

5. *Add unit tests for the new API.*  This is probably the largest portion of the diff, but I haven't checked it thoroughly as these high-level review points must be discussed and solved first.

Sorry if I missed something now, or it's just my lack of understanding asyncio in general.  Plus I get too tired when tackling such a complex and large review late at night.  But that's already lots of food for discussion I suppose.

sveinse:
Thank you for reviewing them @acolomb. You certainly did a thorough job! I'm currently on a sick leave, so I don't think I'll get any change to look at them immediately thou. This is what I propose as next steps:

1. Let me sync this branch so it aligns with the current master. I haven't done that in a very long while
2. I will review your comments and potentially make fixes where the resolution is obvious

Regarding the "NOTE" and "Blocking call". These are markers I put in the after doing an investigation of what breaks. As we've been discussing before, this library does a lot of blocking IO in unconventional calls, such as in properties, so it took some efforts to find them all. My intentions is to remove these markers before the final merge.

canopen/network.py
sveinse: Yes, that should be possible. Why the reluctance to add to the signature of `__init__()` thou? It is not altering any existing behavior.

I personally find the first more elegant that the latter:

```python
# Easy
network = Network(bus=bus, notifier=mynotifier)

# More clunky
network = Network(bus=bus)
network.notifier = mynotifer
```

canopen/network.py
sveinse: Fixed

canopen/network.py
sveinse: Done

canopen/network.py
sveinse: Fixed. I'm curious: What makes you want this to be a property? What's wrong with a method?

canopen/network.py
sveinse: I'm not sure what you precisely mean by the first comment, but the term "callback" is consistently used for any function that that handles incoming CAN messages via the notification & listener system from python-can.

The reason for adding `def dispatch_callbacks()` is that coroutine callbacks needs special steps to be called. In order to run a coroutine a task be created and the future object must be kept so the GC doesn't delete the object before ever running the callback.

canopen/network.py
sveinse: No, that's intentional. The callbacks in canopen run in the rx thread, so calling immediately doesn't affect the main loop.

The rule must be:
* Regular functions: Call them precisely the same way its done today.
* Async functions: Call them from the async event loop.

canopen/network.py
sveinse: All of these comments have been removed.

Not directly related to this PR, but I am inclined to once more mention that the thread safety handling is not strong in this library. All callbacks via. the notification & listener system is run in the RX thread of python-can. The entire call chain from those callbacks are running in different threads than the user application. This is surfacing when working with the asyncio and rx thread. Very few asyncio function s are available from other thread, so one must know in what context the current function is running.

Not intending to beat a dead horse but we discussed it in #284, but it didn't get any traction.

PS! Is the user informed that the user provided callbacks run in a different thread than the main one?

canopen/network.py
sveinse: This is not new for asyncio. `MessageListener.on_message_received()` is already handling the exception today.  I think the intention is that any canopen error should not affect any other can processing, which I think makes sense. The fix here is adding the `Network.on_error()` so it can be reused from other locations.

canopen/nmt.py
sveinse: Removed.

canopen/nmt.py
sveinse: Correct. `threading.Condition()` and `queue.Queue()` and friends cannot be used in an async function since they block.

canopen/nmt.py
sveinse: The core problem is that since `wait_for_heartbeat()` may block, it cannot be called from an async application. Instead you need a coroutine, and that requires `async def async_wait_for_heartbeat()`.

One could probably do `async def async_wait_for_hearbeat(*args, **kwargs)`, but that quickly becomes clunky with type annotations and helping the user to write the right arguments.

As mentioned elsewhere, `asyncio.to_thread()` is a stop-gap solution which we should avoid using unless we have to. The user should not need to do `await asyncio.to_thread(node.nmt.wait_for_heartbeat, timeout)`. That is way too tedious.

canopen/variable.py
sveinse: Yes, I'll have to look into that.

canopen/variable.py
sveinse: Functionally I that would work, but it's very unorthodox. I've never seen this used anywhere. You have to modify it slightly:

```py
@propery
def raw_async(self) -> Awaitable[int]:
    async def _get_raw():
        ...
    return _get_raw()
```

**edit**: I found this statement from a python core developer: https://discuss.python.org/t/async-setter-and-getter/105908/2. I like his statement:

> I usually recommend not making setters do anything “interesting”, which means they shouldn’t ever need to be async (which is “something interesting” by my definition). I don’t even like it when people add a setter to do validation (who expects x = y to raise?).

canopen/variable.py
sveinse: Simply put, the comment is wrong for asyncio. `read()` use the attributes which may lead to blocking IO via `get_data()`, which cannot be used for asyncio. Thus the need for an async call chain that relies on awatiable mechanisms which means `async def aread()` and `async def aget_data()`.

canopen/variable.py
sveinse: Correct. The IO is done in `__init__()` via the `self.read()` call. In an async setup you need to call an async variant, `self.aread()`. However that can't be done from init.

I've never used `Bits` and there i no examples with it so I don't precisely know how its supposed to be used. But perhaps something like this:

```python
# Regular mode
bits = Bits(my_variable)
bits.read()

# Async mode
bits = Bits(my_variable)
await bits.aread()
```

examples/canopen_async.py
sveinse: Fixed.

examples/canopen_async.py
sveinse: `PdoMap.transmit()` calls `network.send_message()` which immediately calls `bus.send()`. After investigating the python-can implementation there is no queue on transmit (discussed here https://github.com/hardbyte/python-can/issues/1137). The transmit happens immediately and no async mechanisms are needed.

examples/canopen_async.py
sveinse: Fixed.

README.rst
sveinse: We are _not_ passing the loop parameter to the `can.Notifier`. The reason is that we want to keep the current notifier / listener back-end callback system. We could use async callbacks if we wanted to. but that would require making a second parallel async variant of the back-end callback system, which is what I've avoided.

The `loop` parameter serves two purposes in `Network`: To set a boolean-like flag if the network is running in async mode vs. non-async mode via the `is_async` and the `Network.dispatch_callbacks()`.

README.rst
sveinse: Yes, I like that idea. When running in `__aenter__`, one is certain its in asyncio mode and that is a good place to set everything up. I'll make a fix for it.

README.rst
sveinse: Yes, let's find good names.

README.rst
sveinse: But this is more typing for the user.

README.rst
sveinse: Yes we can. It does require a refactor of the callback system if we want that. See the main comment for more.

canopen/sdo/server.py
sveinse: The idea I've attempted to be consistent about is that the async API should mirror the regular API (that's what you'd expect from an async API, right?). The `SdoServer.aupload()` implementation happen to be equal to `SdoServer.upload()` and could be removed. On the other hand `SdoClient.aupload()` is not equal to `SdoClient.upload()`. It happens to be that way because of how we've implemented `SdoServer`. Should the user need to know that when using it? If we do remove `SdoServer.aupload()` the user will have to change its implementation if `SdoServer.upload()` becomes blocking and we then need `SdoServer.aupload()`.

canopen/pdo/base.py
sveinse: Yes, this concept is called [sansio](https://sans-io.readthedocs.io/how-to-sans-io.html). This is what we need to make the SDO protocol handlers and all the other completely independent of what IO type (blocking vs async) is used. Basically its a system to put data in -> get something to send. Rinse and repeat until done.

sveinse:
Again, thank you @acolomb for the detailed review. This is a lengthy reponse, with this comment and the answers to the code comments.

I've update the PR to align with the current master and I've made cleanups and some fixes. I hope you'll find the PR less cluttered now.

I recognize that this PR is huge and it contains many different elements. Should we maybe split this PR into multiple parts?

There is an elephant in the room: Is getting this adopted into canopen the way everyone wants it is too much effort? Perhaps the need for having async support isn't that big. This can live side by side.


### Goal

The overall goal for this initiative is to add async support to canopen. With it the user shall be able to decide if the canopen library should be used with regular blocking calls or using asyncio. With async the user will be able to invoke many concurrent requests without needing to revert to threads.

#### Challenges

* How to make the listener notification / callback back-end system workable for both regular non-async and for async mode.

* How to implement application and protocol logic without ending up with duplicated non-async and async implementation.

* Naming conventions for the regular API vs the async API. Where `save()` something exists, should we have `asave` or `async_save?

* API completeness. Should the async API completely mirror the regular API, or just a select few methods?

* How to make good unit tests with both regular and async API


### In use

I'd like to mention that the asyncio port isn't untested. At work we use `canopen-asyncio` as a central development and test tool for the CAN-based products we use and make. We've used it since 2021 (!) when the port was created. These CAN networks typically contains 30-40 canopen nodes, running in parallell with other non-canopen protocols. The asyncio port have been very successful at communicating with many nodes simultaneously, such as doing SDO to different nodes while handling PDOs with others. The code in each task becomes simple to write with no event or "callback-hell" (which is the main advantage of async)


### Answers to key questions

> 2. _Detecting improper use of blocking functions in async context, which would block the main loop._  This is useful but totally optional IMHO, so could better be moved to its own PR to focus on the actual async capabilities first.  An approach with a single `assert canopen.async_guard.ensure_not_async` line (or similar) could be just as good as the decorator approach, but clearer.

Let's do this in a separate PR. When apps are done right, these guards aren't really needed. But this protection has proven to be most useful, especially when working on async vs non-async development. An incorrect blocking IO call is easy to do by mistake and finding them without the guard can be very challenging.

Which of these two approaches do you think is most clean? I assume you prefer alternative 2?

```py
# Alternative 1
@ensure_not_async
def something_to_protect():
    ...

# Alternative 2
def something_to_protect():
    assert canopen.async_guard.ensure_not_async()
    ...
```


> 3. _Arranging user callbacks to be scheduled in the event loop instead of called immediately._  I see several alternatives to the current dispatcher solution, so let's clarify what situation this applies to:  A CAN message arrives on the network and the application needs to passively react to this external trigger.  The reaction may be an internal callback (as in p402.py) or an arbitrarily long-running external application function.  This uncertainty must be contained to not block the main event loop.  Traditionally, it is executed inside the RX thread of python-can's `Notifier` anyway, blocking that at worst.  There is already functionality in there to schedule callbacks as tasks if they are coroutines, execute directly otherwise.  So the easiest would be for the application to provide a coroutine directly as callback -- this just needs to be passed through the canopen library, so that the listener is also an async / await wrapper around the application callback.  But shouldn't it be an application responsibility to decide whether the callback happens inside the RX background thread, or as a task on the event loop?

This is a key question and it has huge impact on the overall design. There are indeed several ways of doing it.

It is correct that python-can have a mechanism for calling coroutines on rx of CAN messages. If the callback, i.e. `can.Listener.__call__()` is a coroutine, it will make an async task.

However, all incoming messages are wrapped via `Network.notify()` which is a regular callback, so any async callbacks deeper in the canopen stack (e.g. in `NmtMaster`, `Emcy`) can't utilize this.

There is a way around this if each callback in canopen implements a `can.Listener` object and registers itself with with the `network.notifier.add_listener()`. This way callbacks will go directly from python can into the callback instead of a long chain `RX trhread -> can.Notifier._on_message_received() -> MessageListener.on_message_received() -> Network.notify() -> the specific callback`. It does require a considerable refactor of the callback system (which is independent of async).

> Arranging user callbacks to be scheduled in the event loop instead of called immediately.

The whole back-end listener / notification system is based on regular functions which are called immediately in the rx thread. Any user provided non-async callbacks will also be executed in the rx thread. In both async and non-async mode currently. Any coroutine callbacks, will always be executed in the event loop thread.


> 4. _Provide async (coroutine) variants of common blocking operation methods._  This is the core of the "async integration" API-wise.  But it's actually much less of a concern than the callback handling.  Because the caller can always just do as you did in many places, calling `await asyncio.to_thread()` on the existing synchronous function.  Some of the really interesting functions are not converted to async variants yet.  I suppose that `to_thread()` approach is a stop-gap solution before implementing real, awaitable internal methods.

Yes, `asyncio.to_thread()` is a stop-gap solution. It is not particularly efficient (since it spawns a thread for each call). We should strive to avoid the use of it and we should definitely not have the user need to rely on it for operating the library in async mode.

We currently need this internally because of the way the back-end listener / notification system is built and the usage of blocking locking primitives.

We could completely avoid `asyncio.to_thread()` if the notifier callback back-end were implemented with async in mind. However that will require double implementation, which is not idea. I have tested this concept. [See here](https://github.com/sveinse/canopen-asyncio/blob/experiment-asyncio-separate-sdo-client/canopen/emcy.py#L30-L66) for examples of how it might look like. Not very elegant.


> 5. _Add unit tests for the new API._  This is probably the largest portion of the diff, but I haven't checked it thoroughly as these high-level review points must be discussed and solved first.

Choosing test strategy was a choice between a rock and a hard place when making the test setup. There are (at least) two options for a dual-stack setup like this:

1. Duplicate all tests. One function for the regular functon test, one for async
2. Make every function runnable in two modes; regular and async mode

So far I've chosen the latter and the first easily makes the async and non-aync drift a part. But I'm not liking the implications the dual mode tests. It makes the unit tests very clunky and very hard to update.

README.rst
sveinse: How soon? I'm having problems with async unittesting in python 3.9 due to a bug in unittest. It would save me some work if I don't need to consider 3.9

3.9 is EOL and 3.10 is also going EOL in October 2026. Why shouldn't we bump the minimum version to 3.11?

canopen/sdo/server.py
sveinse: I did some tests to see what happens if I disable the superfluous async methods from `SdoServer`. And it didn't go so well. The SDO tests in the unit tests test against `LocalNode` that includes a `SdoServer`. They fail when testing the async SDO read/writes.

Which highlights another principal question: Should `LocalNode` or `RemoteNode` behave the same when used? For non-async they will, but they won't for async if we remove these methods.

canopen/variable.py
sveinse: I think we should leave them as-is. Since we have `get_data` and `set_data` that returns `NotImplementedError`, I think it makes sense that the async variants do too. If they aren't implemented, the call would fail, which is is the same behavior as in sync code.

canopen/variable.py
sveinse: Since the last comment, I've cleaned up this file. All the special `aget_*` and `aset_*`. Now the only async interface is `aread()` and `awrite()`, with a neat shortcut that its possible to await the `Variable` object directly to get the raw value. I hope and expect this is more in line with your expectations.

README.rst
sveinse: I've rebuilt the logic around async context (`__aenter__` and `__aexit__`). `Network` must now be used run via `async with`. The context sets up a task group which it uses to keep track of all tasks that are started while running. This takes care of error handling and cleanup.

sveinse:
I've done some considerable cleanups in the last couple of days. The footprint of the PR is much smaller, which I think aligns better with what's wanted and in the spirit of the existing canopen library.

Since the last (long) comment, I've improved the following:

* `Variable` is refactored, where async IO now goes through `aread()` and `awrite()`. The `aget_*/aset_*` API is gone.
* Made a implementation for `Bits`
* Removal of many `ensure_not_async()` guards. They were not needed. Added some user message to the rest.
* Cleanup of `Network`. No extra init args now.
* Setup async in the `async with` context handler in `Network`, including a task group to collect tasks
* Fixup the notification and callback system, including adding `Network.create_task()` which can be run from other threads (typ CAN RX tread callbacks).
* Updated all test cases - simplified and consolutated, so its more concise. Added `README.md` in test/ that describes how tests are set up.
* Updated documentation and readmes.

I hope this all makes changes that is better. I certainly feel I've done many improvements that takes it closer to what I think is right for the library.

acolomb:
Thanks for your continued effort and the recent work you put in. I'm still very much interested in a clean async variant inside this library. And I think we can find a design that minimizes awkwardness, together. I've read through your comments roughly and will engage as time permits. 

Unfortunately, I have an urgent and important project at $work right now that will keep me busy for a few more weeks. Kindly asking for patience therefore. 

As for the newer Python requirement, I'm open to bump it. But usually trying to get most smaller features and fixes merged and released first, to maximize the utility of the last release on an older Python. This has unfortunately stalled again in recent months, waiting for final review on a few PRs. If you have some spare time, help with those reviews would be appreciated, so we can minimize the amount if pending changes. And we should certainly focus on this feature then, not on modernizing type hints or other syntactic sugar which becomes possible with the newer language base.

sveinse:
In order to get some real-life test coverage of the fixes in this branch with other apps, the twin package https://pypi.org/project/canopen-asyncio/2.4.2.dev53/ have been released in a development version. Except for it using the namespace `import canopen_asyncio` it is equal.

PR #360, opened 2023-04-04
Add type annotations [WIP]

### LINKED ISSUE #358: Ambitions for type annotations in canopen
With reference to the discussion in #339 with regards to type hints, I have started looking at the work needed for type annotating canopen. What is the overall ambitions with regards to implementation of type hints? What level of hinting and static code analysis are we aiming for? What tool? mypy, flake8, pyright/pylance ?

I'm a big proponent of type annotated python. It helps development, navigation in the editor works and it shows the expected types for calls and operations and thereby reducing bugs and misunderstandings. I think adding typing to canopen is the right thing to do. The benefits outweighs the disadvantages.

Having a full compliant type annotated library will reduce the degrees of freedom in the codebase. You can do more "tricks" with python than a linter is able to understand, and thus coding for it requires a degree of code conformity. canopen relies on a lot of duck typing and dynamic types (especially variables that conditionally contain `None`). This results a quite a bit of warnings and/or code constructs to avoid the warnings. - And rightly so, as the code in many cases is not executing any type checks to verify the datatype before invoking an operation. Lastly, many of these linting tools are very opinionated on code structure, so we must expect some functional changes.

I've given a stab at implementing type annotation in canopen (see link below). I'm using mypy and pyright. After doing this I think we will have to face a few discussions on the correct and intended behavior on a few details. E.g. the use `None` as optional value.

https://github.com/sveinse/canopen-asyncio/commit/4e8925e1be30fe6c3101e4f050579dbed989ecbb

sveinse:
With reference to discussion in #358, this PR adds typing annotations to canopen.

As of today, this branch is still work in progress and the annotation is not yet complete.

mypy is the primary type checker, however as [my] development is done in VSCode, the code is linted against pyright/pylance.

Closes #358

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include test/sample.eds
 include LICENSE.txt
+include py.typed
--- a/canopen/__init__.py
+++ b/canopen/__init__.py
@@ -12,3 +12,20 @@ except ImportError:
 Node = RemoteNode
 
 __pypi_url__ = "https://pypi.org/project/canopen/"
+
+__all__ = [
+    "__version__",
+    "__pypi_url__",
+    "Network",
+    "NodeScanner",
+    "RemoteNode",
+    "LocalNode",
+    "Node",
+    "SdoCommunicationError",
+    "SdoAbortedError",
+    "import_od",
+    "export_od",
+    "ObjectDictionary",
+    "ObjectDictionaryError",
+    "BaseNode402",
+]
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -1,14 +1,19 @@
+from __future__ import annotations
+from typing import Callable, List, Optional, TYPE_CHECKING, Tuple
 import struct
 import logging
 import threading
 import time
-from typing import Callable, List, Optional
+
+if TYPE_CHECKING:
+    from .network import Network
 
 # Error code, error register, vendor specific data
 EMCY_STRUCT = struct.Struct("<HB5s")
 
 logger = logging.getLogger(__name__)
 
+TCallback = Callable[["EmcyError"], None]
 
 class EmcyConsumer(object):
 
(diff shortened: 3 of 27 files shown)


christiansandberg:
What happens if you add mypy to GitHub actions (.github/workflows/pythonpackage.yml)? Does it pass or are there more things that need fixing?

sveinse:
There are still some mypy issues remaining:

```
canopen\objectdictionary\__init__.py:527: error: Invalid index type "List[int]" for "Dict[str, List[int]]"; expected type "str"  [index]
canopen\objectdictionary\__init__.py:538: error: Invalid index type "List[int]" for "Dict[str, List[int]]"; expected type "str"  [index]
canopen\variable.py:194: error: Incompatible types in assignment (expression has type "List[int]", variable has type "range")  [assignment]
canopen\variable.py:208: error: Argument 1 to "iter" has incompatible type "Dict[str, List[int]]"; expected "SupportsIter[Iterator[int]]"  [arg-type]
canopen\pdo\base.py:100: error: Cannot find implementation or library stub for module named "canmatrix"  [import]
canopen\pdo\base.py:100: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
canopen\pdo\base.py:115: error: Incompatible types in assignment (expression has type "float", variable has type "Optional[int]")  [assignment]
canopen\pdo\base.py:117: error: Incompatible types in assignment (expression has type "float", variable has type "Optional[int]")  [assignment]
canopen\objectdictionary\eds.py:54: error: "ObjectDictionary" has no attribute "__edsFileInfo"  [attr-defined]
canopen\objectdictionary\eds.py:321: error: "Variable" has no attribute "default_raw"; maybe "default"?  [attr-defined]
canopen\objectdictionary\eds.py:322: error: "Variable" has no attribute "default_raw"; maybe "default"?  [attr-defined]
canopen\objectdictionary\eds.py:330: error: "Variable" has no attribute "value_raw"  [attr-defined]
canopen\objectdictionary\eds.py:427: error: "ObjectDictionary" has no attribute "__edsFileInfo"  [attr-defined]
Found 12 errors in 4 files (checked 28 source files)
```

These are related to `Bits`, `PdoBase.export()` which depend on unknown `canmatrix`, and the injection of `__edsFileInfo` attr in `objectdictionary.eds.import_eds()`. Ref my long comment in #358 , these are questions I need some guidance/discussion on how to solve.

It is probably a great idea to incorporate mypy checks into GH actions workflow, but it is not yet implemented.

sveinse:
I don't plan to implement this PR. Instead the type annotations will be split up into multiple PRs of suitable chunks which makes it easier to review. Closing this PR.

Issue #363, opened 2023-04-06
Confusing Variable class names used for different purposes


sveinse:
After spending quite many hours in the codebase, there is one thing I find really confusing and time consuming: Some classes are named the same thing, but with different functions. Applies to `Variable` and partly to `Array` and `Record`. 

```py
    variable.Variable
    pdo.base.Variable  # Inherited from variable.Variable
    sdo.base.Variable  # Inherited from variable.Variable
    objectdictionary.Variable
```

I understand the functions for all of these. and I get it if the intention was to make a duck-type like class, but they're really not all compatible. It does not help that type checkers and linters often displays the base name of the class, not the full path. If the type checker report an error where it expects `Variable`, a small investigation is required to figure out which of these classes is the one that should be used.

I would like to suggest to change the class names into unique names as follows:

```py
    class Variable:  # in variable.py
    class PdoVariable(Variable):
    class SdoVariable(Variable):
    class ODVariable:
```
Likewise for `Array` into `class SdoArray` and `class ODArray`, and `Record` into `class SdoRecord` and `class ODRecord`. I don't intend any changes to the methods and the class inheritance. If we need to keep the old names intact for compatibility, we can do so as aliases in the respective files. I can spin up a PR if there's interest for this.

acolomb:
I wouldn't mind as long as the aliases are still available to not break other code.

PR #368, opened 2023-04-07
Rename Variable, Record and Array (#363)

sveinse:
To create better distinction between the different Variable, Record and Array types used for different purposes. Helps development, as IDE/linters often only display base name of class. Discussed in issue #363 

Unittests pass and generated sphinx documents have been skimmed over.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -180,7 +180,7 @@ class Map:
         #: Ignores SYNC objects up to this SYNC counter value (optional)
         self.sync_start_value: Optional[int] = None
         #: List of variables mapped to this PDO
-        self.map: List["Variable"] = []
+        self.map: List["PdoVariable"] = []
         self.length: int = 0
         #: Current message data
         self.data = bytearray()
@@ -214,7 +214,7 @@ class Map:
         raise KeyError('{0} not found in map. Valid entries are {1}'.format(
             value, ', '.join(valid_values)))
 
-    def __getitem__(self, key: Union[int, str]) -> "Variable":
+    def __getitem__(self, key: Union[int, str]) -> "PdoVariable":
         var = None
         if isinstance(key, int):
             # there is a maximum available of 8 slots per PDO map
@@ -229,7 +229,7 @@ class Map:
                 var = self.__getitem_by_name(key)
         return var
 
-    def __iter__(self) -> Iterable["Variable"]:
+    def __iter__(self) -> Iterable["PdoVariable"]:
         return iter(self.map)
 
     def __len__(self) -> int:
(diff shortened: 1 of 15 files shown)


christiansandberg (approved):
Seems like a good improvement. As long as we keep backwards compatibility I see no problem.

sveinse:
Updated with lastest commits in master.

canopen/pdo/base.py
```diff
@@ -528,11 +528,11 @@ def wait_for_reception(self, timeout: float = 10) -> float:
         return self.timestamp if self.is_received else None
 
 
-class Variable(variable.Variable):
+class PdoVariable(variable.Variable):
```
acolomb, line 531: This is pure bikeshedding, but why do you use CamelCase with the lowercase `Pdo` abbreviation here, while for the OD you use the all-caps `OD` spelling as in `ODVariable`?  Shoult consistenly use uppercase for abbreviations here.

canopen/sdo/base.py
```diff
@@ -80,14 +80,14 @@ def download(
         raise NotImplementedError()
 
 
-class Record(Mapping):
+class SdoRecord(Mapping):
```
acolomb, line 83: Same here, `SDORecord`?  Please just be consistent.

acolomb (changes requested):
Looks good in principle, just a nit-pick about inconsistent uppercase variants.

canopen/pdo/base.py
sveinse: @acolomb you're right, this is inconsistent. I didn't make a very deliberate choice, rather trying to adopt the gist from the rest of canopen. After looking more closer at it, I see canopen is equally inconsistent, e.g. `class PDO` vs `class PdoBase`. Looking at all class names, the majority is CamelCase, with the notable exception of `PDO`, `TPDO`, `RPDO` and not used with an additional suffix.

To be consistent with naming in canopen, `ODVariable` should then be `OdVariable` perhaps?

canopen/sdo/base.py
sveinse: With reference to comment above, `SdoRecord` is correct if `SdoBase` is correct I would argue.

canopen/pdo/base.py
acolomb: Yes, I guess that's an acceptable solution.  Just don't introduce even more inconsistency ;-)

christiansandberg: Naming is probably inconsistent everywhere else anyway, so it's not super important. Not sure what's best for readability, but I'm nowadays leaning more towards keeping abbreviations upper-case even if it makes it harder to separate the words.

Issue #273, opened 2021-10-21
Read PDO configuration from EDS OD


sveinse:
In my setup I read the PDO configuration from the remote node. To save bus traffic, how can I read the same PDO configuration from the EDS file? This remote note has static PDO configuration, so there no need to run SDOs to the remote to get it. I believe the EDS file contain the information.

```python
    # Create the node object and load the OD
    node: canopen.RemoteNode = network.add_node(nodeid, 'PulseOD.eds')

    # Read the PDO configuration from the target
    await node.tpdo.aread()
    await node.rpdo.aread()
```

PR #370, opened 2023-04-08
Implement reading PDO from OD defaults (#273)

sveinse:
A small improvement based on discussion in #273 that allows reading the PDO configuration from the object dictionary _defaults_ rather than requesting them from the remote node.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -50,10 +50,10 @@ class PdoBase(Mapping):
     def __len__(self):
         return len(self.map)
 
-    def read(self):
+    def read(self, from_od=False):
         """Read PDO configuration from node using SDO."""
         for pdo_map in self.map.values():
-            pdo_map.read()
+            pdo_map.read(from_od=from_od)
 
     def save(self):
         """Save PDO configuration to node using SDO."""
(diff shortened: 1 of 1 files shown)


samsamfire:
Hi, I did something similar on my side for two main reasons (just putting this for explanation):

- When using a large OD, reading via SDO can be quite long : ~5 seconds. When you are testing and just running scripts, this can be really annoying.
- Sometimes you would like to use PDOs but don't want to use SDO as this can break communication if there is already a master on the bus talking to the node.

Anyway, cool improvement !

canopen/pdo/base.py
```diff
@@ -313,43 +313,49 @@ def add_callback(self, callback: Callable[["Map"], None]) -> None:
         """
         self.callbacks.append(callback)
 
-    def read(self) -> None:
+    def read(self, from_od=False) -> None:
         """Read PDO configuration for this map using SDO."""
-        cob_id = self.com_record[1].raw
+
+        def _raw_from(param):
```
acolomb, line 319: Do we need an underscore name here?  The function is only defined within this scope, so there is nothing to be gained by trying to hide it?

acolomb (commented):
Nice addition.  Only one nit-pick comment.

acolomb:
And I think this needs documentation?

canopen/pdo/base.py
sveinse: No, not really. Only a convention. My habit is to name functions that are defined within a fn or meth using a underscore prefix, but it is not necessity.

PR #378, opened 2023-05-24
Feature/eds parse factor and description

marcovannoord:
For my work, i needed to use the Factor and Description, that already seem to have been implemented CANopen. However, they were currently not/never set by an EDS-file.
I added parsing for the Factor and Description, and i have added some unit-tests for them.


unsanded:
I agree that this would be a nice addition, because it feels kind of weird to always use `var.raw` instead of `var.phys`. And it would also prevent errors.

marcovannoord:
@christiansandberg What is your opinion about this MR?

christiansandberg:
Sorry for the delay. I think it makes sense to be able to specify it. Not sure what the EDS spec says about custom attributes but it should be OK.

PR #380, opened 2023-05-26
Enable inline comments when reading files

XorJoep:
Some dcf/eds files contain inline comments, seperated by ';'. This fix allows those files to be read.

For example: 
DataType=0x0007  ; Unsigned32


Jeroenve:
When will a new release with this change be pushed to https://pypi.org/project/canopen/?

christiansandberg:
I'll create a new release in the coming days.

Issue #385, opened 2023-07-07
PDOMapping in EDS file does not check for ValueError Exception


Obsnold:
I've just encountered some EDS files that have the PDOMapping value with no value after the = sign e.g.

```
[3020sub0]
ParameterName=Number of entries
ObjectType=0x07
DataType=0x0005
AccessType=ro
DefaultValue=0x9
PDOMapping=
```
when I try to load the file it throws the error:
```
ValueError: invalid literal for int() with base 0: ''
```

At first I though the EDS file had been corrupted or generated incorrectly( it states it was made with CANopen Design Tool 2.2.34).  

Also it only seems to happen on SubIndex 0.

I double checked the spec CiA 306 version 1.3.0 and noticed this about arrays:

> the entry PDOMapping for all sub-indexes except 0 and 255 is given by the entry PDOMapping of the main object. sub-index 0 is assumed not to be mappable

so that's why I'm guessing the tool though it was ok to leave it blank. 

I saw that the library handled the same situation for other values (DefaultValue seem to have the same issue a lot) , so I'' create a pull request to handle it in the same way for PDOMapping.

PR #386, opened 2023-07-07
Check for ValueError exception when reading PDOMapping entrys from ED…

Obsnold:
…S files.
Resolves the following issue:
- https://github.com/christiansandberg/canopen/issues/385

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -274,8 +274,10 @@ def build_variable(eds, section, node_id, index, subindex=0):
             logger.warning("%s has an unknown or unsupported data type (%X)", name, var.data_type)
             # Assume DOMAIN to force application to interpret the byte data
             var.data_type = datatypes.DOMAIN
-
-    var.pdo_mappable = bool(int(eds.get(section, "PDOMapping", fallback="0"), 0))
+    try:
+        var.pdo_mappable = bool(int(eds.get(section, "PDOMapping", fallback="0"), 0))
+    except ValueError:
+        pass
 
     if eds.has_option(section, "LowLimit"):
         try:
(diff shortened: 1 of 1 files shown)


acolomb (commented):
I think this possibly hides real errors? Like if someone writes `PDOMapping=yes`

Wouldn't it be better to check explicitly for an empty string and convert it to zero?

Obsnold:
Sure that sounds good to me, I'll try and update ti today.

Obsnold:
Alright I've updated it, please take another look.
Thanks!

canopen/objectdictionary/eds.py
```diff
@@ -274,8 +274,10 @@ def build_variable(eds, section, node_id, index, subindex=0):
             logger.warning("%s has an unknown or unsupported data type (%X)", name, var.data_type)
             # Assume DOMAIN to force application to interpret the byte data
             var.data_type = datatypes.DOMAIN
-
-    var.pdo_mappable = bool(int(eds.get(section, "PDOMapping", fallback="0"), 0))
+    
+    pdo_mappable_string = eds.get(section, "PDOMapping", fallback="0")
+    if pdo_mappable_string != "":
+        var.pdo_mappable = bool(int(pdo_mappable_string, 0))
```
christiansandberg, line 280: I guess it is still possible to have invalid input.

Maybe something like this?

```python
var.pdo_mappable = eds.get(section, "PDOMapping", fallback="0") not in ("", "0")
```

Obsnold: I think we want the invalid input to throw an exception so real errors are caught, though I guess it's a question whether `PDOMapping=` is valid input or not? 

I no longer have access to the EDS files that were causing the issue so I don't have anything to test against.
If I have time I'll try and poke around to see if I can find any other files that exhibit this issue.

it does make me wonder if the handling of invalid input on the other entries should also throw an error?

acolomb: Re-reading the spec (CiA 306), I think we should not accept `PdoMapping=` as a "false" option value.  It clearly states that the option must have either the value 0 or 1.  We are already much more lenient by accepting any nonzero integer as true.  But no value at all is not valid.

In general, I agree that our EDS error handling is suboptimal, as parsing errors are usually only logged as warnings, but ignored.  There is no way of reacting programmatically to the errors.  At the least, we should have a `strict` mode parameter which raises exceptions.

PR #402, opened 2023-10-24
Removed automatic change from NMT BOOTING to NMT PRE-OPERATIONAL

### LINKED ISSUE #401: Can't wait for RemoteNode to reach PRE-OPERATIONAL
## Problem
The NMT heartbeat implementation does not allow for receiving the bootup status from the node. It would be nice if the bootup status would just be received and stored in the state field. Without this state there is no way of checking if the node is still booting.

### Goal (Functional)
My goal is to, via NMT, reset the node and wait for it to reach Pre-Operational again. My node is running hardware that requires some time to setup, this time can differ based on the amount of devices connected to the node what makes it difficult to define a time. While setting up the node is in a BOOTING / INITIALISATION NMT state. After all setups are completed, the node enters PRE-OPERATIONAL what makes it ready to be started and used. I can't define a specific time to wait between resetting and starting again so I would like to wait for the pre-operational state. Only I can not do this due to the booting state being updated to pre-operational.

### Solution
Remove this if statement in the callback. I don't know what other implications this might have.
https://github.com/christiansandberg/canopen/blob/6f9f06eb4a2d8989583e35ec0d4b34d7705726d0/canopen/nmt.py#L125

Jef-GB:
# Description
Removed the if statement for automatically setting the NMT state to pre-operational when the node published a message with the value booting.

## Changes
- Remove auto change to Pre-Op from Booting

### Linked Issue
Fixes #401

### DIFF
--- a/canopen/nmt.py
+++ b/canopen/nmt.py
@@ -122,11 +122,8 @@ class NmtMaster(NmtBase):
             logger.debug("Received heartbeat can-id %d, state is %d", can_id, new_state)
             for callback in self._callbacks:
                 callback(new_state)
-            if new_state == 0:
-                # Boot-up, will go to PRE-OPERATIONAL automatically
-                self._state = 127
-            else:
-                self._state = new_state
+            
+            self._state = new_state
             self._state_received = new_state
             self.state_update.notify_all()
 


acolomb:
Why do you think this change is correct?

The CANopen standard clearly defines that a node enters the PRE-OPERATIONAL state automatically, right after sending its boot-up message (section 7.3.2.2.1).  So I think the previous behavior is correct.

acolomb:
Ping @Jef-GB, I'm about to close this as I don't see any wrong behavior to fix here.  Unless you have a good explanation why it's necessary.

erlend-aasland:
> The CANopen standard clearly defines that a node enters the PRE-OPERATIONAL state automatically, right after sending its boot-up message (section 7.3.2.2.1). So I think the previous behavior is correct.

What about non-compliant and buggy nodes? Would it not make sense to just await the PRE-OPERATIONAL heartbeat instead?

acolomb:
It's perfectly normal for a device to not send any more NMT messages after its boot-up. Heartbeat needs to be enabled explicitly. Assuming standard-compliant behavior by default is the Right Thing to do, any known buggy devices could be handled by subscribing via `add_heartbeat_callback()` and tracking state manually.

erlend-aasland:
> It's perfectly normal for a device to not send any more NMT messages after its boot-up.

In that case, there's no need for this change. The docs could be made a little bit more explicit about this, though.

PR #405, opened 2023-11-16
Add `from_od` flag to the `RemoteNode.load_configuration`

vongostev:
In the library we have some methods to load settings from OD, for example `node.rpdo.read(from_od=True)`, but we can not do this for entire configuration.
This PR fixes it.

### DIFF
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -145,7 +145,7 @@ class RemoteNode(BaseNode):
                 logger.warning('[ERROR SETTING object {0:#06x}:{1:#06x}]  {2}'.format(index, subindex, str(e)))
                 raise
 
-    def load_configuration(self):
+    def load_configuration(self, from_od: bool = False):
         ''' Load the configuration of the node from the object dictionary.'''
         for obj in self.object_dictionary.values():
             if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
@@ -154,4 +154,4 @@ class RemoteNode(BaseNode):
                         self.__load_configuration_helper(subobj.index, subobj.subindex, subobj.name, subobj.value)
             elif isinstance(obj, ODVariable) and obj.writable and (obj.value is not None):
                 self.__load_configuration_helper(obj.index, None, obj.name, obj.value)
-        self.pdo.read()  # reads the new configuration from the driver
+        self.pdo.read(from_od=from_od)  # reads the new configuration from the driver
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -33,7 +33,7 @@ Callback = Callable[[int, bytearray, float], None]
 class Network(MutableMapping):
     """Representation of one CAN bus containing one or more nodes."""
 
-    def __init__(self, bus=None):
+    def __init__(self, bus: can.BusABC | None = None):
         """
         :param can.BusABC bus:
             A python-can bus instance to re-use.
@@ -110,7 +110,8 @@ class Network(MutableMapping):
                 if node.object_dictionary.bitrate:
                     kwargs["bitrate"] = node.object_dictionary.bitrate
                     break
-        self.bus = can.interface.Bus(*args, **kwargs)
+        if self.bus is None:
+            self.bus = can.Bus(*args, **kwargs)
         logger.info("Connected to '%s'", self.bus.channel_info)
         self.notifier = can.Notifier(self.bus, self.listeners, 1)
         return self


acolomb:
What exactly is the use case for this?  `load_configuration()` actually does not upload the objects *from* the remote node, but it downloads the OD variable values *to* the remote node.  Then it goes on to read back the applied PDO configuration.

How does it make sense to buy in to the vast exchange of SDO downloads, but then saving a couple of SDO uploads for reading back the actually accepted configuration?

canopen/node/remote.py
```diff
@@ -154,4 +154,4 @@ def load_configuration(self):
                         self.__load_configuration_helper(subobj.index, subobj.subindex, subobj.name, subobj.value)
             elif isinstance(obj, ODVariable) and obj.writable and (obj.value is not None):
                 self.__load_configuration_helper(obj.index, None, obj.name, obj.value)
-        self.pdo.read()  # reads the new configuration from the driver
+        self.pdo.read(from_od=from_od)  # reads the new configuration from the driver
```
sveinse, line 157: This seems to be the reason for adding the `from_od` argument. The current imeplementations calls `self.pdo.read()` with default parameters will read from the remote.

Maybe it's better to do `self.pdo.read(True)` and not add the `from_od` argument? The function description does not allude to it doing something against the remote.

sveinse:
Perhaps its better to just call `self.pdo.read(True)` and skip the `from_od` argument to the function?

acolomb:
> Perhaps its better to just call `self.pdo.read(True)` and skip the `from_od` argument to the function?

The `.load_configuration()` method is more equivalent to `pdo.write()`, though.  It goes through all objects in the OD actually and downloads the values to the remote node via SDO.  Then to avoid having mismatching information about the PDO-related objects, it reads back (uploads) the PDO records again via SDO.

In essence, there is nothing to fix here IMHO, just don't call `node.load_configuration()` if you want to work with the values from the OD.  They are automatically read in when initializing the node's OD from a DCF file (or defaults from the EDS).  The only missing step would be to update the PDO configuration as well, which `node.pdo.read(from_od=True)` does.

sveinse:
The PDO state is independent of the OD state and when the `RemoteNode()` class instance is created. The PDO objects are created and referenced to the OD, but their mapping state vars, like `.cob_id`, `.rtr_allowed` etc. are only populated with a call to `PdoMap.read()`. So the PDO objects are not useable out of box. `RemoteNode(..., load_od=True)` can be used but it ends up doing remote SDO-requests since `PdoMap.read()` has default `from_od=False`.

Maybe `PdoMaps().__init__()` when the `PdoMap()` instance is created from the OD should also call `PdoMap.read(from_od=True)` to sync the PDO state with the (default) OD state? This way the PDOs would ready to use after creating the `RemoteNode` instance. If the user needs to update _from_ the remote PDO configuration, the `PdoMap.read(from_od=False)` will do the job as per usual.

erlend-aasland:
André:
> The `.load_configuration()` method is more equivalent to `pdo.write()`, though. It goes through all objects in the OD actually and downloads the values to the remote node via SDO. Then to avoid having mismatching information about the PDO-related objects, it reads back (uploads) the PDO records again via SDO.

It would be nice to add hint about this in the `.load_configuration` docs.

erlend-aasland:
> Maybe `PdoMaps().__init__()` when the `PdoMap()` instance is created from the OD should also call `PdoMap.read(from_od=True)` to sync the PDO state with the (default) OD state? This way the PDOs would ready to use after creating the `RemoteNode` instance. If the user needs to update _from_ the remote PDO configuration, the `PdoMap.read(from_od=False)` will do the job as per usual.

First: I think this discussion belongs on the tracker, not on a PR.
Second: I prefer APIs without side-effects, especially when instantiating objects. IOW, when I create a `RemoteNode` instance, I would not want it to default to doing I/O-operations; I prefer to control those explicitly, since it makes for code that is easier to reason about.

sveinse:
> Second: I prefer APIs without side-effects, especially when instantiating objects. IOW, when I create a `RemoteNode` instance, I would not want it to default to doing I/O-operations; I prefer to control those explicitly, since it makes for code that is easier to reason about.

It doesn't do IO unless `RemoteNode(..., load_od=True)` is used.

erlend-aasland:
FTR, I'd like to take a stab at improving the docs; it is a good opportunity for me to get to know the various APIs of this excellent library better.

Cookie-licking disclaimer: if I don't post a PR within a day or two, feel free to beat me to it :)

PR #426, opened 2024-05-09
Add support for using dot (.) to get subobjects from the OD

sveinse:
This PR introduces the ability to use dot to access sub objects from the OD.

```python
od: ObjectDictionary
v = od["Parameter"]["Subparam"].raw   # The current way
v = od["Parameter.Subparam"].raw      # The newly added way
```

The rationale is that when working with PDOs this syntax is used and it would make sense to have the same mechanism for SDO.

```python
rpdo = remotenote.rpdo[1]
rpdo["Parameter.Subparam"].phys = 80

sdo = remotenode.sdo
sdo["Parameter.SubParam"].phys = 80
```

The current ways of accessing using multiple `__getitem__` works fine. The only drawback is if an object has period (.) in its name.

Unittests have been updated as a part of this PR.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -109,8 +109,8 @@ class ObjectDictionary(MutableMapping):
         item = self.names.get(index) or self.indices.get(index)
         if item is None:
             if isinstance(index, str) and '.' in index:
-                parts = index.split('.')
-                return self[parts[0]][".".join(parts[1:])]
+                idx, sub = index.split('.', maxsplit=1)
+                return self[idx][sub]
             name = "0x%X" % index if isinstance(index, int) else index
             raise KeyError("%s was not found in Object Dictionary" % name)
         return item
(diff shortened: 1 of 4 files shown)


christiansandberg (approved):
Seems like a good addition. Maybe it should also be added to the docs?

sveinse:
@christiansandberg I added a few examples for the documentation

canopen/objectdictionary/__init__.py
```diff
@@ -108,6 +108,9 @@ def __getitem__(
         """Get object from object dictionary by name or index."""
         item = self.names.get(index) or self.indices.get(index)
         if item is None:
+            if isinstance(index, str) and '.' in index:
+                parts = index.split('.')
+                return self[parts[0]][".".join(parts[1:])]
```
erlend-aasland, line 113: More compact and also faster variant:

```suggestion
                idx, sub = index.split('.', maxsplit=1)
                return self[idx][sub]
```

erlend-aasland: Using positional-only arguments (`split('.', 1)`), is even faster, but I would prefer the keyword argument for readability.

sveinse: Thank you. I will update the PR

erlend-aasland (approved):

sveinse:
As this is approved by @christiansandberg, can someone with write access, please commit this? @acolomb perhaps? Thanks!

sveinse:
Thank you @christiansandberg

Issue #274, opened 2021-10-28
OD values from .dcf not loaded


martinlucan:
Since I need a version compatible with Python 2, I use a fork of the version which is automatically installed via “pip install canopen” (ver. 1.2.1), with Cyclic Synchronous modes additionally implemented in p402.py file. Unfortunately, I have noticed that the configuration saved in the standard .dcf file is NOT uploaded at motor startup. Therefore I have to always use manufacturers' software to change some parameters. Both, .dcf and .eds files are supported in documentation. Do you know what might cause this problem of not uploading a configuration? Is .eds format preferred over .dcf?

martinlucan:
Thank you @acolomb for the answer. We would prefer to upload parameters from .dcf to the device at startup. Could you please provide me with some example code on how to do so?

acolomb:
Can you upload the DCF file here? I don't know how it will work exactly, but I'm sure a solution is possible when reusing some code parts from this library.

martinlucan:
Sure, I guess, it doesn't matter on a .dcf file, but here is a sample file:

[dcf.zip](https://github.com/christiansandberg/canopen/files/7496949/dcf.zip)

martinlucan:
I simply made use of` import_eds` function from `objectdictionary.eds`, although parameters from .dcf file are not loaded anyways.   @acolomb do you know what might be the issue? Do I have to use any other function? Thank you.

acolomb:
As I said, I don't think the code is designed to read the actual object *values* from DCF files.  Just the object dictionary *structure*.  You may be able to extend that.  Sorry I don't have time right now to look into it deeper.

acolomb:
Fixed by #427.  Sorry for not really understanding the purpose of the DCF file usage here when I first looked at the issue.

PR #427, opened 2024-05-13
Add support for loading DCF configuration to remote node

meifonglow:
As part of instrument commissioning process, I want to apply the configuration values stored in a DCF to a remote node, including PDO mapping and communications parameters.  If a value was not specified, then the EDS default value is used instead.

This commit provides this capability.

Potentially addresses #274.


canopen/node/remote.py
```diff
@@ -147,8 +147,19 @@ def __load_configuration_helper(self, index, subindex, name, value):
...
     def load_configuration(self):
         ''' Load the configuration of the node from the object dictionary.'''
+        # First apply PDO configuration from object dictionary
+        self.pdo.read(from_od=True)
+        self.pdo.save()
+
+        # Now apply all other records in object dictionary
         for obj in self.object_dictionary.values():
-            if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
+            if "PDO" in obj.name:
```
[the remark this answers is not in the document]
meifonglow: Do you mean explicitly ignoring objects with indices within supported PDO ranges, i.e: `0x1400 <= obj.index < 0x1C00` ?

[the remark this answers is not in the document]
christiansandberg: Yes exactly. I'm thinking using names might yield both false positives and false negatives.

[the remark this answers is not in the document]
meifonglow: Thanks that makes sense.
I've also removed excluding objects with "COB-ID" keyword.  They are being applied OK by `__load_configuration_helper()` and would have been incorrectly ignored in previous commit.

canopen/node/remote.py
```diff
@@ -147,8 +147,16 @@ def __load_configuration_helper(self, index, subindex, name, value):
...
+        # First apply PDO configuration from object dictionary
+        self.pdo.read(from_od=True)
+        self.pdo.save()
+
+        # Now apply all other records in object dictionary
         for obj in self.object_dictionary.values():
-            if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
+            if 0x1400 <= obj.index < 0x1c00:
+                # Ignore PDO related objects
+                logger.debug(f"{obj.index:#06x}: {obj.name} already applied by pdo.save()")
```
acolomb, line 158:
```suggestion
                logger.debug("0x%04X: %s already applied by pdo.save()", obj.index, obj.name)
```

See #443.  Do we really need this much debug logging though?

canopen/pdo/base.py
```diff
@@ -314,11 +314,20 @@ def add_callback(self, callback: Callable[["Map"], None]) -> None:
         self.callbacks.append(callback)
 
     def read(self, from_od=False) -> None:
-        """Read PDO configuration for this map using SDO."""
+        """Read PDO configuration for this map
```
acolomb, line 317:
```suggestion
        """Read PDO configuration for this map.
```

acolomb (changes requested):
Looks good overall, just a couple of style nit-picks.

canopen/node/remote.py
christiansandberg: Agree, it is something that would be logged every time so it may not add much value.

christiansandberg (approved):

canopen/node/remote.py
meifonglow: Fair enough.  Removing.

meifonglow, line 158: Removing as per consensus.
```suggestion
```

canopen/node/remote.py
```diff
@@ -147,8 +147,15 @@ def __load_configuration_helper(self, index, subindex, name, value):
...
+        # First apply PDO configuration from object dictionary
+        self.pdo.read(from_od=True)
+        self.pdo.save()
+
+        # Now apply all other records in object dictionary
         for obj in self.object_dictionary.values():
-            if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
+            if 0x1400 <= obj.index < 0x1c00:
+                # Ignore PDO related objects
+            elif isinstance(obj, ODRecord) or isinstance(obj, ODArray):
```
meifonglow, line 158: Fixing up incorrect suggestion.

```suggestion
            if 0x1400 <= obj.index < 0x1c00:
                # Ignore PDO related objects
                pass
            elif isinstance(obj, ODRecord) or isinstance(obj, ODArray):
```

canopen/node/remote.py
```diff
@@ -147,8 +147,16 @@ def __load_configuration_helper(self, index, subindex, name, value):
...
+        self.pdo.read(from_od=True)
+        self.pdo.save()
+
+        # Now apply all other records in object dictionary
         for obj in self.object_dictionary.values():
-            if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
+            if 0x1400 <= obj.index < 0x1c00:
+                # Ignore PDO related objects
+                pass
+            elif isinstance(obj, ODRecord) or isinstance(obj, ODArray):
```
acolomb, line 159:
```suggestion
                continue
            if isinstance(obj, ODRecord) or isinstance(obj, ODArray):
```

No need for else if you skip the iteration right away.

acolomb (approved):

PR #428, opened 2024-05-14
build: Modernize, use setuptools 69

acolomb:
Migrate `setup.cfg` to `pyproject.toml` for newer setuptools compat and create py.typed marker for automatic type checker discovery.

The `py.typed` marker should solve the "Cannot find implementation or library stub" errors from mypy if installed properly.  Since the code already has internal type hints, adding the marker is sufficient for mypy to use them.

The `setuptools>=69` dependency includes a fix to automatically include the `py.typed` marker in the distribution, thus making type hints usage much easier.  The newer version however fails to build when the basic project metadata is missing in `pyproject.toml`.  Thus the settings from `setup.cfg` are migrated entirely, upgrading to `setuptools_scm >= 8` in
the process.

This might need a requirement bump to Python >= 3.8 as well, since that's what the new setuptools version requires.


sveinse (approved):
Technically I'm not a collaborator for this project, but this change looks good.

acolomb:
@christiansandberg I would like your opinion on this, do you see any risk for larger breakage?

pyproject.toml
```diff
@@ -1,6 +1,40 @@
 [build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
+requires = ["setuptools>=69", "wheel", "setuptools_scm>=8"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "canopen"
+authors = [
+    {name = "Christian Sandberg", email = "christiansandberg@me.com"},
```
christiansandberg, line 8: Feel free to add more people here. 🙂

pyproject.toml
```diff
@@ -1,6 +1,40 @@
...
 [tool.setuptools_scm]
-write_to = "canopen/_version.py"
+version_file = "canopen/_version.py"
+
+[tool.pytest.ini_options]
+testpaths = [
+    "test",
+]
+filterwarnings = [
+    "ignore::DeprecationWarning",
```
christiansandberg, line 39: Could this be useful to see anyway?

christiansandberg, line 39: Could this be useful to see anyway?

christiansandberg (approved):
Sorry for the absence. Go ahead and merge!

pyproject.toml
acolomb: Maybe so.  I suggest we leave it be until after the release.  The only warnings I get currently are for `datetime.utcnow()` and the `bustype` argument to `can.interface.Bus`.

pyproject.toml
acolomb: Alright, I added myself and @af-silva, completing the top-three list of contributors according to GitHub.

Email address taken from one of your commits, @af-silva, hope that's alright?

af-silva: @acolomb It is ok :)

sveinse: I counted up the PRs that go into this release, and a large fraction of them are mine. Please feel free to use my name and email (sveinse@seldal.com) in any release notes.

PR #431, opened 2024-05-14
Rename Maps to PdoMaps and Map to PdoMap

sveinse:
This PR renames the class `canopen.pdo.base.Map` to `canopen.pdo.base.PdoMap` and correspondingly `.Map` to `PdoMap` in line with the similar change as PR #368. The names `PdoMap()` and `PdoMaps()` with respect to upper case "PDO" is chosen to be consistent with how `Sdo*` have been named.

Compatibility entries have been added, so this won't break old code.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -124,7 +124,7 @@ class PdoBase(Mapping):
             pdo_map.stop()
 
 
-class PdoMaps(Mapping[int, "PdoMap"]):
+class PdoMaps(Mapping):
     """A collection of transmit or receive maps."""
 
     def __init__(self, com_offset, map_offset, pdo_node: PdoBase, cob_base=None):
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -27,7 +27,7 @@ class PdoBase(Mapping):
 
     def __init__(self, node):
         self.network = None
-        self.map = None  # instance of Maps
+        self.map = None  # instance of PdoMaps
         self.node = node
 
     def __iter__(self):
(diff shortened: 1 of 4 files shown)


erlend-aasland:
Why not keep the old names as aliases coupled with DeprecationWarnings if used? That way you'll give users a chance to update their code before actually performing the breaking change.

sveinse:
Mainly because there is no DeprecatedWarnings on the other change (#368) so I figured that wasn't wanted behavior. Also note that this PR doesn't break anything.

We can add DeprecatedWarnings if that is what we want to have.

erlend-aasland:
Ah, I missed that you already provided aliases for the old variant; nice. I would consider adding `.. versionchanged` directives in the docs to inform users of the change, though.

> Mainly because there is no DeprecatedWarnings on the other change [...]

👍

sveinse:
> I would consider adding `.. versionchanged` directives in the docs to inform users of the change, though.

While I agree, I see there is no such framework established in the existing docs. Or any change log for that matter. I'm not sure this simple PR should establish that precedence thou.

erlend-aasland:
Fair enough!

erlend-aasland (approved):

canopen/pdo/base.py
```diff
@@ -124,7 +124,7 @@ def stop(self):
             pdo_map.stop()
 
 
-class Maps(Mapping):
+class PdoMaps(Mapping[int, "PdoMap"]):
```
sveinse, line 127: This syntax is not compatible with py3.8, while it works with >py3.9. With respect to PR #428, are we aiming for minimum 3.8?

sveinse: I will revert and test it with py3.8 to verify there is no other surprises

acolomb (approved):

acolomb:
@sveinse Could you please check if there are any references to the old names in the docs / examples?

sveinse:
> @sveinse Could you please check if there are any references to the old names in the docs / examples?

Yes, I did not find any more that the ones that are included in the PR. I just double checked. 👌

PR #433, opened 2024-05-15
Clean up imports of old modules

sveinse:
The canopen package have import leftovers from earlier versions of Python. This PR cleans up the no longer valid imports.

* `import Queue` dates to py 2
* `import collections.*Mapping` was changed to `collections.abc.*Mapping` in py 3.3
* `import ConfigParser` dates to py 2
* `import xml.etree.cElementTree` is deprecated since py 3.3. It will pick the fastest option when using `xml.etree.ElementTree`
* Fixed `.base` into `canopen.pdo.base` in canopen/pdo/__init__.py
* Fixed `.base` into `canopen.sdo.base` in canopen/sdo/__init__.py

### DIFF
--- a/canopen/pdo/__init__.py
+++ b/canopen/pdo/__init__.py
@@ -4,7 +4,7 @@ from canopen import node
 from canopen.pdo.base import PdoBase, Maps
 
 # Compatibility
-from .base import Variable
+from canopen.pdo.base import Variable
 
 logger = logging.getLogger(__name__)
 
--- a/canopen/sdo/__init__.py
+++ b/canopen/sdo/__init__.py
@@ -4,4 +4,4 @@ from canopen.sdo.server import SdoServer
 from canopen.sdo.exceptions import SdoAbortedError, SdoCommunicationError
 
 # Compatibility
-from .base import Variable, Record, Array
+from canopen.sdo.base import Variable, Record, Array
--- a/canopen/lss.py
+++ b/canopen/lss.py
@@ -1,10 +1,7 @@
 import logging
 import time
 import struct
-try:
-    import queue
-except ImportError:
-    import Queue as queue
+import queue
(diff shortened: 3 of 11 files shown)


canopen/pdo/__init__.py
```diff
@@ -4,7 +4,7 @@
 from canopen.pdo.base import PdoBase, Maps
 
 # Compatibility
-from .base import Variable
+from canopen.pdo.base import Variable
```
acolomb, line 7: Why change to absolute package reference here?  I'd actually rather go the other way and use relative imports within the same package.  Maybe it's a matter of taste, but I liked the old style.

canopen/sdo/__init__.py
```diff
@@ -4,4 +4,4 @@
 from canopen.sdo.exceptions import SdoAbortedError, SdoCommunicationError
 
 # Compatibility
-from .base import Variable, Record, Array
+from canopen.sdo.base import Variable, Record, Array
```
acolomb, line 7: Same here.

acolomb (commented):
Looks good, but I'd like some discussion / reasoning about using absolute imports within the same package.  DRY tells us to leave out the package name in places within the same package, and I've seen very few use cases where relative imports don't work.

canopen/pdo/__init__.py
sveinse: No other reason that PEP 8 recommends using absolute imports, so it's become a habit for many projects. PEP 8 also states "explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose". So either is acceptable, but with preference to the former. And lastly, the _rest_ of canopen is using absolute imports. I find it messy that the import styles are mixed.

Issue #439, opened 2024-05-19
Provided value of bytes pass through encode_raw()


sveinse:
`objectdictionary.__init__.ODVariable.encode_raw()` converts native datatypes to bytes.

https://github.com/christiansandberg/canopen/blob/54ebbd45a8084f664c21be270fae0f7a1a658355/canopen/objectdictionary/__init__.py#L403-L404

What is the use-case for this logic? It means that when a value of bytes is passed, it will bypass the encoding and use the user provided bytes value rather than raising `TypeError`. Do you remember the intention @christiansandberg?

I'm currently working on adding the missing datatypes, and I encountered this code and is curious of its use case. The unittests pass if this is commented out, so we don't have a test case for it either.

acolomb:
I think it's a useful option, if the byte-order handling is already taken care of externally.  For example I have a use case where values can either be programmed to an EEPROM directly, or sent via CANopen.  If the EEPROM representation is already known, it makes no sense to parse those bytes into an integer and then re-encode that to the CANopen little-endian format.

Another use case would be strings with a known encoding, bypassing the encoding for the two expected string object data types.  Or even completely free-form OCTET_STRING objects, which need to be passed in verbatim.  I actually do have such a thing in use, for a two-dimensional array of values which is stored in a single CANopen object entry.

So I don't see how this could be not useful.  It should definitely be kept.

christiansandberg:
OCTET_STRING and DOMAIN support but maybe it could be more strict to only apply if data type is actually that.

sveinse:
On a second review of the code, I see that `encode_raw()` have the override ability by inputting the data as `bytes` or `bytearray` as discussed. However, `decode_raw()` does not have this ability.

@acolomb have a good use case: If the assumptions on encoding e.g. for UNICODE_STRING is wrong. I.e. an application wants to use something else than "utf-16-le", then this becomes a way to override the library behavior. -- But then we need it in both directions.

acolomb:
Well it's useful as is. If one wants the undecoded bytes from an upload, there is always a possibility to use one of the lower level methods, where no decoding is done at all. I don't think we need perfect symmetry here just for the sake of it.

sveinse:
Thanks. I got the answers I was looking for.

Issue #436, opened 2024-05-18
Inconsistent data truncation of unsupported data types in SDO upload


sveinse:
I'm having problems reading SDO objects that are using UNSIGNED48. It used to work before PR #395 by @ljungholm - but it turns out to be coincidental.

https://github.com/christiansandberg/canopen/blob/24df6e846a4e71508fd82c378d1385d65f000c94/canopen/sdo/client.py#L123-L133

I believe the root cause is that canopen doesn't support the UNSIGNED48 type (which I'll contribute in another PR). The unfortunate behavior is that due to the unknown type, the data gets truncated to 1 byte. It didn't do that before #395 .

```
>> upload((8192, 2), {})
<-- 602     :  40 00 20 02 00 00 00 00
--> 582     :  41 00 20 02 08 00 00 00
  << read_response() = b'A\x00 \x02\x08\x00\x00\x00'
<-- 602     :  60 00 00 00 00 00 00 00
--> 582     :  00 b2 01 20 02 91 12 00
  << read_response() = b'\x00\xb2\x01 \x02\x91\x12\x00'
<-- 602     :  70 00 00 00 00 00 00 00
--> 582     :  1d 00 00 00 00 00 00 00
  << read_response() = b'\x1d\x00\x00\x00\x00\x00\x00\x00'
     data=b'\xb2\x01 \x02\x91\x12\x00\x00'
     response_size = 8
     var.datatype = 25   # UNSIGNED48 (unsupported)
     var_size = 1   # Because len(var) on unknown datatypes is 8
     data = b'\xb2'   # Truncated due to ^
<< upload() = b'\xb2'
```

Observations:
* Reading SDO objects _without_ OD record will preserve data as-is
* Reading SDO objects _with_ OD, but of unsupported type will return data of size 1
* `ODVariable.__len__()` return 8 on unknown types, while it reports number of bits in the other types. This is why it gets truncated to one byte.

Would you agree that `upload()` should return the the full data for both of the first use cases? Is the fix that `ODVariable.__len__()` on unknown datatypes should return 64?

sveinse:
> Would you agree that `upload()` should return the the full data for both of the first use cases? Is the fix that `ODVariable.__len__()` on unknown datatypes should return 64?

[...]

PR #406, opened 2023-11-28
24 bit support

raffi-g:
The unsigned24 part was already added in https://github.com/christiansandberg/canopen/pull/267. I took the idea from there, but made it a little simpler. Also I added the signed version and extended the unit tests.

PR #440, opened 2024-05-19
Implement the remaining canopen datatypes

### LINKED ISSUE #320: Request Data Type Additions
Requesting support for more Object Dictionary Data Type Support.
Specific interest in "UNSIGNED24".
Data Types are described in CAN-in-Automation Design Specification 301, section 7.4.7.1 (Table 44) (page 85, see attached image)

I believe this would require add to this list found in https://github.com/christiansandberg/canopen/blob/c7ad0a822a20a765f0fc7c43be0f4450b7a536ef/canopen/objectdictionary/__init__.py

`class Variable(object):
    """Simple variable."""

    STRUCT_TYPES = {
        BOOLEAN: struct.Struct("?"),
        INTEGER8: struct.Struct("b"),
        INTEGER16: struct.Struct("<h"),
        INTEGER32: struct.Struct("<l"),
        INTEGER64: struct.Struct("<q"),
        UNSIGNED8: struct.Struct("B"),
        UNSIGNED16: struct.Struct("<H"),
        UNSIGNED32: struct.Struct("<L"),
        UNSIGNED64: struct.Struct("<Q"),
        REAL32: struct.Struct("<f"),
        REAL64: struct.Struct("<d")
    }`

**Full Copy of Table 44 from CiA Documentation:**
Index Object Name
0001h DEFTYPE BOOLEAN
0002h DEFTYPE INTEGER8
0003h DEFTYPE INTEGER16
0004h DEFTYPE INTEGER32
0005h DEFTYPE UNSIGNED8
0006h DEFTYPE UNSIGNED16
0007h DEFTYPE UNSIGNED32
0008h DEFTYPE REAL32
0009h DEFTYPE VISIBLE_STRING
000Ah DEFTYPE OCTET_STRING
000Bh DEFTYPE UNICODE_STRING
000Ch DEFTYPE TIME_OF_DAY
000Dh DEFTYPE TIME_DIFFERENCE
000Eh reserved
000Fh DEFTYPE DOMAIN
0010h DEFTYPE INTEGER24
0011h DEFTYPE REAL64
0012h DEFTYPE INTEGER40
0013h DEFTYPE INTEGER48
0014h DEFTYPE INTEGER56
0015h DEFTYPE INTEGER64
0016h DEFTYPE UNSIGNED24
0017h reserved
0018h DEFTYPE UNSIGNED40
0019h DEFTYPE UNSIGNED48
001Ah DEFTYPE UNSIGNED56
001Bh DEFTYPE UNSIGNED64
001Ch – 001Fh reserved
0020h DEFSTRUCT PDO_COMMUNICATION_PARAMETER
0021h DEFSTRUCT PDO_MAPPING
0022h DEFSTRUCT SDO_PARAMETER
0023h DEFSTRUCT IDENTITY
0024h – 003Fh reserved
0040h – 005Fh DEFSTRUCT Manufacturer-specific Complex Data types
0060h – 007Fh DEFTYPE Device profile specific Standard Data types 1st logical device
0080h – 009Fh DEFSTRUCT Device profile specific Complex Data types 1st logical device
00A0h – 00BFh DEFTYPE Device profile specific Standard Data types 2nd logical device
00C0h – 00DFh DEFSTRUCT Device profile specific Complex Data types 2nd logical device
00E0h – 00FFh DEFTYPE Device profile specific Standard Data types 3rd logical device
0100h – 011Fh DEFSTRUCT Device profile specific Complex Data types 3rd logical device
0120h – 013Fh DEFTYPE Device profile specific Standard Data types 4th logical device
0140h – 015Fh DEFSTRUCT Device profile specific Complex Data types 4th logical device
0160h – 017Fh DEFTYPE Device profile specific Standard Data types 5th logical device
0180h – 019Fh DEFSTRUCT Device profile specific Complex Data types 5th logical device
01A0h – 01BFh DEFTYPE Device profile specific Standard Data types 6th logical device
01C0h – 01DFh DEFSTRUCT Device profile specific Complex Data types 6th logical device
01E0h – 01FFh DEFTYPE Device profile specific Standard Data types 7th logical device
0200h – 021Fh DEFSTRUCT Device profile specific Complex Data types 7th logical device
0220h – 023Fh DEFTYPE Device profile specific Standard Data types 8th logical device
0240h – 025Fh DEFSTRUCT Device profile specific Complex Data types 8th logical device

### LINKED ISSUE #437: Object Dictionary module | UNSIGNED48 | CANOpen Lift 417 IO-Port EDS
Please add UNSIGNED48

File : ~/.local/lib/python3.11/site-packages/canopen/objectdictionary/__init__.py

**TOP Part**

```
UNSIGNED48 = 0x19  # Assuming 0x19 is the identifier for UNSIGNED48

def unpack_unsigned48(data):
    """Unpack 48-bit unsigned integer from bytes."""
    assert len(data) == 6, "UNSIGNED48 data should be exactly 6 bytes"
    return int.from_bytes(data, byteorder='little', signed=False)

def pack_unsigned48(value):
    """Pack 48-bit unsigned integer into bytes."""
    assert 0 <= value < (1 << 48), "Value out of range for UNSIGNED48"
    return value.to_bytes(6, byteorder='little')
```

**Decode-Part:**
 ```
   def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
        if self.data_type == UNSIGNED48:
            return unpack_unsigned48(data)
        elif self.data_type == VISIBLE_STRING:
            return data.rstrip(b"\x00").decode("ascii", errors="ignore")
        elif self.data_type == UNICODE_STRING:
            # Is this correct?
            return data.rstrip(b"\x00").decode("utf_16_le", errors="ignore")
            try:
                value, = self.STRUCT_TYPES[self.data_type].unpack(data)
                return value
            except struct.error:
                raise ObjectDictionaryError(
                    "Mismatch between expected and actual data size")
        else:
            # Just return the data as is
            return data
```
`
**Encode Part**
 ```
   def encode_raw(self, value: Union[int, float, str, bytes, bytearray]) -> bytes:
        if isinstance(value, (bytes, bytearray)):
            return value

        elif self.data_type == UNSIGNED48:
            return pack_unsigned48(value)


        elif self.data_type == VISIBLE_STRING:
            return value.encode("ascii")
        elif self.data_type == UNICODE_STRING:
            # Is this correct?
            return value.encode("utf_16_le")
                value = int(value)
            if self.data_type in NUMBER_TYPES:
                if self.min is not None and value < self.min:
                    logger.warning(
                        "Value %d is less than min value %d", value, self.min)
                if self.max is not None and value > self.max:
                    logger.warning(
                        "Value %d is greater than max value %d",
                        value,
                        self.max)
            try:
                return self.STRUCT_TYPES[self.data_type].pack(value)
            except struct.error:
                raise ValueError("Value does not fit in specified type")
        elif self.data_type is None:
        else:
            raise TypeError(
                "Do not know how to encode %r to data type %Xh" % (
                    value, self.data_type))

```
`

sveinse:
* Add datatype defs for the canopen standard types
* Rename datatypes_24bit.py to datatypes_struct.py
* Replace Unsigned24 and Interger24 by generic UnsignedN and IntegerN respectively
* Add EDS-file containing all datatypes
* Added tests for encoding and decoding all datatypes
* Added tests for SDO uploads of all datatypes

Related to #436 and #439. This issue will close #437. Closes #320


sveinse:
There are some notable comments for this PR.

1) The PR currently fails because of the issue in #436. It fails because this PR contains a test that is affected by it. So I need to find a resolution to it before this PR can be finished.

2) I've taken the liberty to remove `Unsigned24` and `Integer24` and replace them by new generic classes `UnsignedN` and `IntegerN`. I grepped for it and it doesn't seem to be used anywhere else.

3) The file `datatypes_24bit.py` to `datatypes_struct.py` since the name is more generic.

sveinse:
BTW: The following code asks if 'utf-16-le' is correct for `UNICODE_STRING`. The canopen standard doesn't define any encoding variant to use, except that it should be 16-bit glyphs. I think its a good assumption and choice to use utf-16-le here.

https://github.com/christiansandberg/canopen/blob/24df6e846a4e71508fd82c378d1385d65f000c94/canopen/objectdictionary/__init__.py#L388-L390

canopen/objectdictionary/__init__.py
```diff
@@ -387,7 +395,7 @@ def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
             return data.rstrip(b"\x00").decode("ascii", errors="ignore")
         elif self.data_type == UNICODE_STRING:
             # Is this correct?
-            return data.rstrip(b"\x00").decode("utf_16_le", errors="ignore")
+            return data.decode("utf_16_le", errors="ignore").rstrip("\x00")
```
acolomb, line 398: What's this change about?  Please adjust the comment accordingly.

It does make sense IMHO to strip a trailing NUL character, and after decoding that's more sensible.  The type is defined as `ARRAY [length] OF UNSIGNED16`, thus a single NUL cannot be trailing.  And if the last byte happens to be a NUL, then removing it definitely breaks decoding the last glyph.  If there is geniunely a single trailing NUL byte in the input, it will be ignored anyway by the decode option.

canopen/objectdictionary/__init__.py
```diff
@@ -387,7 +395,7 @@ def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
             return data.rstrip(b"\x00").decode("ascii", errors="ignore")
```
acolomb, line 395: Why not switch around the stripping and decoding here as well, for consistency?

canopen/objectdictionary/__init__.py
```diff
@@ -407,6 +415,8 @@ def encode_raw(self, value: Union[int, float, str, bytes, bytearray]) -> bytes:
         elif self.data_type == UNICODE_STRING:
             # Is this correct?
```
acolomb, line 416: I'd assume this *IS* correct.  UNSIGNED16 is defined as little-endian and an array of that equals UNICODE_STRING.  Could remove the question.

canopen/objectdictionary/datatypes_struct.py
```diff
@@ -0,0 +1,62 @@
+import struct
```
acolomb, line 1: Wouldn't it be better named `datatypes_integer`?

canopen/objectdictionary/datatypes_struct.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValueError("Width must be a multiple of 8")
+        if width <= 0 or width > 64:
+            raise ValueError("Invalid width for UnsignedN")
+        elif width <= 8:
+            self.__st = struct.Struct("B")
+        elif width <= 16:
+            self.__st = struct.Struct("<H")
+        elif width <= 32:
+            self.__st = struct.Struct("<L")
+        elif width <= 64:
```
acolomb, line 19: I would prefer `else` here, makes it easier to reason about `__st` always being initialized in an instance.

canopen/objectdictionary/datatypes_struct.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValueError("Width must be a multiple of 8")
+        if width <= 0 or width > 64:
+            raise ValueError("Invalid width for IntegerN")
+        elif width <= 8:
+            self.__st = struct.Struct("b")
+        elif width <= 16:
+            self.__st = struct.Struct("<h")
+        elif width <= 32:
+            self.__st = struct.Struct("<l")
+        elif width <= 64:
```
acolomb, line 49:
```suggestion
        else:
```

canopen/objectdictionary/datatypes_struct.py
```diff
@@ -0,0 +1,62 @@
...
+            self.__st = struct.Struct("<q")
+
+    def unpack(self, __buffer):
```
acolomb, line 52: What's the deal with a dunder for the argument name?

test/all_datatypes.eds
```diff
@@ -0,0 +1,310 @@
+[FileInfo]
+FileName=all.eds
+FileVersion=1
+FileRevision=1
+EDSVersion=4.0
+Description=OD that implements all available object types
```
acolomb, line 6: Not really accurate, it's all of the numeric *data* types, plus some compound types.

test/test_sdo.py
```diff
@@ -2,8 +2,11 @@
 import unittest
 # import binascii
 import canopen
+from canopen.objectdictionary import ODVariable
+import canopen.objectdictionary.datatypes as dt
 
 EDS_PATH = os.path.join(os.path.dirname(__file__), 'sample.eds')
+DATAEDS_PATH = os.path.join(os.path.dirname(__file__), 'all_datatypes.eds')
```
acolomb, line 9: Why not simply integrate the data types test objects into the existing sample?

test/test_sdo.py
```diff
@@ -164,5 +167,337 @@ def test_add_sdo_channel(self):
         self.assertIn(client, self.network[2].sdo_channels)
 
 
+class TestSDOClientDatatypes(unittest.TestCase):
+    """FIXME"""
```
acolomb, line 171: FIXME.

acolomb (changes requested):
Looks good overall, with some outstanding issues noted below.

I haven't checked the tests thoroughly, but trusting in your diligence that they're correct.

canopen/objectdictionary/__init__.py
sveinse: BTW: Technically NUL is not specified as termination of string in the canopen standard, so this is a feature which the canopen python module adds on top. This is the same choice that it use utf_16_le, which is also not specified in the standard.

test/all_datatypes.eds
sveinse: yes, good point. Its missing `TIME_OF_DAY`,  `TIME_DIFFERENCE`, `PDO_COMMUNICATION_PARAMETER`, `PDO_MAPPING`, `SDO_PARAMETER`.

canopen/objectdictionary/datatypes_struct.py
sveinse: Inheritance to what was used in `Integer24` and `Unsigned24`.

canopen/objectdictionary/__init__.py
sveinse: I think order doesn't matter, as "ascii" doesn't encode `\x00`

canopen/objectdictionary/__init__.py
sveinse: Canopen does not specify _which_ 16-bit encoding to use, only that the glyphs are 16-bit little-endian symbols, so there is design choice taken here.

canopen/objectdictionary/datatypes_struct.py
sveinse: I was thinking if we get any future data types which are not limited to integer

test/test_sdo.py
sveinse: Separation of concerns. I didn't want to taint the existing test cases that sample uses.

sveinse:
Comments updated

canopen/objectdictionary/__init__.py
acolomb: Yes, I'm aware of that.  Still it will be common to get NUL-terminated strings from C-based device implementations, and we don't want that terminator to leak into Python strings.  Thus it makes sense to do it like this.

canopen/objectdictionary/__init__.py
acolomb: Right, it doesn't mention UTF-16 explicitly.  But it's definitely the most common choice for 16-bit unicode.  And the little-endian part is clear from the fact that it's UNSIGNED16 in CANopen.

So I think the comment can go, let's wait for the first user to hit any other encoding before we start thinking about a possible alternative.

canopen/objectdictionary/datatypes_struct.py
acolomb: Well, there is no good reason for splitting out the module `datatypes_24bit` in the first place, other than it was the choice of the PR author.  We could just as well move the class definitions to `datatypes.py` directly.  They're small enough and don't introduce any new dependencies.

canopen/objectdictionary/datatypes_struct.py
acolomb: That's not a good reason :-)  Please simplify to more Pythonic expectations.

test/all_datatypes.eds
acolomb: My main concern was the "object types" vs. "data types" mixup.

test/test_sdo.py
acolomb: Alright.

canopen/objectdictionary/__init__.py
sveinse: What I meant was `data.rstrip(b"\x00").decode("ascii") == data.decode("ascii").rstrip("\x00")`

acolomb: That's clear.  It just seems clearer if the order of actions is the same as for the 16-bit case.

canopen/objectdictionary/datatypes_struct.py
sveinse: I like that proposal. Let me do that.

canopen/objectdictionary/__init__.py
```diff
@@ -7,7 +7,7 @@
 import logging
 
 from canopen.objectdictionary.datatypes import *
-from canopen.objectdictionary.datatypes_24bit import Integer24, Unsigned24
+from canopen.objectdictionary.datatypes_struct import IntegerN, UnsignedN
```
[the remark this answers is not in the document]
sveinse: That is actually not as straight forward as I first thought it was. The width is bound when the instance is created. Setting `Integer24 = IntegerN(24)` is not right, as the former is a class while the second is an instance. I _could_ perhaps make a type factory, but it quickly becomes hairy.

sveinse:
Updates pushed. There are currently two open issues

* How to deal with `Integer24` and `Unsigned24` compatibility
* The outcome of the data truncation issue as described in #436

canopen/objectdictionary/__init__.py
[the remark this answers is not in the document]
acolomb: Is there something like `functools.partial()` for class constructors?

Or you could derive from it with an overridden constructor.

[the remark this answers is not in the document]
sveinse: Yeah, maybe that's easiest:

```python
class Integer24(IntegerN):
    def __init__(self):
        super().__init__(24)
```
It ensures that `Integer24` is a class (and not a `functool.partial` or similar) and its instantiable. It violates the Liskov principle that some linters like to point out.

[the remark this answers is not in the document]
acolomb: I don't think it violates the Liskov Principle. That talks about properties of object instances, while the constructor is explicitly for *creating* an instance of a concrete type.  See for example https://stackoverflow.com/questions/5490824/should-constructors-comply-with-the-liskov-substitution-principle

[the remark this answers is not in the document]
sveinse: I'll add it, no problem.

But let me comment that I think this is unnecessary. The `Integer24` and `Unsigned24` classes exists solely for filling in `STRUCT_TYPES` for encoding. Should it even be used by users? The principal question is: Do we _always_ need to be backwards compatible? It's a strict and often a difficult requirement to put on ourselves. Is the risk of breakage very large in this instance?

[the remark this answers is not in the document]
acolomb: Well, I don't know. Just assuming that the user who contributed the feature needed it in their code. And I have no idea what that code looks like or in which way the classes are used. But pulling the rug after they have made the effort of getting it added upstream is certainly not a nice surprise.

Of course the class itself is rather meaningless, you mainly need the type code constant externally. But let's ask the user who contributed it at least, whether the internal class is needed as part of the public API use.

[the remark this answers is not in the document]
sveinse: @raffi-g This PR adds support for the remaining integer datatypes to canopen. As can be seen from the changeset, this is done by implementing new generic `IntegerN` and `UnsignedN` class that supports all types, including 24-bit. Thus the `Integer24` and `Unsigned24` implemented in PR #406 can be superseded. The question is if we need to keep these two classes for compatibility or if they can be replaced. As the author for the 24-bit classes, do you have any objections to deleting them?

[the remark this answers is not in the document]
raffi-g: Thanks for reaching out. I have no objections to the changes, as I don't use the two 24bit classes directly in my application.

canopen/objectdictionary/datatypes.py
```diff
@@ -10,16 +11,89 @@
...
 INTEGER_TYPES = SIGNED_TYPES + UNSIGNED_TYPES
 FLOAT_TYPES = (REAL32, REAL64)
 NUMBER_TYPES = INTEGER_TYPES + FLOAT_TYPES
 DATA_TYPES = (VISIBLE_STRING, OCTET_STRING, UNICODE_STRING, DOMAIN)
+
+
+class UnsignedN:
+    """ struct-like class for packing and unpacking unsigned integers of arbitrary width.
+    The width must be a multiple of 8 and must be between 8 and 64.
+    """
```
acolomb, line 43:
```suggestion
    """Packing and unpacking unsigned integers of arbitrary width, like struct.Struct.

    The width must be a multiple of 8 and must be between 8 and 64.
    """
```

canopen/objectdictionary/datatypes.py
```diff
@@ -10,16 +11,89 @@
...
+
+    @property
+    def size(self):
+        return self.width // 8
+
+
+class IntegerN:
+    """ struct-like class for packing and unpacking integers of arbitrary width.
+    The width must be a multiple of 8 and must be between 8 and 64.
+    """
```
acolomb, line 73:
```suggestion
    """Packing and unpacking integers of arbitrary width, like struct.Struct.

    The width must be a multiple of 8 and must be between 8 and 64.
    """
```

test/test_sdo.py
```diff
@@ -164,5 +167,338 @@ def test_add_sdo_channel(self):
         self.assertIn(client, self.network[2].sdo_channels)
 
 
+class TestSDOClientDatatypes(unittest.TestCase):
+    """Test the SDO client uploads with the different data types in CANopen.
+    """
```
acolomb, line 172:
```suggestion
    """Test the SDO client uploads with the different data types in CANopen."""
```

acolomb (changes requested):
Thanks for considering the previous feedback.  Some styling nit-picks remain, and an actual fix of wrong numeric data type constants.

Regarding the 24-bit compatibility classes, I'm okay with dropping them after @raffi-g's feedback.

I've commented on #436 with my current understanding of the problem, and a possibly easy fix.  Should be able to get that squared away soon.

canopen/objectdictionary/datatypes.py
```diff
@@ -10,16 +11,89 @@
...
+INTEGER56 = 0x14
 INTEGER64 = 0x15
 UNSIGNED24 = 0x16
+UNSIGNED40 = 0x18
+UNSIGNED48 = 0x19
+UNSIGNED56 = 0x1A
 UNSIGNED64 = 0x1B
+PDO_COMMUNICATION_PARAMETER = 0x1C
+PDO_MAPPING = 0x1D
+SDO_PARAMETER = 0x1E
```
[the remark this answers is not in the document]
sveinse: Hah! Nice catch! Thanks.

canopen/objectdictionary/__init__.py
```diff
@@ -7,7 +7,6 @@
 import logging
 
 from canopen.objectdictionary.datatypes import *
-from canopen.objectdictionary.datatypes_24bit import Integer24, Unsigned24
```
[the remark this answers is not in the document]
sveinse: `import *` logic is not recommended in pep 8 and thus many linters complain about it. So perhaps we should go by a var `from canopen.objectdictionary import datatypes as dt` or some similar mechanism.

canopen/objectdictionary/__init__.py
```diff
@@ -384,10 +391,13 @@ def add_bit_definition(self, name: str, bits: List[int]) -> None:
...
     def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
         if self.data_type == VISIBLE_STRING:
-            return data.rstrip(b"\x00").decode("ascii", errors="ignore")
+            # Strip any trailing NUL characters from C-based systems
+            return data.decode("ascii", errors="ignore").rstrip("\x00")
         elif self.data_type == UNICODE_STRING:
-            # Is this correct?
-            return data.rstrip(b"\x00").decode("utf_16_le", errors="ignore")
+            # The canopen standard does not specify the encoding. This
+            # library assumes utf-16-le, being the most common encoding format.
```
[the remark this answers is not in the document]
sveinse: "UTF-16" is ambiguous without byte order. When using IANA naming, it should be "UTF-16LE".

canopen/objectdictionary/__init__.py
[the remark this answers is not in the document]
acolomb: Yes, that would be cleaner.  Time for a different PR though, I'd say.

canopen/objectdictionary/__init__.py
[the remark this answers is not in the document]
acolomb: Well, that's what I've been saying, the LE part is strictly defined by CANopen, because `ARRAY OF UNSIGNED16` means each glyph is represented as a 16-bit number in little-endian byte order.  What is not defined by the standard is the encoding (mapping of glyphs to numbers), so the comment should make clear that this little bit is the library's design choice.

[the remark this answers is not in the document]
sveinse: Yes, I think we agree, so maybe we talk past each other. I simply meant to change line 398 from "UTF-16" to "UTF-16LE". Regardless of any chose encoding extensions, reading the stream as big-endian will fail miserably, so it must be specified as LE.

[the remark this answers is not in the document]
acolomb: Of course the code needs to specify utf-16-le, as it does.  I was just pointing out that the little-endian part is NOT a design choice, but mandated by the standard.  Therefore IMO the comment should be restricted to the encoding part, to make it clear that UTF-16 is *chosen*, leading to utf-16-le combined with the *specification*.  It was a deliberate choice to leave out the LE part when talking about the design choice.

sveinse:
Should we disable the two failing tests and make a new PR from the result of the discussions in #436? This way we can complete this PR and focus on solving the other.
```
FAILED test/test_sdo.py::TestSDOClientDatatypes::test_unknown_datatype112 - AssertionError: b'\xb2' != b'\xb2\x01 \x02\x91\x12\x03\x19!p\xfe\xfd\xfc\xfb'
FAILED test/test_sdo.py::TestSDOClientDatatypes::test_unknown_datatype32 - AssertionError: b'\xfe' != b'\xfe\xfd\xfc\xfb'
```

sveinse:
Please disregard the "Proposed fix for..." reference. It was a hasty mistake.

acolomb:
> Should we disable the two failing tests and make a new PR [...]?

Testing the solution to #436 requires the same tests.  Ideally we'd have one commit where they fail, then the fix.  And finally squash both when merging.  I guess disabling with a `FIXME` is appropriate here, since work on the other issue is ongoing and foreseeable to fix them soon.

canopen/objectdictionary/datatypes.py
```diff
@@ -10,16 +11,92 @@
...
-UNSIGNED_TYPES = (UNSIGNED8, UNSIGNED16, UNSIGNED24, UNSIGNED32, UNSIGNED64)
+SIGNED_TYPES = (INTEGER8, INTEGER16, INTEGER24, INTEGER32, INTEGER40, INTEGER48, INTEGER56, INTEGER64)
+UNSIGNED_TYPES = (UNSIGNED8, UNSIGNED16, UNSIGNED24, UNSIGNED32, UNSIGNED40, UNSIGNED48, UNSIGNED56, UNSIGNED64)
 INTEGER_TYPES = SIGNED_TYPES + UNSIGNED_TYPES
 FLOAT_TYPES = (REAL32, REAL64)
 NUMBER_TYPES = INTEGER_TYPES + FLOAT_TYPES
 DATA_TYPES = (VISIBLE_STRING, OCTET_STRING, UNICODE_STRING, DOMAIN)
+
+
+class UnsignedN:
```
acolomb, line 41: I just noticed that using these classes (just as the previous 24-bit only variants) would break with PDO that are not aligned to full bytes.  That's because the code there actually examines the `.format` attribute, which is only available on the real `Struct` class:

https://github.com/christiansandberg/canopen/blob/94f337d779c9294db99083821a6a834fe58b167a/canopen/pdo/base.py#L567

Let's add a read-only `str` property to forward `.format` to the `self.struct.format` instance variable.

sveinse: What _exactly_ is this test doing? It tests that the lower-case of the struct format string is not empty? So is it then a method to determine all types that are struct native, i.e. not 24, 48, 56 and so on? This code is very strange to me. The comment sais "Check if the variable is signed and if the data is negative prepend signedness". I think I need some enlightenment here.

acolomb: The test hinges on the implementation detail that the struct mini-language uses uppercase codes (B, H, L, Q) for unsigned and lowercase (b, h, l, q) for signed types.  It then goes on to check whether the value is in the upper half of the representable number space (with the specified number of bits), which indicates that it was a negative number being truncated.  In that case, padding with 1-bits is added to the left to make it a truly negative integer (in two's complement) after converting back to the full-width integer type.

This is needed for partially mapped signed integer variables.  Judging from the name, the `STRUCT_TYPES` listing was meant to indicate we do have a `struct`-based en- / decoder available, thus it can be assumed that the uppercase / lowercase distinction works.

I think the easiest way to fix this is to actually inherit from `struct.Struct` in your generic integer classes, so that we do have access to a real `.format` attribute again.  Inheritance instead of composition, as an OOP specialist would put it.

sveinse: Ahhh. It sais `islower()` not `lower()`... Sorry, I misread it.

I agree, let's implement some upper case letter for unsigned and lower case for signed. I think simply proxying `format` to `self.struct.format` in the `UnsignedN` and `SignedN` object will suffice.

acolomb: Hm. I cannot say I share that experience.  Tests seem to pass alright with the above implementation.

sveinse: If the code is safe, I'd say this is good. -- There are probably other ways to declare `class UnsignedN` to be `struct.Struct`-like so that a `dict[int, Struct]` annotation would work.

Allow me to be direct, @erlend-aasland: Are there any reasons to shy away from inheriting cpython classes, such as `struct.Struct` using the above code?

erlend-aasland: `struct.Struct` is a base type; you can safely inherit from it. Builtin types and stdlib types that are not safe to inherit from are not base types (or _should not_ be base types). See also:

- https://docs.python.org/3.14/c-api/typeobj.html#c.Py_TPFLAGS_BASETYPE

I see now that that page needs some care (yes, the triple question marks...)

erlend-aasland: FTR, the `struct.Struct` type spec can be found here:

https://github.com/python/cpython/blob/4055577221f5f52af329e87f31d81bb8fb02c504/Modules/_struct.c#L2239-L2246

acolomb: Wow, it never crossed my mind that a Python class in the stdlib could NOT be subclassed.  If such a thing exists, I hope there is a big fat runtime or compiler warning, besides being clearly stated in the documentation.

erlend-aasland: You'll get a `TypeError`:

```python
Python 3.13.0b1 (v3.13.0b1:2268289a47, May  8 2024, 06:41:53) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> class C(datetime.timezone): ...
... 
  File "<python-input-1>", line 1, in <module>
    class C(datetime.timezone): ...
TypeError: type 'datetime.timezone' is not an acceptable base type
>>> 
```

For most user-visible types, there is a notice of some kind in the docs:

- https://docs.python.org/3/library/datetime.html#datetime.tzinfo

erlend-aasland: Anyway (continuing the digression): most non-base types are not user visible. For example, the `_struct.unpack_iter` type:

```python
Python 3.13.0b1 (v3.13.0b1:2268289a47, May  8 2024, 06:41:53) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import struct
>>> tp = type(struct.iter_unpack("b", b'x'))
>>> tp.__name__
'unpack_iterator'
>>> class C(tp): ...
... 
  File "<python-input-3>", line 1, in <module>
    class C(tp): ...
TypeError: type '_struct.unpack_iterator' is not an acceptable base type
>>> 
```

acolomb:
@sveinse I feel pretty confident this is going in the right direction and will work well except for the failure #436, which we should postpone until after a short-term release.  But if conflicts are resolved and the failing tests masked out (temporarily), IMHO we could target this for the upcoming release as well.  The errors only do manifest under certain conditions, right?  So we could at least deliver support for the additional types, even if still buggy in some edge-cases?

WDYT @christiansandberg?

sveinse:
@acolomb It would be awesome to have this included in the release.

The failing tests _only_ fail because of the lack of support for custom datatypes as per #436. It would be fine to disable these two tests for now and just state that canopen doesn't support custom types (as it never has).

I'm a bit afk this week, so if you'd like to add the struct proposal, please do so. I believe the PR is opened for maintainer push. Thanks!

christiansandberg (approved):

acolomb (approved):

PR #441, opened 2024-05-19
Minor linting nitpicks and Python 2 removal

### LINKED ISSUE #346: Drop Python 2.7 support for better code quality and support of new language features
Hi,

I would like to propose to drop the support for python 2.7. Official support for this version (EOL) ended on 31.12.2020 which was three years ago and over 12 years after 2.7 was released and over 20 years after Python 2.x. Even Debian oldstable has support for Python 3.7.
My main reason for this suggestion is code quality and the new language features (like type hints!). My PyLint can't handle the mixed style of the current scripts and things like:

```
try:
    from configparser import RawConfigParser, NoOptionError, NoSectionError
except ImportError:
    from ConfigParser import RawConfigParser, NoOptionError, NoSectionError
```

```
try:
    # Python 3
    eds.read_file(fp)
except AttributeError:
    # Python 2
    eds.readfp(fp)
```

 are ugly and smelly and can lead to further code errors. What do you guys think?

sveinse:
* Removed minor formatting/linting nitpicks that doesn't change any code
* Removed py2 handling in import_eds
* Made file open more robust in import_eds

This PR is a simple cleanup. It removes some old references to python 2, including the open file logic in `import_eds()`. It closes #346


canopen/objectdictionary/eds.py
```diff
@@ -19,19 +19,17 @@
...
+            fp = open(source)
+            opened_here = True
         eds.read_file(fp)
-    except AttributeError:
-        # Python 2
-        eds.readfp(fp)
     finally:
         # Only close object if opened in this fn
-        if not hasattr(source, "read"):
+        if opened_here:
```
acolomb, line 32: The boolean flag seems like a good idea.  But I don't see why anything before the `open()` call should be enclosed in the `try ... finally` block?

acolomb (commented):
Cleanups are alright, but the logic change around `open()` could be simplified.

canopen/objectdictionary/eds.py
sveinse: How would you do it without?

```python
if hasattr(source, "read"):
    fp = source
try:
    if not hasattr(source, "read"):
        fp = open(source)
        opened_here = True
    eds.read_file(fp)
finally:
    ...
````

This is the same line count and a little less DRY, right?

christiansandberg (approved):

canopen/objectdictionary/eds.py
acolomb: Sorry, you're right. It doesn't make much sense to avoid that. I probably misread the diff.

acolomb (approved):

PR #442, opened 2024-05-20
Cleanup logging to use legacy format

sveinse:
This PR cleans up the use of the logger which recommends using the "%s" format and pass its arguments to logger instead of using `.format()` or f-strings.

### DIFF
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -138,7 +138,7 @@ class RemoteNode(BaseNode):
                 # Abort codes other than "Attempt to write a read-only object"
                 # should still be reported.
                 logger.warning('[ERROR SETTING object %#06x:%#06x]  %s',
-                               index, subindex, str(e))
+                               index, subindex, e)
                 raise
 
     def load_configuration(self):
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -122,18 +122,13 @@ class RemoteNode(BaseNode):
         """
         try:
             if subindex is not None:
-                logger.info(str('SDO [{index:#06x}][{subindex:#06x}]: {name}: {value:#06x}'.format(
-                    index=index,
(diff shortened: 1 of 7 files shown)


canopen/node/remote.py
```diff
@@ -142,7 +137,8 @@ def __load_configuration_helper(self, index, subindex, name, value):
             if e.code != 0x06010002:
                 # Abort codes other than "Attempt to write a read-only object"
                 # should still be reported.
-                logger.warning('[ERROR SETTING object {0:#06x}:{1:#06x}]  {2}'.format(index, subindex, str(e)))
+                logger.warning('[ERROR SETTING object %#06x:%#06x]  %s',
+                               index, subindex, str(e))
```
erlend-aasland, line 141: The `%s` format specifier will implicitly convert the argument to a string.

```suggestion
                               index, subindex, e)
```

canopen/profiles/p402.py
```diff
@@ -425,11 +425,11 @@ def op_mode(self, mode):
...
                         "Timeout setting node {0}'s new mode of operation to {1}.".format(
                             self.id, mode))
-            logger.info('Set node {n} operation mode to {m}.'.format(n=self.id, m=mode))
+            logger.info('Set node %s operation mode to %s.', self.id, mode)
         except SdoCommunicationError as e:
-            logger.warning('[SDO communication error] Cause: {0}'.format(str(e)))
+            logger.warning('[SDO communication error] Cause: %s', e)
         except (RuntimeError, ValueError) as e:
-            logger.warning('{0}'.format(str(e)))
+            logger.warning(str(e))
```
erlend-aasland, line 432: Dunno if this one is worth it; perhaps from a consistency point of view 🤷 

```suggestion
            logger.warning("%s", e)
```

PR #443, opened 2024-05-20
Cleanup f-strings, logging and formats

sveinse:
This PR changes all occurrences of `%` formatting and `str.format()` to f-strings.

Please review, as it is easy to sneak in a typo in these changes.

### DIFF
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -127,7 +127,7 @@ class RemoteNode(BaseNode):
                 self.sdo[index][subindex].raw = value
             else:
                 self.sdo[index].raw = value
-                logger.info('SDO [%#06x]: %s: %#06x',
+                logger.info('SDO [0x%04X]: %s: %#06x',
                             index, name, value)
         except SdoCommunicationError as e:
             logger.warning(str(e))
@@ -137,7 +137,7 @@ class RemoteNode(BaseNode):
             if e.code != 0x06010002:
                 # Abort codes other than "Attempt to write a read-only object"
                 # should still be reported.
-                logger.warning('[ERROR SETTING object 0x%04X:%02X]  %s',
+                logger.warning('[ERROR SETTING object 0x%04X:%02X] %s',
                                index, subindex, e)
                 raise
 
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -142,7 +142,9 @@ class RemoteNode(BaseNode):
             if e.code != 0x06010002:
                 # Abort codes other than "Attempt to write a read-only object"
                 # should still be reported.
-                logger.warning('[ERROR SETTING object {0:#06x}:{1:#06x}]  {2}'.format(index, subindex, str(e)))
+                logger.warning('[ERROR SETTING object 0x%#04X:%02X]  %s',
+                               index, subindex, e)
+)
                 raise
 
     def load_configuration(self):
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -122,18 +122,13 @@ class RemoteNode(BaseNode):
         """
         try:
             if subindex is not None:
-                logger.info(str('SDO [{index:#06x}][{subindex:#06x}]: {name}: {value:#06x}'.format(
-                    index=index,
-                    subindex=subindex,
-                    name=name,
-                    value=value)))
+                logger.info('SDO [0x%04X][0x%02X]: %s: %#06x',
+                            index, subindex, name, value)
                 self.sdo[index][subindex].raw = value
             else:
                 self.sdo[index].raw = value
-                logger.info(str('SDO [{index:#06x}]: {name}: {value:#06x}'.format(
-                    index=index,
-                    name=name,
-                    value=value)))
+                logger.info('SDO [%#06x]: %s: %#06x',
+                            index, name, value)
         except SdoCommunicationError as e:
             logger.warning(str(e))
         except SdoAbortedError as e:
@@ -142,9 +137,8 @@ class RemoteNode(BaseNode):
             if e.code != 0x06010002:
                 # Abort codes other than "Attempt to write a read-only object"
                 # should still be reported.
-                logger.warning('[ERROR SETTING object 0x%#04X:%02X]  %s',
+                logger.warning('[ERROR SETTING object 0x%04X:%02X]  %s',
                                index, subindex, e)
-)
                 raise
 
     def load_configuration(self):
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -8,6 +8,7 @@ import logging
 
 from canopen.objectdictionary.datatypes import *
 from canopen.objectdictionary.datatypes_24bit import Integer24, Unsigned24
+from canopen.utils import pretty_index
 
 logger = logging.getLogger(__name__)
 
@@ -108,8 +109,7 @@ class ObjectDictionary(MutableMapping):
             if isinstance(index, str) and '.' in index:
                 idx, sub = index.split('.', maxsplit=1)
                 return self[idx][sub]
-            name = f"0x{index:04X}" if isinstance(index, int) else f"{index!r}"
-            raise KeyError(f"{name} was not found in Object Dictionary")
+            raise KeyError(f"{pretty_index(index)} was not found in Object Dictionary")
         return item
 
     def __setitem__(
@@ -180,12 +180,12 @@ class ODRecord(MutableMapping):
         self.names = {}
 
     def __repr__(self) -> str:
-        return f"<{type(self).__qualname__} {self.name!r} at 0x{self.index:04X}>"
+        return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
 
     def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
         item = self.names.get(subindex) or self.subindices.get(subindex)
         if item is None:
-            raise KeyError(f"Subindex {subindex!r} was not found")
+            raise KeyError(f"Subindex {pretty_index(None, subindex)} was not found")
         return item
 
     def __setitem__(self, subindex: Union[int, str], var: "ODVariable"):
@@ -239,7 +239,7 @@ class ODArray(Mapping):
         self.names = {}
 
     def __repr__(self) -> str:
-        return f"<{type(self).__qualname__} {self.name!r} at 0x{self.index:04X}>"
+        return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
 
     def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
         var = self.names.get(subindex) or self.subindices.get(subindex)
@@ -258,7 +258,7 @@ class ODArray(Mapping):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
         else:
-            raise KeyError(f"Could not find subindex {subindex!r}")
+            raise KeyError(f"Could not find subindex {pretty_index(None, subindex)}")
         return var
 
     def __len__(self) -> int:
@@ -337,8 +337,8 @@ class ODVariable:
         self.pdo_mappable = False
 
     def __repr__(self) -> str:
-        suffix = f":{self.subindex:02X}" if isinstance(self.parent, (ODRecord, ODArray)) else ""
-        return f"<{type(self).__qualname__} {self.qualname!r} at 0x{self.index:04X}{suffix}>"
+        subindex = self.subindex if isinstance(self.parent, (ODRecord, ODArray)) else None
+        return f"<{type(self).__qualname__} {self.qualname!r} at {pretty_index(self.index, subindex)}>"
 
     @property
     def qualname(self) -> str:
@@ -426,7 +426,7 @@ class ODVariable:
             raise ObjectDictionaryError("Data type has not been specified")
         else:
             raise TypeError(
-                f"Do not know how to encode {value!r} to data type {self.data_type:X}h")
+                f"Do not know how to encode {value!r} to data type 0x{self.data_type:X}")
 
     def decode_phys(self, value: int) -> Union[int, bool, float, str, bytes]:
         if self.data_type in INTEGER_TYPES:
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -108,7 +108,7 @@ class ObjectDictionary(MutableMapping):
             if isinstance(index, str) and '.' in index:
                 idx, sub = index.split('.', maxsplit=1)
                 return self[idx][sub]
-            name = f"0x{index:X}" if isinstance(index, int) else index
+            name = f"0x{index:04X}" if isinstance(index, int) else f"{index!r}"
             raise KeyError(f"{name} was not found in Object Dictionary")
         return item
 
@@ -185,7 +185,7 @@ class ODRecord(MutableMapping):
     def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
         item = self.names.get(subindex) or self.subindices.get(subindex)
         if item is None:
-            raise KeyError(f"Subindex {subindex} was not found")
+            raise KeyError(f"Subindex {subindex!r} was not found")
         return item
 
     def __setitem__(self, subindex: Union[int, str], var: "ODVariable"):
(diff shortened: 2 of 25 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -108,8 +108,8 @@ def __getitem__(
             if isinstance(index, str) and '.' in index:
                 idx, sub = index.split('.', maxsplit=1)
                 return self[idx][sub]
-            name = "0x%X" % index if isinstance(index, int) else index
-            raise KeyError("%s was not found in Object Dictionary" % name)
+            name = f"0x{index:X}" if isinstance(index, int) else index
```
acolomb, line 111: Should be padded to four digits?

Don't we already have an attribute for this format? (Sorry, reviewing on mobile, don't have much context right now).

canopen/objectdictionary/eds.py
```diff
@@ -45,7 +45,7 @@ def import_eds(source, node_id):
     if eds.has_section("Comments"):
         linecount = int(eds.get("Comments", "Lines"), 0)
         od.comments = '\n'.join([
-            eds.get("Comments", "Line%i" % line)
+            eds.get("Comments", f"Line{line}")
```
acolomb, line 48: This might break when line is not an integer. Previously, it would raise an exception. The new code happily generates invalid EDS files. So better to put the format specifier also in the F-string?

canopen/objectdictionary/eds.py
```diff
@@ -356,15 +356,15 @@ def export_common(var, eds, section):
...
-            section = "%04X" % var.index
+            section = f"{var.index:04X}"
         else:
             # nested variable
-            section = "%04Xsub%X" % (var.index, var.subindex)
+            section = f"{var.index:04X}sub{var.subindex:X}"
 
         export_common(var, eds, section)
-        eds.set(section, "ObjectType", "0x%X" % VAR)
+        eds.set(section, "ObjectType", f"0x{VAR:X}")
```
acolomb, line 365: Why is this capitalized? Did it ever work?

canopen/objectdictionary/eds.py
```diff
@@ -463,7 +463,7 @@ def export_record(var, eds):
     for rate in od.device_information.allowed_baudrates.union(
             {10e3, 20e3, 50e3, 125e3, 250e3, 500e3, 800e3, 1000e3}):
         eds.set(
-            "DeviceInfo", "BaudRate_%i" % (rate/1000),
+            "DeviceInfo", f"BaudRate_{rate//1000}",
```
acolomb, line 466: Is truncating division here really the same as the previous implicit conversion to int? (or was there an exception for bitrates not divisible by 1000?)

canopen/sdo/client.py
```diff
@@ -487,14 +487,14 @@ def __init__(self, sdo_client, index, subindex=0, request_crc_support=True):
...
             self.sdo_client.abort(0x05040001)
-            raise SdoCommunicationError("Unexpected response 0x%02X" % res_command)
+            raise SdoCommunicationError(f"Unexpected response 0x{res_command:02X}")
         # Check that the message is for us
         if res_index != index or res_subindex != subindex:
             self._error = True
-            raise SdoCommunicationError((
-                "Node returned a value for 0x{:X}:{:d} instead, "
+            raise SdoCommunicationError(
+                f"Node returned a value for 0x{res_index:X}:{res_subindex:d} instead, "
```
acolomb, line 495: Should use hex here as well for the subindex. As previously discussed.

canopen/sdo/client.py
```diff
@@ -656,14 +656,14 @@ def __init__(self, sdo_client, index, subindex=0, size=None, request_crc_support
...
-                "Unexpected response 0x%02X" % res_command)
+                f"Unexpected response 0x{res_command:02X}")
         # Check that the message is for us
         if res_index != index or res_subindex != subindex:
             self.sdo_client.abort()
-            raise SdoCommunicationError((
-                "Node returned a value for 0x{:X}:{:d} instead, "
+            raise SdoCommunicationError(
+                f"Node returned a value for 0x{res_index:X}:{res_subindex:d} instead, "
                 "maybe there is another SDO client communicating "
```
acolomb, line 665: Use hex for subindex?

acolomb (changes requested):
Only checked on the surface, could use some more time when back at a PC keyboard.

There's someone semantic differences you should male sure are intended. And some opportunities to fix inconsistent formats while already touching the lines.

sveinse:
> And some opportunities to fix inconsistent formats while already touching the lines.

Initially I had planned to make this a like-for-like change to ensure no changed behavior. However, if it's wanted (it seems it is) we can fix inconsistencies.

canopen/objectdictionary/__init__.py
sveinse: It sits in a context where we _don't_ have the object, so no attribute I'm afraid. But we can change this to `:04X` as discussed in other issues.

canopen/objectdictionary/eds.py
sveinse: `line` cannot become anything else than int as the line below is `for line in range(1, ...)`.

And I would be worried if the assurance of generating proper EDS hinges on a specific format specifier. If that would be the case, I'd say we need something more explicit like a test and raising an error. ;)

canopen/objectdictionary/eds.py
sveinse: Probably not.

canopen/objectdictionary/eds.py
sveinse: It is the same

```
>>> "%i" % (2900/1000)     # It floors the value
'2'
>>> f"{2900/1000}"     # Don't want this
'2.9'
>>> f"{2900//1000}"    # f-string equivalent
'2'
>>> f"{2900/1000:d}"    # Giving float to :d doesn't work
  File "<stdin>", line 1, in <module>
ValueError: Unknown format code 'd' for object of type 'float'
```

canopen/sdo/client.py
sveinse: Agreed. Will be updated.

sveinse (commented):
@acolomb thank you for very valuable feedback 👌 . I will review and post an update.

sveinse:
I made the liberty to merge PR #442 into this PR and close 442. The PRs did get into each other's feet.

canopen/objectdictionary/__init__.py
acolomb: Yes, please.

canopen/objectdictionary/eds.py
acolomb: Ha ha, easy to confuse with the local variable `var`... Thank you, now it's clear with looking at the full code :grin:

acolomb (changes requested):
It's a lot to review, but I think I've seen all and marked additional fixes spotted.

canopen/objectdictionary/__init__.py
```diff
@@ -185,7 +185,7 @@ def __repr__(self) -> str:
     def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
         item = self.names.get(subindex) or self.subindices.get(subindex)
         if item is None:
-            raise KeyError("Subindex %s was not found" % subindex)
+            raise KeyError(f"Subindex {subindex} was not found")
```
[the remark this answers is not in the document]
acolomb: I'd keep it as-is.  For integers, if there's no `0x` prefix, it can be assumed to be decimal.  Since this is only in an exception message, a developer will need to investigate anyway.

canopen/objectdictionary/__init__.py
```diff
@@ -185,7 +185,7 @@ def __repr__(self) -> str:
     def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
         item = self.names.get(subindex) or self.subindices.get(subindex)
         if item is None:
-            raise KeyError("Subindex %s was not found" % subindex)
+            raise KeyError(f"Subindex {subindex!r} was not found")
```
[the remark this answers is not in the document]
acolomb: This is also a possible pattern for the earlier KeyError instance.

sveinse:
> It's a lot to review, but I think I've seen all and marked additional fixes spotted.

Thank you for the review. Before diving into the comments: I see that we tend to think "while at this, why can't just fix...". I'm worried that by fixing this, we mix the PR that makes no change to behavior with improvements that does. Do we care about that here? Other projects I've been involved with wants to keep them separate.

acolomb:
In general, yes.  No-op formatting or internal changes should be separated from feature work.  But in this case, it'll get ridiculous to touch the same line in two PRs when it's only cleanup.  Especially when there would be an intermediate step (e.g. changing to f-string first, then adjusting for hex output) which no-one cares about.

canopen/profiles/p402.py
```diff
@@ -423,13 +422,12 @@ def op_mode(self, mode):
...
-                            self.id, mode))
-            logger.info('Set node {n} operation mode to {m}.'.format(n=self.id, m=mode))
+                        f"Timeout setting node {self.id}'s new mode of operation to {mode}.")
+            logger.info('Set node %s operation mode to %s.', self.id, mode)
         except SdoCommunicationError as e:
-            logger.warning('[SDO communication error] Cause: {0}'.format(str(e)))
+            logger.warning('[SDO communication error] Cause: %s', e)
         except (RuntimeError, ValueError) as e:
-            logger.warning('{0}'.format(str(e)))
+            logger.warning(str(e))
```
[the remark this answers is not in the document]
sveinse: I'm not expert in DS402, so I'm not completely sure what may trigger these exceptions. I can see the `RuntimeError` a few lines up, but I don't know what can cause a `RuntimeError`. While at it: Both of these exceptions swallow the error. Is that intentional? It removes the user from being able to deal with it (except for log warnings).

canopen/objectdictionary/__init__.py
```diff
@@ -258,7 +258,7 @@ def __getitem__(self, subindex: Union[int, str]) -> "ODVariable":
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
         else:
-            raise KeyError("Could not find subindex %r" % subindex)
+            raise KeyError(f"Could not find subindex {subindex!r}")
```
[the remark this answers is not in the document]
sveinse: I have added this and generally replaced all prints with this one. Please review the next commit.

canopen/objectdictionary/__init__.py
```diff
@@ -427,8 +426,7 @@ def encode_raw(self, value: Union[int, float, str, bytes, bytearray]) -> bytes:
             raise ObjectDictionaryError("Data type has not been specified")
         else:
             raise TypeError(
-                "Do not know how to encode %r to data type %Xh" % (
-                    value, self.data_type))
+                f"Do not know how to encode {value!r} to data type {self.data_type:X}h")
```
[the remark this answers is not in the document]
sveinse: Fixed

canopen/pdo/base.py
```diff
@@ -201,8 +201,8 @@ def __getitem_by_index(self, value):
                 valid_values.append(var.index)
                 if var.index == value:
                     return var
-        raise KeyError('{0} not found in map. Valid entries are {1}'.format(
-            value, ', '.join(str(v) for v in valid_values)))
+        raise KeyError(f'{value} not found in map. Valid entries are ' +
+            ", ".join(str(v) for v in valid_values))
```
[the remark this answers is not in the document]
sveinse: Fixed

acolomb (approved):
I think this is good to go.  Maybe one last look when I'm less tired.  There's two unresolved conversations left, but only with partly unrelated questions, so they shouldn't prevent this from going over the finish line.

canopen/profiles/p402.py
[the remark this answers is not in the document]
acolomb: Not quite sure about the `ValueError`, but it could be catching any `NoneType` access there.  The `RuntimeError` shouldn't be caught, as the docstring explicitly lists it as a possible exception.  But that's material for another PR I'd say.

canopen/profiles/p402.py
```diff
@@ -249,7 +249,7 @@ def _init_tpdo_values(self):
             if tpdo.enabled:
                 tpdo.add_callback(self.on_TPDOs_update_callback)
                 for obj in tpdo:
-                    logger.debug('Configured TPDO: {0}'.format(obj.index))
+                    logger.debug('Configured TPDO: %s', obj.index)
```
acolomb, line 252: This should be formatted as hex `0x%04X` or using `pretty_index()` for consistency.

canopen/profiles/p402.py
```diff
@@ -260,31 +260,31 @@ def _init_rpdo_pointers(self):
         for rpdo in self.rpdo.values():
             if rpdo.enabled:
                 for obj in rpdo:
-                    logger.debug('Configured RPDO: {0}'.format(obj.index))
+                    logger.debug('Configured RPDO: %s', obj.index)
```
acolomb, line 263: This should be formatted as hex `0x%04X` or using `pretty_index()` for consistency.

acolomb (commented):
Two more small issues that I'd like your opinion on.

canopen/profiles/p402.py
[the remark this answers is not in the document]
sveinse: Yes, please. I'm resolving this conversation.

canopen/profiles/p402.py
sveinse: Nice catch! Although `pretty_index()` will be called even if logging is not printed. In a way it negates the reason for having the legacy "%s" syntax. All in all, I think having it outweighs the downside.

canopen/node/remote.py
```diff
@@ -122,18 +122,13 @@ def __load_configuration_helper(self, index, subindex, name, value):
         """
         try:
             if subindex is not None:
-                logger.info(str('SDO [{index:#06x}][{subindex:#06x}]: {name}: {value:#06x}'.format(
-                    index=index,
-                    subindex=subindex,
-                    name=name,
-                    value=value)))
+                logger.info('SDO [0x%04X][0x%02X]: %s: %#06x',
```
sveinse, line 125: @acolomb I found this (and similar prints). I assume we want to fix these as well?

canopen/profiles/p402.py
acolomb: Right, then let's ignore this case.

acolomb: Sorry, that was a misunderstanding.  I thought we agreed on avoiding the function call overhead because logging will very likely not even use the value?

sveinse:
@acolomb I didn't see your latest replies about `pretty_index`, but as you can see, my latest commit contains a wholesale change to `pretty_index` for consistency. I like that we control how index:subindex is printed in one spot. I don't think the overhead is that big (nothing we do here in canopen really is).

acolomb:
> But reading the comments, do we want to revert this one?

I'd vote for reverting, yes.  Sorry for the confusion.  I just misread your "outweighs the downsides" comment.

Note that conditional evaluation of the formatting is not really the only benefit from using the %-formatting for logging.  The biggest advantage I see is that `logging.Handler` instances get these values passed in as distinct objects, thus can do with them whatever is necessary.  Such as translating the message separately or dumping them in some other machine-readable format like JSON, ...

canopen/profiles/p402.py
sveinse: Reverted

canopen/profiles/p402.py
sveinse: Skipping this for the same reason the last commit was reverted

sveinse:
@acolomb I have reverted the last commit and resolved the comments regarding formatting. As far as I can tell, I think I have completed all follow-up tasks, wouldn't you agree?

canopen/profiles/p402.py
acolomb, line 252:
```suggestion
                    logger.debug('Configured TPDO: 0x%04X', obj.index)
```

canopen/profiles/p402.py
acolomb, line 263:
```suggestion
                    logger.debug('Configured RPDO: 0x%04X', obj.index)
```

acolomb (commented):
Sorry for the back and forth on the last two changes.  I simply saw the index (int) being formatted as string and wanted consistency with many other places where we use `0x%04X` formatting in logging calls.  Calling for `pretty_index()` usage here feels overkill as discussed, and as you noticed yourself, there are many other places where that could be applied as well then.

Let's just harmonize on printing in hex in these two cases (like elsewhere) and call it good.

acolomb:
Thank you very much for your patience and diligence in this cleanup effort!  Nice to see people attending to detail and working to reduce our technical debt. :1st_place_medal:

PR #447, opened 2024-05-27
Proposed fix for #440

sveinse:
This is a proposed fix for #440

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -8,6 +8,7 @@ import logging
 
 from canopen.objectdictionary.datatypes import *
 from canopen.objectdictionary.datatypes_24bit import Integer24, Unsigned24
+from canopen.utils import pretty_index
 
 logger = logging.getLogger(__name__)
 
@@ -389,6 +390,13 @@ class ODVariable:
             # Is this correct?
             return data.rstrip(b"\x00").decode("utf_16_le", errors="ignore")
         elif self.data_type in self.STRUCT_TYPES:
+            size = self.STRUCT_TYPES[self.data_type].size
+            if len(data) > size:
+                logger.warning("Excessive data in %s. Data type 0x%X expects %s bytes, got %s",
+                               pretty_index(self.index, self.subindex), self.data_type,
+                               size, len(data))
+                # Truncate the data to the expected size
+                data = data[:size]
             try:
                 value, = self.STRUCT_TYPES[self.data_type].unpack(data)
                 return value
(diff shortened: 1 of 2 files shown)


sveinse:
(It fails checks because it depends on `pretty_index()` from #443, which is a minor detail in this discussion.)

canopen/objectdictionary/__init__.py
```diff
@@ -429,6 +429,16 @@ def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
             # Strip any trailing NUL characters from C-based systems
             return data.decode("utf_16_le", errors="ignore").rstrip("\x00")
         elif self.data_type in self.STRUCT_TYPES:
+            size = self.STRUCT_TYPES[self.data_type].size
+            logger.warning(f"Decoding data {data!r} of type 0x%X at %s",
+                           self.data_type, pretty_index(self.index, self.subindex))
+            logger.warning(f"Size: {size} bytes, data length: {len(data)} bytes")
```
acolomb, line 435: This is definitely too much logging, both by number of messages and at a too high severity level.

And please try to stick to the previously discussed rule: `logging` methods use `%` formats and parameters, and f-strings for everything else.

canopen/objectdictionary/__init__.py
```diff
@@ -429,6 +429,16 @@ def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
...
+            if len(data) > size:
+                logger.warning("Excessive data in %s. Data type 0x%X expects %s bytes, got %s",
+                               pretty_index(self.index, self.subindex), self.data_type,
+                               size, len(data))
+                # Truncate the data to the expected size
+                data = data[:size]
```
acolomb, line 441: It's way easier switching to the `.unpack_from()` method which simply discards trailing extra data.  I don't think this severe warning is generally desired.

acolomb (changes requested):
I've replied to the discussion in #436 concerning the approach taken here.  Will come up with a less intrusive suggestion shortly.

canopen/objectdictionary/__init__.py
sveinse: Yes, I agree. These are left-overs from my own debugging. My bad. I'll remove them.

canopen/objectdictionary/__init__.py
sveinse: I'm going to remove the warning altogether.

sveinse:
Superseded by #591. Closing

PR #451, opened 2024-06-04
Modernize type annotations and fix some discrepancies

acolomb:
* Clarify that each SDO object has its corresponding OD entry linked in the `.od` attribute, not the whole `ObjectDictionary`.

* `__iter__()` should return `Iterator` like in `Mapping`, not `Iterable`.

* Use postponed evaluation of type annotations (PEP-563 / PEP-649 style) instead of stringized type annotations.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
 import threading
 import math
-from typing import Callable, Dict, Iterator, List, Optional, Union
+from typing import Callable, Dict, Iterator, List, Optional, Union, TYPE_CHECKING
 from collections.abc import Mapping
 import logging
 import binascii
@@ -8,6 +9,12 @@ import binascii
 from canopen.sdo import SdoAbortedError
 from canopen import objectdictionary
 from canopen import variable
+
+if TYPE_CHECKING:
+    from canopen.network import Network
+    from canopen import LocalNode, RemoteNode
+    from canopen.pdo import RPDO, TPDO
+    from canopen.sdo import SdoRecord
 
 PDO_NOT_VALID = 1 << 31
 RTR_NOT_ALLOWED = 1 << 30
@@ -22,10 +29,10 @@ class PdoBase(Mapping):
         Parent object associated with this PDO instance
     """
 
-    def __init__(self, node):
-        self.network = None
-        self.map = None  # instance of PdoMaps
-        self.node = node
+    def __init__(self, node: Union[LocalNode, RemoteNode]):
+        self.network: Optional[Network] = None
+        self.map: Optional[PdoMaps] = None
+        self.node: Union[LocalNode, RemoteNode] = node
 
     def __iter__(self):
         return iter(self.map)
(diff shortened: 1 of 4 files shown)


acolomb:
This doesn't include the fixes from #446, but goes in a similar direction.  Could merge those cleanup-only changes into this PR though if you agree @friederschueler.

acolomb:
Integrated the relevant changes from #446 here.  This is certainly not complete yet, for all of the code-base.  But better modernize a tiny bit than never.

acolomb:
@friederschueler, @sveinse would you mind reviewing this change in total?  Would like to get it into the next release, which we should make soon because of the breakage mentioned in #455.

canopen/pdo/base.py
```diff
@@ -22,10 +29,10 @@ class PdoBase(Mapping):
...
     """
 
-    def __init__(self, node):
-        self.network = None
-        self.map = None  # instance of PdoMaps
-        self.node = node
+    def __init__(self, node: Union[LocalNode, RemoteNode]):
+        self.network: Optional[Network] = None
+        self.map: Optional[PdoMaps] = None
+        self.node: Union[LocalNode, RemoteNode] = node
```
sveinse, line 35: What is the intention of having `Union[LocalNode, RemoteNode]` as just using `BaseNode`? Do we expect other `BaseNode` derivatives that are not compatible with `PdoBase`

canopen/sdo/base.py
```diff
@@ -30,7 +31,7 @@ def __init__(
         self,
         rx_cobid: int,
         tx_cobid: int,
-        od: ObjectDictionary,
+        od: objectdictionary.ObjectDictionary,
```
sveinse, line 34: I have read somewhere that using modules for reference to type hints should be avoided. But I've tested the following code with 3.8 and it seems to work. So perhaps I'm misremembering.

```py
import canopen
def foo(a: canopen.Node):
    pass
```
```

sveinse (approved):
@acolomb I think it looks good. Take a look at my two comments, but neither is errors, so I'm approving.

canopen/pdo/base.py
acolomb: Simply because we need the `.sdo` attribute here and that is not part of `BaseNode`.  I previously asked the same question in https://github.com/christiansandberg/canopen/pull/446#discussion_r1615969558.

canopen/sdo/base.py
acolomb: Well if you do find such a recommendation and it has a valid reason, then we can still change that later.  I don't see any reason why we shouldn't do it and as it seems to work (at least mypy doesn't complain here), let's try.

canopen/pdo/base.py
friederschueler: Yes, otherwise we only have network, object_dictionary and id attributes. Also explicit is better than implicit, thats why I don't like defining these attributes in the BaseNode which would be another solution.

sveinse: Great, thanks.

Perhaps we should at some point make a `NodeProtocol` for the required attributes instead of using a exhaustive list? It might not be worth the efforts thou, since the `Union` is only used a few places.

acolomb: Agreed, I think the `Union` approach is simple enough and warranted.  We can still rework it when there's a stronger argument for it.

PR #454, opened 2024-06-04
Test and fix non-byte-aligned PDO mappings

acolomb:
The PDO mapping code supports variables with bit offsets and lengths not a multiple of 8 bits, *in theory*.  However the code is buggy and doesn't actually work correctly on many cases.

Add a test case for some more exotic variations of mappings which occupy bytes only partially, even spanning several PDO bytes with unaligned start and end bits.  Introduce test objects for `UNSIGNED16` and `UNSIGNED32` to use as additional dummies in the test.  Half of the tested objects are signed to see any handling problems there.  The test verifies the PDO byte representation after setting dummy values and the round-trip from decoding these bytes again.

To fix the new test cases, rework `get_data()` and `set_data()` in class `PdoVariable` completely.  More details in the commit messages.


acolomb:
Notice the failing test without any other code changes.

test/test_pdo.py
```diff
@@ -55,6 +55,35 @@ def test_bit_mapping(self):
...
+        # Write some values
+        pdo['UNSIGNED8 value'].raw = 3
+        pdo['INTEGER8 value'].raw = -2
+        pdo['UNSIGNED32 value'].raw = 0x987654
+        pdo['UNSIGNED16 value'].raw = 0x321
+        pdo['INTEGER16 value'].raw = -1
+        pdo['INTEGER32 value'].raw = -1071
+
+        # Check expected data
+        self.assertEqual(pdo.data, b'\xe3\x4f\x65\x87\x19\x32\x8f\xde')
```
sveinse, line 77: Has this encoding been independently verified?

It would be nice with a breakdown of the encoding in the comments here. It took me some time to recognize that the whole PDO is little endian, not just field by field.

acolomb: I calculated it with the help of a spreadsheet:

Offset | Length | Value | Value Hex | Unsigned | High DWORD | Value Hex Shifted | Low DWORD
-- | -- | -- | -- | -- | -- | -- | --
0 | 4 | 3 | 3 | 3 | 00000000 | 0000000000000003 | 00000003
4 | 8 | -2 | FE | 254 | 00000000 | 0000000000000FE0 | 00000FE0
12 | 24 | 9991764 | 987654 | 9991764 | 00000009 | 0000000987654000 | 87654000
36 | 12 | 801 | 321 | 801 | 00003210 | 0000321000000000 | 00000000
48 | 3 | -1 | 7 | 7 | 00070000 | 0007000000000000 | 00000000
51 | 13 | -1071 | 1BD1 | 7121 | DE880000 | DE88000000000000 | 00000000
64 |   |   |   |   |   |   |  

The final shifted values are then easily bit-ORed (manually) and the complete sequence reversed by bytes.

sveinse: Oddly enough, the canopen standard doesn't really seem to mention anything about non 8-bit aliged PDO encoding. I find much chatter on Google that it formally does not support this. Does this imply that there is a defacto standard for it?

(We're using a variant of canfestival (https://github.com/Laerdal/canfestival-laerdal) which does support packing)

What I'd like to double-check is the order of objects. E.g. say the PDO shall contain two 16-bit objects A and B in that order. `A = 0x1234` and `B = 0x5678`. This can either be encoded as b`\x34\x12\x78\x56` or b`\x78\x56\x34\x12`. The example above is the latter format if I'm not mistaking, but the graphics for PDO packing in the canopen standard could be interpreted to the former format. -- But I'm not certain, and I have to verify with actual HW to see what's right.

acolomb: Will have to check and verify with the standard to be sure. From memory, I recall that the device EDS specifies a granularity which sets the lower bound on mapping alignment, e.g. word, byte, nibble or bit. I don't see why it shouldn't be used with 1-bit granularity, which I've seen commonly in other CAN protocols (like J1939, NMEA2000). Note that the existing test is left untouched and its assumptions still hold. It does contain nibble mappings already.

Regarding packing order, I think it's consistent with our API and I implemented on that basis. Basically, to take a side in the Holy Endianness Wars, little-endian is the only sensible way for a good reason. The confusion stems from humans being used to write numbers (hex, decimal, whatever) in big-endian notation, which is the common oddball and makes everything look complicated when talking about little-endian format. As a reminder: little-endian means lowest address holds the least significant digit. 

Our API, and AIUI also the CANopen mapping model, starts from bit zero. Each mapping adds some bits and the running summation is the offset / address of the following mapping. So a boolean needs one byte (the first, which would be LSB in a bigger type) and occupies the lsb therein. Now we are at offset 1 and continue adding e.g. an UNSIGNED8 value, bit by bit. Starting from its lsb=0 at offset 1 up to its msb=7 at offset 8, which ends up as first bit (lsb) of the second PDO byte. Rinse and repeat. 

Those semantics are essentially the same as regarding the whole PDO as a 64-bit integer and mapping a variable by shifting its (numeric) value left (toward higher significance) as much as the cumulated bit offset indicates.

It really makes sense if you imagine every number written in binary and putting down the digits left-to-right, as you would note down a sequence of bytes. (RTL vs. LTR is really interchangeable, it only should match between writing direction in your head and number notation wrt. increasing significance.)

Given this explanation I have a hard time imagining any other logical way to pack up the PDO mappings. What was the other possible systematic again?

acolomb: *Side note to put a nail in big-endian's coffin: Little-endian not only has a linear, non-inverted relation between offset and significance (both monotonically increasing). Which is mostly cosmetic. The main advantage is that each sequentially processed bit keeps its significance over time. You always know the first bit corresponds to 2^0.*

*Whereas with big-endian you would need to know the total bit width in advance or keep counting until all bits have been received, before you know whether the first bit corresponds to 2^15 or whatever else.*

*This applies to endianness with respect to transmitted bytes just the same, of course. If that is your atomic processing width being put in a sequence.*

sveinse: Let me be neutral about it: The choice of little vs big endian is arbitrary. Guess that's why its such a popular flame war topic. Either have advantages and disadvantages. Either is used in the world around us.

acolomb: Here I have to disagree (although irrelevant to the PR). It is not arbitrary because we have TIME as a uni-directional ordering that cannot be arbitrarily reversed. So if we're talking about digits written on a piece of paper - completely arbitrary since the "address space" is freely assignable. But when doing things sequentially, it does matter if what I know at time T is well-defined, or depends on some other info that I will only know at T+x. That's the point I was making above with respect to the bit significance.

Of course, knowing the field width ahead of time (out of band) makes it just a calculation effort but otherwise equivalent. But little-endian transmission is the only sequencing that doesn't require this a priori knowledge to reach the same property.

Enter quantum computing... ;-D

sveinse: DAC and ADCs commonly use big-endian, because the biggest significance bits comes first followed by each bit of lesser and lesser influence. This is how these converters often are implemented physically.

acolomb: Very true and a good example. It definitely has the advantage of knowing the full-scale value in advance though, let alone because the voltages are strictly limited. When filling an address space of known limited size from data of the same known size, it is actually arbitrary from which end you start.

acolomb:
@sveinse, @christiansandberg, @af-silva  this hasn't received much real-world testing yet on my end.  Mainly for lack of a current project where PDOs are used extensively.  So I'd be happy to get some feedback from testing in other applications.

Should we aim to get this into the upcoming release though, or better let it mature and hope there are enough beta testers?

sveinse:
@acolomb I don't have a current (HW) use case for unaligned PDOs, so I am not able to test.

acolomb:
@sveinse Well my main concern is not breaking the currently supported use-cases.  I.e. if you can verify that all your (aligned) PDO mappings are working fine, that would be enough to consider it for merging (after v2.3.0).  The current code for unaligned mappings is broken beyond repair AFAICT, it had actual logic errors besides not supporting all possible cases.  So I suspect nobody really used it, that's why the errors went unnoticed.

sveinse:
I think it is important to note that unless there exists an amendment from CiA about non-aligned PDOs, this is officially not supported by canopen I think. There are canopen implementation might have a defacto practice, but I have no idea if we all do it the same. Have we found any references to any documentation that describes this anywhere?

Unfortunately, we only use the async port of canopen at work, so I'd need to port this over to canopen-asyncio to be able to set this into a suitable use case for testing. TBH, I don't know if I'll get the time this week.

af-silva:
@acolomb On my end is the same, right now I don't have access to HW for testing this, and I don't have any use case for non-aligned PDO's. On my main previous project I had 8 motors being controlled simultaneously using this API but my use case in particular the "regular" PDO's suffice. My main motivation for contributing to this project was that at the time I ended up making changes to the original DS 402 profile, as a way to understand how thinks worked but also to implement my requirements and try to simplify its usage. 
Sorry but regarding this I'm not much of a help.

acolomb:
> I think it is important to note that unless there exists an amendment from CiA about non-aligned PDOs, this is officially not supported by canopen I think. There are canopen implementation might have a defacto practice, but I have no idea if we all do it the same. Have we found any references to any documentation that describes this anywhere?

Section 7.4.8.2 in the CiA 301 spec lays out the "PDO mapping parameter record" datatype (0x21).  It contains up to 64 (0x40) "objects to be mapped".  That only makes sense if the extreme case is filling the PDO with 64 single bits (BOOLEAN values).  So much for the minimum length an object mapping can occupy within the PDO.

Sections 7.5.2.36 and 7.5.2.38 specify "The length contains the length of the application object in bit. This may be used to verify the mapping."  Which indicates that the length parameter within the mapping should not differ from the length of the object's data type.  I.e. "partially" mapped objects (such as only the low word of a 32-bit integer, if 16 bits is enough in practice) are not explicitly allowed.  On the other hand, it is not forbidden either, it just says the length *may* be used for verification, not that it *must* match.

The fact that the length is given in bits here, not bytes, hints toward arbitrary bit lengths being supported.  Same goes for the example in section 7.1.5, where a STRUCT OF (INTEGER10, UNSIGNED5) is defined.  Having such non-aligned lengths automatically leads to non-aligned mapping boundaries.  I.e. if the first mapped object is a single BOOLEAN, all following whole-byte mapped objects will start and end on an odd bit index, therefore spanning partial PDO bytes.

To summarize:
* Partial mappings (less bits than the object's "bit sequence" length) are not mentioned explicitly, but also not forbidden.  We support those already by overriding the `length` parameter in `add_variable()`.  I've actually used such mappings in practice, for the case "16 bits is enough of a 32-bit object".
* Non-aligned mappings are a direct consequence of having data types with bit sequences where the length is not a multiple of 8.  BOOLEAN being the most prominent violation.  That's what this PR is about.  It was supported in theory, but only partially and with lots of broken edge cases.

sveinse:
I concur with the assessment here. It makes sense.

acolomb:
This PR has been sitting for some time, probably without any further real-world testing (mostly that existing applications don't break).  I updated it now to match current master.  If anyone has spare cycles to test this in action, please do so now and holler if anything seems odd.

Otherwise I think we'll bite the bullet and merge in a couple of weeks, to (hopefully) give it some more testing on the master branch before a new release.

PR #456, opened 2024-06-11
Python Warnings and PEP8: unused vars, var naming, typing, spelling etc

friederschueler:
Fixed a lot of small things, I know this is a nightmare to review, but we would go crazy if we open 20 pull requests, right?

the only functional changes are:

    def send_switch_state_selective(self, vendor_id, product_code, revision_number, serial_number)

I changed the signature of the method, upperCase is not allowed according to PEP8.

     def connect(self, channel=None, interface=None, **kwargs)

I also changed this signature, to resemble the signature of the can.Bus class. If you are using the old bustype argument, you will get a warning that you use both, the old and the new "interface" together.

If you don't like this change, we can find another solution. In the beginning the docstring was old and wrong.

### DIFF
--- a/test/test_sdo.py
+++ b/test/test_sdo.py
@@ -110,8 +110,7 @@ class TestSDO(unittest.TestCase):
             (RX, b'\xa1\x00\x00\x00\x00\x00\x00\x00')
         ]
         data = b'A really really long string...'
-        with self.network[2].sdo['Writable string'].open(
-            'wb', size=len(data), block_transfer=True) as fp:
+        with self.network[2].sdo['Writable string'].open('wb', size=len(data), block_transfer=True) as fp:
             fp.write(data)
 
     def test_block_upload(self):
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ target/
 
 # IDEs
 .vscode/
+.idea/
 
 *.dbc
 
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -1,5 +1,5 @@
-import struct
 import logging
+import struct
 import threading
 import time
 from typing import Callable, List, Optional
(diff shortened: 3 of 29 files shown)


acolomb:
Thanks for taking time to clean up!

As you say, it will take some time and effort reviewing this.  I know there is much to be improved and we held back on that to concentrate on actual fixes first.  Will be happy to look at it when we have a new release out, restoring python-can compatibility.

Anyway, I do like not mixing up functional changes with pure cleanup.  I.e. better have several small PRs where we can review a chunk and then finish that part, than having a very large diff and then getting held up by discussions around the functional changes.

test/test_sdo.py
```diff
@@ -110,8 +110,7 @@ def test_block_download(self):
             (RX, b'\xa1\x00\x00\x00\x00\x00\x00\x00')
         ]
         data = b'A really really long string...'
-        with self.network[2].sdo['Writable string'].open(
-            'wb', size=len(data), block_transfer=True) as fp:
+        with self.network[2].sdo['Writable string'].open('wb', size=len(data), block_transfer=True) as fp:
```
sveinse, line 113: Do we have an agreement on line length in canopen? Personally I think it should be more than 80, so I support this change.

sveinse (commented):
First of all, I'd like to say that this is fantastic work. I love that we pick up the code and brush off some old dust and make it up to code. I have a few comments. Some are nitpickings, and some are questions.

acolomb:
I also started a review on this, but didn't get very far.  In short, YES, THIS IS WAY TOO MUCH in a single PR.

Please split it up into digestible pieces and definitely do separate styling fixes from functional changes:

* Import cleanup / ordering.
* Type hinting / `__future__` style annotations.
* Typos (without API changes).
* Additional checks just to make some linter happy.
* Logic errors / behavior changes.

Those are some categories I'd want separated *at the least*.  And for the last category, each change probably deserves its own PR with an explanation of the specific change.  We do care about API compatibility and it's better to have small change sets for being able to dissect / pinpoint where something stopped working.  Also line-blaming is very useful but broken by such "mixed bag" PRs.

Sorry for this extra trouble, but in this case I must insist, to keep review time in a sensible time frame.  And I am aware that I previously merged a PR from you that didn't adhere to these guidelines, for not wanting to hold up progress too much.

Regarding linters, I don't think the project should be enforcing any, but they can be seen as useful tools to mark questionable code.  But answering the questions should still be done by a human developer, even if that means to ignore certain warnings because one knows better.  Silencing them for a specific linter will lead to a mess of linter-specific constructs in the long run, which I'd like to avoid in the name of code readability and making it easier to concentrate on the code's actual logic.  Sometimes a `is not None` check will make the linter happy, but it is not actually necessary and slows down execution as well as reading the code.

friederschueler:
Ok, I will create some smaller Pull requests, but also keep in mind that this increases the workload for the developer and while the overall code structure and quality is pretty good in this library, some parts (p402.py e.g.) really need a lot of work. But yes, API and  other functional-changes will get an extra PR.

And I also agree that you should not preemptively optimize, especially not for metrics produced by linters. On the other hand, from my experience linters are way more accurate than human developers, most of the suggested changes have a valid reason.

Regarding line length:

https://github.com/christiansandberg/canopen/discussions/460

sveinse:
I've noticed a strong correlation between how simple/clean a PR is and how long time it takes to get it merged. And I'd say I have empirical data, as I have contributed with ~14 PRs the last year. I see clearly that clean changes flows much more efficiently than messy ones.  It might feel like a bigger job preparing the changes, but after this number of PRs I'm convinced it makes the workload easier in the end. Both for the contributor and for the reviewers.

PR #457, opened 2024-06-11
Github: Mark issues as stale and close them

friederschueler:
Mark issues as stale after 30 days and close them after 60 more days of inactivity (90 days total)


acolomb:
I don't agree with such a policy.  We need more people triaging issues, not just have them automatically closed when nobody has had time to look through them.

PR #464, opened 2024-06-17
Export PdoMap and PdoVariable classes in pdo module

acolomb:
With commit 0285f582beb344de708ce8308274a99bc04988a1, the `Map` class was inadvertently removed from the imports directly available within the `canopen.pdo` package.  However, it is referenced as `canopen.pdo.PdoMap` in the `p402` module and documentation.

Restore that import using the new names `PdoMap` and `PdoVariable`.  Move the compatibility define to the very bottom, just like in other modules.  Provide an `__all__` listing to clarify what constitutes the public API.

### DIFF
--- a/canopen/pdo/__init__.py
+++ b/canopen/pdo/__init__.py
@@ -1,10 +1,18 @@
 import logging
 
 from canopen import node
-from canopen.pdo.base import PdoBase, PdoMaps
+from canopen.pdo.base import PdoBase, PdoMap, PdoMaps, PdoVariable
 
-# Compatibility
-from canopen.pdo.base import Variable
+
+__all__ = [
+    "PdoBase",
+    "PdoMap",
+    "PdoMaps",
+    "PdoVariable",
+    "PDO",
+    "RPDO",
+    "TPDO",
+]
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


canopen/pdo/__init__.py
```diff
@@ -77,3 +85,7 @@ def stop(self):
                 pdo.stop()
         else:
             raise TypeError('The node type does not support this function.')
+
+
+# Compatibility
+Variable = PdoVariable
```
sveinse, line 91: I wonder if `Variable` should be listed in `__all__`. In principle `__all__` contains the entire public API of the module. However, it's not very common to list objects defined in the file in `__all__` is it?

acolomb: I don't know what's common practice with regard to that `__all__` attribute.  But my understanding is that it controls what `from ... import *` does.  In that case, `Variable` should not be imported, as it is a deprecated alias only kept for compatibility.  But if someone references it directly, that still works.

erlend-aasland: Svein:
> However, it's not very common to list objects defined in the file in `__all__` is it?

That's of course possible; for example, a sentinel or a singleton could also be a part of a public API.

André:
> [...] my understanding is that it controls what `from ... import * does`.
> [...] But if someone references it directly, that still works.

Both are correct. `from canopen.pdo import Variable` still works.

PR #468, opened 2024-06-20
build: Explicitly specify packages to install (fixes #467)

### LINKED ISSUE #467: Packaging fails when having venvs inside the sourctree
If virtual environment is installed inside the canopen code base (which is common practice), building the canopen package will fail. Setuptools seems to incorrectly disover the 3.8 venv as a pacakge and fails. See below. One could argue this is a corner-case, but I think the robustness of the package should be increased.

The root cause is that the `pyproject.toml` is not specifying the package it should install. 

```
PS C:\Work\canopen-wip2> .\venv38\Scripts\python.exe -m pip install -e .                    
Obtaining file:///C:/Work/canopen-wip2
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error

  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [14 lines of output]
      error: Multiple top-level packages discovered in a flat-layout: ['venv38', 'canopen'].

      To avoid accidental inclusion of unwanted files or directories,
      setuptools will not proceed with this build.

      If you are trying to create a single distribution with multiple packages
      on purpose, you should not rely on automatic discovery.
      Instead, consider the following options:

      1. set up custom discovery (`find` directive with `include` or `exclude`)
      2. use a `src-layout`
      3. explicitly set `py_modules` or `packages` with a list of names

      To find more information, look for "package discovery" on setuptools docs.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error
````

### PR DESCRIPTION
**Pull request by sveinse**
This PR fixes #467 when trying to install canopen with an virtual environment inside the source tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ target/
 
 *.kdev4
 *.kate-swp
+
+# Virtual environments
+venv*/
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,3 @@ target/
 
 *.kdev4
 *.kate-swp
-
-# Virtual environments
-venv*/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,9 @@ dynamic = ["version"]
 documentation = "https://canopen.readthedocs.io/en/stable/"
 repository = "https://github.com/christiansandberg/canopen"
 
+[tool.setuptools]
+packages = ["canopen"]
+
 [tool.setuptools_scm]
 version_file = "canopen/_version.py"
 

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -73,3 +73,6 @@ target/
 
 *.kdev4
 *.kate-swp
+
+# Virtual environments
+venv*/
```
acolomb, line 78: Please not.  I'd rather get rid of random peoples' IDE clutter landing on the maintained ignore list.  Git also has a local ignore function if you want to make sure it's not shown in status.  No need to assume that everyone wants to ignore any directory named like `venv*`.

acolomb (changes requested):
Looks good except for the ignore change.  Have you verified the contents of the generated source package / built wheel remain the same, i.e. nothing is suddenly missing?

.gitignore
sveinse: Alright, fair. It _is_ rather standard and many projects choses to, but it's not PEP-ed, so its ok not to. -- But then I would suggest we remove the other IDE excludes as well as you point out.

sveinse:
I have verified that the `-m build` output is equal.

acolomb (approved):

acolomb:
@sveinse This is causing trouble with the v2.4.0 release (see https://github.com/canopen-python/canopen/actions/runs/16746953346/job/47407670531):

```
/tmp/build-env-bi6esm13/lib/python3.13/site-packages/setuptools/command/build_py.py:212: _Warning: Package 'canopen.node' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'canopen.node' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'canopen.node' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'canopen.node' to be distributed and are
        already explicitly excluding 'canopen.node' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************
```

AFAICT, we should not list the top-level package explicitly in order for setuptools to keep doing its automatic discovery.  Ideally, we would add appropriate configuration in pyproject.toml to regain the auto-discovery of subpackages, while limiting the discovery scope to the `canopen` top-level package.

Would you have experience with such errors / configuration?  I'll try to dig into it as well...

sveinse:
@acolomb I'm out travelling these days, so I don't have access to a development setup. I haven't encountered this one before. 

I'm not sure what the trigger from setuptools might be that makes it think this is a separate package. Throwing out some ideas and guesswork:
- Perhaps `from canopen import node` is missing from `canopen/__init__,py` and listed in its `__all__`?
- Perhaps `canopen/node/__init__.py` needs an `__all__ = [LocalNode, RemoteNode]`

PR #469, opened 2024-06-20
Ensure proper close after open in export_od

sveinse:
This PR increases the robustness on file open in export_od(). It ensures that if the function opens a file, it will close it.

PS! The buried import is itching to be fixed, but let's not do that in this PR, please.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -29,29 +29,32 @@ def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str]
     :rtype: str or None
     """
 
-    doctypes = {"eds", "dcf"}
-    if isinstance(dest, str):
-        if doc_type is None:
-            for t in doctypes:
-                if dest.endswith(f".{t}"):
-                    doc_type = t
-                    break
+    opened_here = False
+    try:
+        doctypes = {"eds", "dcf"}
+        if isinstance(dest, str):
+            if doc_type is None:
+                for t in doctypes:
+                    if dest.endswith(f".{t}"):
+                        doc_type = t
+                        break
 
-        if doc_type is None:
-            doc_type = "eds"
-        dest = open(dest, 'w')
-    assert doc_type in doctypes
+            if doc_type is None:
+                doc_type = "eds"
+            dest = open(dest, 'w')
+            opened_here = True
+        assert doc_type in doctypes
 
-    if doc_type == "eds":
-        from canopen.objectdictionary import eds
-        return eds.export_eds(od, dest)
-    elif doc_type == "dcf":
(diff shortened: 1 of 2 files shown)


acolomb (approved):

canopen/objectdictionary/__init__.py
```diff
@@ -29,29 +29,33 @@ def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str]
...
+        doctypes = {"eds", "dcf"}
+        if isinstance(dest, str):
+            if doc_type is None:
+                for t in doctypes:
+                    if dest.endswith(f".{t}"):
+                        doc_type = t
+                        break
+
+            if doc_type is None:
+                doc_type = "eds"
```
erlend-aasland, line 43: This could be a nice case of `for ... else`, if you want to save a few lines.

```suggestion
                for t in doctypes:
                    if dest.endswith(f".{t}"):
                        doc_type = t
                        break
                else:
                    doc_type = "eds"
```

erlend-aasland (approved):

canopen/objectdictionary/__init__.py
sveinse: Ah nice. I've never quite gotten for..else into my fingers, but I like this 👌

acolomb (approved):

canopen/objectdictionary/__init__.py
```diff
@@ -29,29 +29,32 @@ def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str]
...
+            opened_here = True
+
+        if doc_type == "eds":
+            from canopen.objectdictionary import eds
+            return eds.export_eds(od, dest)
+        elif doc_type == "dcf":
+            from canopen.objectdictionary import eds
+            return eds.export_dcf(od, dest)
+        else:
+            raise NotImplementedError(f"No support for the {doc_type!r} format")
```
erlend-aasland, line 53: Hm, I noticed this a little bit too late... this is a little bit unconventional; `NotImplementedError` is not meant for such use. IMO, `ObjectDictionaryError` would be more appropriate.

sveinse: Right. Agreed. It's not always easy to "just fix" something like the close issue and then uncovering other things that needs fixing too. ;)

erlend-aasland: Yeah, one fix quickly leads to the discovery of other issues :) I'll create an issue about this.

erlend-aasland: I created #475.

erlend-aasland: For the record; before this PR, this check was an `assert` (IOW, `AssertionError`, unless running with `-O` or `-OO`).

PR #473, opened 2024-06-29
Expose and improve the docstrings of `export_od` and `import_id`

### LINKED ISSUE #472: `import_od` and `export_od` are undocumented
_Disclaimer: apologies in advance if these are intentionally excluded from the docs._

It seems to me, judging from their docstrings, that `import_od` and `export_od` have been unintentionally excluded from the docs. Adding them is trivial:

```diff
diff --git a/doc/od.rst b/doc/od.rst
index 6040bb5..4b88e5f 100644
--- a/doc/od.rst
+++ b/doc/od.rst
@@ -54,6 +54,10 @@ You can access the objects using either index/subindex or names::
 API
 ---
 
+.. autofunction:: canopen.export_od
+
+.. autofunction:: canopen.import_od
+
 .. autoclass:: canopen.ObjectDictionary
    :members:
```

However, I think this simple patch should be coupled with some minor improvements of the docstrings in question. I can open a PR, unless you prefer these APIs should remain undocumented.

erlend-aasland:
Resolves #472

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -25,12 +25,15 @@ def export_od(
     :param od:
         The object dictionary to be exported.
     :param dest:
-        The export destination as a filename, a file-like object, or None.
+        The export destination as a filename, a file-like object, or ``None``.
         If ``None``, the document is written to :data:`sys.stdout`.
     :param doc_type:
        The type of document to export.
-       If *dest* is a filename, *doc_type* defaults to its extension.
-       Otherwise, the document type defaults to ``"eds"``.
+       If *dest* is a file-like object or ``None``,
+       *doc_type* must be explicitly provided.
+       If *dest* is a filename and its extension is ``".eds"`` or ``".dcf"``,
+       *doc_type* defaults to that extension (the preceeding dot excluded);
+       else, it defaults to ``"eds"``.
     """
 
     opened_here = False
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -15,17 +15,22 @@ from canopen.utils import pretty_index
 logger = logging.getLogger(__name__)
 
 
-def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str] = None):
-    """ Export :class: ObjectDictionary to a file.
+def export_od(
+    od: ObjectDictionary,
+    dest: Union[str, TextIO, None] = None,
+    doc_type: Optional[str] = None
+) -> Union[str, None]:
+    """ Export an object dictionary.
 
     :param od:
-        :class: ObjectDictionary object to be exported
+        The :class:`!ObjectDictionary` object to be exported.
     :param dest:
-        export destination. filename, or file-like object or None.
-        if None, the document is returned as string
-    :param doc_type: type of document to export.
-       If a filename is given for dest, this default to the file extension.
-       Otherwise, this defaults to "eds"
+        The export destination as a filename, a file-like object, or None.
+        If ``None``, the document is returned as string.
+    :param doc_type:
+       The type of document to export.
+       If *dest* is a filename, *doc_type* defaults to its extension.
+       Otherwise, the document type defaults to ``"eds"``.
     :rtype: str or None
     """
 
@@ -61,10 +66,10 @@ def import_od(
     """Parse an EDS, DCF, or EPF file.
 
     :param source:
-        Path to object dictionary file or a file like object or an EPF XML tree.
-
-    :return:
-        An Object Dictionary instance.
+        Path to object dictionary file, a file like object, or an EPF XML tree.
+    :param node_id:
+        For EDS and DCF files, the node ID to use.
+        For other formats, this parameter is ignored.
     """
     if source is None:
         return ObjectDictionary()
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -33,7 +33,7 @@ def export_od(
        *doc_type* must be explicitly provided.
        If *dest* is a filename and its extension is ``.eds`` or ``.dcf``,
        *doc_type* defaults to that extension (the preceeding dot excluded);
-       else, it defaults to ``"eds"``.
+       else, it defaults to ``eds``.
     :raises ValueError:
         When exporting to an unknown format.
     """
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -19,19 +19,18 @@ def export_od(
     od: ObjectDictionary,
     dest: Union[str, TextIO, None] = None,
     doc_type: Optional[str] = None
-) -> Union[str, None]:
+) -> None:
     """Export an object dictionary.
 
     :param od:
         The :class:`!ObjectDictionary` object to be exported.
     :param dest:
         The export destination as a filename, a file-like object, or None.
-        If ``None``, the document is returned as string.
+        If ``None``, the document is written to :data:`sys.stdout`.
     :param doc_type:
        The type of document to export.
        If *dest* is a filename, *doc_type* defaults to its extension.
        Otherwise, the document type defaults to ``"eds"``.
-    :rtype: str or None
     """
 
     opened_here = False
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -66,7 +66,7 @@ def import_od(
     """Parse an EDS, DCF, or EPF file.
 
     :param source:
-        Path to object dictionary file, a file like object, or an EPF XML tree.
+        The path to object dictionary file, a file like object, or an EPF XML tree.
     :param node_id:
         For EDS and DCF files, the node ID to use.
         For other formats, this parameter is ignored.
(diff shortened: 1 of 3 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -61,10 +66,10 @@ def import_od(
     """Parse an EDS, DCF, or EPF file.
 
     :param source:
-        Path to object dictionary file or a file like object or an EPF XML tree.
-
-    :return:
-        An Object Dictionary instance.
+        Path to object dictionary file, a file like object, or an EPF XML tree.
+    :param node_id:
+        For EDS and DCF files, the node ID to use.
```
erlend-aasland, line 71: I this param docstring should be expanded; AFAICS, the node ID is only used for calculating indexes[^1]. I'm not sure if it is a bug that it's not used for the resulting ODs `.node_id` member.

[^1]: I might be using the wrong terminology here.

erlend-aasland:
Docstring changes:

- Consistent punctuation and capitalisation.
- Mark up literals using monospaced font.
- Remove unneeded return type annotation.
- Use typing to autogenerate param list and return info.

canopen/objectdictionary/__init__.py
```diff
@@ -15,17 +15,22 @@
...
+def export_od(
+    od: ObjectDictionary,
+    dest: Union[str, TextIO, None] = None,
+    doc_type: Optional[str] = None
+) -> Union[str, None]:
+    """Export an object dictionary.
 
     :param od:
-        :class: ObjectDictionary object to be exported
+        The :class:`!ObjectDictionary` object to be exported.
```
erlend-aasland, line 26: Perhaps it is better to spell it out; the param list will include the link to ObjectDictionary (and there is no need to link to it twice in one line.

```suggestion
        The object dictionary to be exported.
```

canopen/objectdictionary/__init__.py
```diff
@@ -15,17 +15,22 @@
...
     :param dest:
-        export destination. filename, or file-like object or None.
-        if None, the document is returned as string
-    :param doc_type: type of document to export.
-       If a filename is given for dest, this default to the file extension.
-       Otherwise, this defaults to "eds"
+        The export destination as a filename, a file-like object, or None.
+        If ``None``, the document is returned as string.
```
erlend-aasland, line 29: Further testing reveals that this is not true; we're delegating to the stdlib `configparser` library here, which writes to `sys.stdout` (which _hopefully_ is the actual stdout, and not a monkey-patched thingy). Moreover, this is not tested; I'll add tests in a separate PR.

```suggestion
        If ``None``, the document is printed to :data:`sys.stdout`.
```

Alternatively:

```suggestion
        If ``None``, the document is written to :data:`sys.stdout`.
```

erlend-aasland: I went for the former: see commit bc4f739997d864580ac378dce31afb11418425f2.

canopen/objectdictionary/__init__.py
erlend-aasland: IMO, it is better to spell it out; if the reviewer disagrees, one can simply revert the offending commit.

canopen/objectdictionary/__init__.py
```diff
@@ -15,18 +15,22 @@
...
-    :param doc_type: type of document to export.
-       If a filename is given for dest, this default to the file extension.
-       Otherwise, this defaults to "eds"
-    :rtype: str or None
+        The export destination as a filename, a file-like object, or None.
+        If ``None``, the document is written to :data:`sys.stdout`.
+    :param doc_type:
+       The type of document to export.
+       If *dest* is a filename, *doc_type* defaults to its extension.
+       Otherwise, the document type defaults to ``"eds"``.
```
erlend-aasland, line 33: This is not true if *dest* is ``None``; in that case, *doc_type* MUST be supplied.

acolomb: Correct.  Feel free to set it straight in the comment.

erlend-aasland: Addressed in a79102b

canopen/objectdictionary/__init__.py
```diff
@@ -15,18 +15,25 @@
...
-    :rtype: str or None
+        The export destination as a filename, a file-like object, or ``None``.
+        If ``None``, the document is written to :data:`sys.stdout`.
+    :param doc_type:
+       The type of document to export.
+       If *dest* is a file-like object or ``None``,
+       *doc_type* must be explicitly provided.
+       If *dest* is a filename and its extension is ``".eds"`` or ``".dcf"``,
+       *doc_type* defaults to that extension (the preceeding dot excluded);
+       else, it defaults to ``"eds"``.
```
acolomb, line 36: Just a nit-pick, I find it awkward to have quotes inside quotes here.  Please remove the inner ones.
```suggestion
       If *dest* is a filename and its extension is ``.eds`` or ``.dcf``,
       *doc_type* defaults to that extension (the preceeding dot excluded);
       else, it defaults to ``eds``.
```

acolomb (changes requested):

canopen/objectdictionary/__init__.py
acolomb: The correct terminology would be "interpolation" of values containing a `$NODEID` placeholder: https://github.com/christiansandberg/canopen/blob/5db59130423948e1e8034a67983ca00afd0269d8/canopen/objectdictionary/eds.py#L225

The `ObjectDictionary` instance is generic if loaded from an EDS file, but node-specific if loaded from a DCF (which has a `DeviceComissioning` section).  In the latter case, the `node_id` is in fact stored to the instance.  Passing the parameter is just a way to override the node ID supplied in the DCF file, and to adjust the interpolated COB-ID values in DCF- or EDS-based object dictionaries.

canopen/objectdictionary/__init__.py
erlend-aasland: I'll make the change, though I will point out that it is normal practice to mark up string literals as they are written. Without quotes, there is no visual difference in the docs between identifiers and string literals. See the CPython docs as an example of this convention:

https://docs.python.org/3/library/string.html

acolomb: OK, I didn't understand this as meaning a string literal.  Sure, a variable will hold `".eds"`, but the file's extension is still only `.eds`.  Most of all, I think the code / docstring should be readable (in unrendered form) and the code-style double backquotes are sufficient to see the intention.  It's probably a bit different from CPython docs, where the syntax is more in focus and the docs are hardly ever consumed without rendering.

erlend-aasland: Fair enough!

erlend-aasland:
On hold until #476 has landed.

canopen/objectdictionary/__init__.py
erlend-aasland: > In the latter case, the `node_id` is in fact stored to the instance.

Sure, but that's the `DeviceComissioning` _Node ID_ field; not the `node_id` parameter of `import_od`.

AFAICS, the `node_id` parameter is solely used for "interpolation" via the `build_variable()` helper.

acolomb: If given, the `node_id` parameter overrides the one from the DCF: https://github.com/christiansandberg/canopen/blob/503918c9bd9ff0499a604c7805e7627ca38fb332/canopen/objectdictionary/eds.py#L90

canopen/objectdictionary/__init__.py
```diff
@@ -15,18 +15,27 @@
...
-    :rtype: str or None
+        The export destination as a filename, a file-like object, or ``None``.
+        If ``None``, the document is written to :data:`sys.stdout`.
+    :param doc_type:
+       The type of document to export.
+       If *dest* is a file-like object or ``None``,
+       *doc_type* must be explicitly provided.
+       If *dest* is a filename and its extension is ``.eds`` or ``.dcf``,
+       *doc_type* defaults to that extension (the preceeding dot excluded);
+       else, it defaults to ``"eds"``.
```
acolomb, line 36: Here as well.
```suggestion
       else, it defaults to ``eds``.
```

canopen/objectdictionary/__init__.py
erlend-aasland: Sure; but `node_id` is only passed as an argument to subsequent `build_variable()` calls; it does not end up in `od.node_id`. I understood your remarks as if the latter would be true.

acolomb: Oh, you're right.  Sorry for my confusion.  Then yes, that's a bug.  When reading a DCF file and overriding the `$NODEID` interpolation value, then we shouldn't keep storing the previous ID read from the file.

The only place where the attribute is used seems to be in `BaseNode.__init__()` if no ID was passed in there as a parameter.  Seems we have too many ways of instantiating stuff...  And of course when exporting a DCF again.  In both cases, using the override from the `import_od()` argument is the sensible thing to do.

Should we fix this here or would you make a separate PR?

acolomb (approved):

erlend-aasland:
Thanks for the review and your insights.

canopen/objectdictionary/__init__.py
erlend-aasland: For the record, the issue/PR were created:

- #478 
- #484

PR #476, opened 2024-07-02
Raise ValueError for unknown formats in the import/export OD APIs

### LINKED ISSUE #475: Consider using ObjectDictionaryError iso. NotImplementedError in export/import OD APIs
Discovered while reviewing #469. (See https://github.com/christiansandberg/canopen/pull/469#discussion_r1661613953.)

In most cases, `ObjectDictionaryError` is used for the `export_od` and `import_od` APIs, but for unsupported file formats, a `NotImplementedError` is raised. This is unconventional; `NotImplementedError` is not a good fit for such an error case, as it was designed for other purposes (see the CPython docs).

The `NotImplementedError` in `export_od` is fairly new (introduced with #469), and has not been a part of a formal release; it can be changed without backwards compatibility concerns. The `raise` in `import_od` has been there longer, _but_ these are (for now) undocumented, though public, APIs. One can argue that since they are undocumented, it may be ok to change `import_od` as well.

Consider switching to `ObjectDictionaryError`.

erlend-aasland:
Fixes #475

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -50,7 +50,11 @@ def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str]
             from canopen.objectdictionary import eds
             return eds.export_dcf(od, dest)
         else:
-            raise ValueError(f"No support for the {doc_type!r} format")
+            allowed = ", ".join(doctypes)
+            raise ValueError(
+                f"Cannot export to the {doc_type!r} format; "
+                f"supported formats: {allowed}"
+            )
     finally:
         # If dest is opened in this fn, it should be closed
         if opened_here:
@@ -89,7 +93,11 @@ def import_od(
         return epf.import_epf(source)
     else:
         doc_type = suffix[1:]
-        raise ValueError(f"No support for the {doc_type!r} format")
+        allowed = ", ".join(["eds", "dcf", "epf"])
+        raise ValueError(
+            f"Cannot import from the {doc_type!r} format; "
+            f"supported formats: {allowed}"
+        )
 
 
 class ObjectDictionary(MutableMapping):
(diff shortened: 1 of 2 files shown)


erlend-aasland:
I'm fine with `ValueError`; it is a good alternative to `ObjectDictionaryError`.

canopen/objectdictionary/__init__.py
```diff
@@ -50,7 +50,7 @@ def export_od(od, dest: Union[str, TextIO, None] = None, doc_type: Optional[str]
             from canopen.objectdictionary import eds
             return eds.export_dcf(od, dest)
         else:
-            raise NotImplementedError(f"No support for the {doc_type!r} format")
+            raise ValueError(f"No support for the {doc_type!r} format")
```
erlend-aasland, line 53: Consider also providing a more helpful error message.

```suggestion
            allowed = ", ".join(doctypes)
            raise ValueError(f"Cannot export to the {doc_type!r} format; allowed formats: {allowed}")
```

If you agree, let's do a similar change for the import API.

erlend-aasland: Alternatively:

```suggestion
            allowed = ", ".join(doctypes)
            raise ValueError(f"Cannot export to the {doc_type!r} format; supported formats: {allowed}")
```

acolomb: Sure, go ahead.  Exception messages should be as helpful as possible.

test/test_eds.py
```diff
@@ -1,6 +1,7 @@
 import os
 import unittest
 import canopen
+from canopen.objectdictionary import ObjectDictionaryError
```
acolomb, line 4: No longer needed.

erlend-aasland: Good catch; thanks!

canopen/objectdictionary/__init__.py
erlend-aasland: Addressed in 410b683de521255d6f79992285cf2f7dfa8fc831. Marking as resolved.

test/test_eds.py
erlend-aasland: Addressed in 9490fbf06dcf92632790d6d4b9f966eda9105d4e; marking as resolved.

erlend-aasland:
Thanks a lot for the review and the helpful discussion; highly appreciated.

PR #484, opened 2024-07-02
Save the passed node ID to the resulting OD in `import_od()`

### LINKED ISSUE #478: The `node_id` parameter of `import_od` is not saved to the resulting OD object
Per now, it is only used for "interpolation".

Discovered while working on #473.

See https://github.com/christiansandberg/canopen/pull/473#discussion_r1660015548

erlend-aasland:
Fixes #478

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -87,9 +87,8 @@ def import_eds(source, node_id):
     if eds.has_section("DeviceComissioning"):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
 
-        default = None if node_id is None else {"NodeID": node_id}
-        id_ = eds.get("DeviceComissioning", "NodeID", vars=default)
-        od.node_id = int(id_, base=0)
+        id_ = int(eds.get("DeviceComissioning", "NodeID"), base=0)
+        od.node_id = id_ if node_id is None else node_id
 
     for section in eds.sections():
         # Match dummy definitions
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -88,7 +88,8 @@ def import_eds(source, node_id):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
 
         if node_id is None:
-            node_id = int(eds.get("DeviceComissioning", "NodeID"), base=0)
+            if val := eds.get("DeviceComissioning", "NodeID", fallback=None):
+                node_id = int(val, base=0)
         od.node_id = node_id
 
     for section in eds.sections():
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -86,8 +86,10 @@ def import_eds(source, node_id):
 
     if eds.has_section("DeviceComissioning"):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
-        od.node_id = int(eds.get("DeviceComissioning", "NodeID"), 0)
-        node_id = node_id or od.node_id
+
+        default = None if node_id is None else {"NodeID": node_id}
+        id_ = eds.get("DeviceComissioning", "NodeID", vars=default)
+        od.node_id = int(id_, base=0)
 
     for section in eds.sections():
         # Match dummy definitions
@@ -119,7 +121,7 @@ def import_eds(source, node_id):
                 storage_location = None
 
             if object_type in (VAR, DOMAIN):
-                var = build_variable(eds, section, node_id, index)
+                var = build_variable(eds, section, od.node_id, index)
                 od.add_object(var)
             elif object_type == ARR and eds.has_option(section, "CompactSubObj"):
                 arr = objectdictionary.ODArray(name, index)
@@ -127,7 +129,7 @@ def import_eds(source, node_id):
                     "Number of entries", index, 0)
                 last_subindex.data_type = datatypes.UNSIGNED8
                 arr.add_member(last_subindex)
-                arr.add_member(build_variable(eds, section, node_id, index, 1))
+                arr.add_member(build_variable(eds, section, od.node_id, index, 1))
                 arr.storage_location = storage_location
                 od.add_object(arr)
             elif object_type == ARR:
@@ -149,7 +151,7 @@ def import_eds(source, node_id):
             entry = od[index]
             if isinstance(entry, (objectdictionary.ODRecord,
                                   objectdictionary.ODArray)):
-                var = build_variable(eds, section, node_id, index, subindex)
+                var = build_variable(eds, section, od.node_id, index, subindex)
                 entry.add_member(var)
 
         # Match [index]Name
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -87,8 +87,9 @@ def import_eds(source, node_id):
     if eds.has_section("DeviceComissioning"):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
 
-        id_ = int(eds.get("DeviceComissioning", "NodeID"), base=0)
-        od.node_id = id_ if node_id is None else node_id
+        if node_id is None:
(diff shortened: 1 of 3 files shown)


erlend-aasland:
Note: I deliberately changed the value of `NodeID` in `sample.eds`, since `base=0` is used when converting from `str` to `int`, just to make sure we test that aspect as well.

acolomb (changes requested):
Wouldn't it make sense to stick with the `node_id` parameter in places where it's used?  Just *in addition*, override the `od.node_id` attribute if the argument was given.  I think it's safer not having to remember the `od.` prefix in any later usage within the function.

erlend-aasland:
> Wouldn't it make sense to stick with the `node_id` parameter in places where it's used? Just _in addition_, override the `od.node_id` attribute if the argument was given. I think it's safer not having to remember the `od.` prefix in any later usage within the function.

That would absolutely reduce the diff, but I generally don't like shadowing parameters/variables, which is why I went for the `od` member. There is of course the risk that new code would be written using `node_id`, and not the correct `od.node_id`, _but_ such code would probably be copy-and-pasted from the existing snippets that correctly use `od` member anyway.

I don't have a strong opinion about this; I'll change it as you wish.

acolomb:
Yea I prefer overriding the `node_id` parameter.  It's like the basic pattern for optional parameters, where `None` gets replaced by a default value (from the file in this case).  And for the acutal change, we don't "shadow" the parameter, but rather just need to make sure it is respected when copying to the OD instance.  And it's less cognitive load when reading without the `od.` prefix.

acolomb (approved):

erlend-aasland:
Addressed in 8a22e7ed329afbda46ac629c7c6974223477c0c2. _But_, there's a case we're not testing: if the passed `node_id` argument is `None` _and_ there is not `Device Comissioning.Node ID` entry, we'll end up with a node ID of zero, which is mean for broadcast only AFAIK. Do we want to emit a warning, or perhaps raise an exception in that case? Or perhaps another default?

acolomb:
An `ObjectDictionary` instance does not need a `node_id` at all, so keeping it at `None` is fine IMHO.  Think of it as a blueprint for a `Node`, which is where the node ID really gets relevant.  It's absolutely expected to share the same OD instance between several nodes, so the OD's node ID is ambiguous anyway.  The whole saving within the OD instance serves only a single purpose: Loading DCF files (for a single, specific, pre-parameterized node) with the same code as for EDS.  That's really the only case when `ObjectDictionary.node_id` makes sense.

erlend-aasland:
For the case described in https://github.com/christiansandberg/canopen/pull/484#issuecomment-2207281646, of course the node ID ends up being `None`, not 0.

erlend-aasland:
Thanks for the insight!

erlend-aasland:
I still would consider adding this patch to the PR before merging:

<details>

```diff
diff --git a/test/test_eds.py b/test/test_eds.py
index f46cc0e..00d5013 100644
--- a/test/test_eds.py
+++ b/test/test_eds.py
@@ -4,7 +4,8 @@ import canopen
 from canopen.objectdictionary.eds import _signed_int_from_hex
 from canopen.utils import pretty_index
 
-EDS_PATH = os.path.join(os.path.dirname(__file__), 'sample.eds')
+SAMPLE_EDS = os.path.join(os.path.dirname(__file__), 'sample.eds')
+DATATYPES_EDS = os.path.join(os.path.dirname(__file__), 'datatypes.eds')
 
 
 class TestEDS(unittest.TestCase):
@@ -48,7 +49,7 @@ class TestEDS(unittest.TestCase):
     }
 
     def setUp(self):
-        self.od = canopen.import_od(EDS_PATH, 2)
+        self.od = canopen.import_od(SAMPLE_EDS, 2)
 
     def test_load_nonexisting_file(self):
         with self.assertRaises(IOError):
@@ -59,17 +60,22 @@ class TestEDS(unittest.TestCase):
             canopen.import_od(__file__)
 
     def test_load_file_object(self):
-        with open(EDS_PATH) as fp:
+        with open(SAMPLE_EDS) as fp:
             od = canopen.import_od(fp)
         self.assertTrue(len(od) > 0)
 
     def test_load_implicit_nodeid(self):
         # sample.eds has a DeviceComissioning section with NodeID set to 0x10.
-        od = canopen.import_od(EDS_PATH)
+        od = canopen.import_od(SAMPLE_EDS)
         self.assertEqual(od.node_id, 16)
 
+    def test_load_implicit_nodeid_no_devicecomissioning_section(self):
+        # datatypes.eds has no DeviceComissioning section.
+        od = canopen.import_od(DATATYPES_EDS)
+        self.assertEqual(od.node_id, None)
+
     def test_load_explicit_nodeid(self):
-        od = canopen.import_od(EDS_PATH, node_id=3)
+        od = canopen.import_od(SAMPLE_EDS, node_id=3)
         self.assertEqual(od.node_id, 3)
 
     def test_variable(self):
```

</details>

canopen/objectdictionary/eds.py
```diff
@@ -86,8 +86,10 @@ def import_eds(source, node_id):
 
     if eds.has_section("DeviceComissioning"):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
-        od.node_id = int(eds.get("DeviceComissioning", "NodeID"), 0)
-        node_id = node_id or od.node_id
+
+        if node_id is None:
+            node_id = int(eds.get("DeviceComissioning", "NodeID"), base=0)
```
acolomb, line 91: I wonder what happens when the option is not set within the section.  https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.get is unclear on whether an exception is raised.

erlend-aasland: A `DeviceComissioning` section without a `NodeID` option generates a `NoOptionError` exception.

acolomb: It's also unclear in the CiA306 spec whether the options are required, but I'd assume they are optional.  Thus better to explicitly ignore `KeyError` or whatever exception occurs here.  Of course, a `fallback=None`  argument with conditional assignment to the OD would work without caring about exceptions.

erlend-aasland: Sure, but `int(None)` gives you a `TypeError`.

acolomb:
> I still would consider adding this patch to the PR before merging:

Sure, go ahead.  The separate sample EDS files were introduced recently for the PDO stuff IIRC.

canopen/objectdictionary/eds.py
acolomb: That's what I meant with "conditional assignment".  If the option is not present, don't try converting at all.  Whether you detect that by catching the exception, or the `None` fallback value, is up to you.

erlend-aasland: Addressed in ac79a35c792fc60bb0fabad56e60f4d659747a1b.

canopen/objectdictionary/eds.py
```diff
@@ -86,8 +86,11 @@ def import_eds(source, node_id):
 
     if eds.has_section("DeviceComissioning"):
         od.bitrate = int(eds.get("DeviceComissioning", "Baudrate")) * 1000
-        od.node_id = int(eds.get("DeviceComissioning", "NodeID"), 0)
-        node_id = node_id or od.node_id
+
+        if node_id is None:
+            if val := eds.get("DeviceComissioning", "NodeID", fallback=None):
+                node_id = int(val, base=0)
```
acolomb, line 92: Great! Couldn't have written it more beautifully.  I still need to get used to the walrus operator, though I know and love it from Go (in slightly different form).

Do the same for bitrate?

erlend-aasland: > Do the same for bitrate?

Sure; preferably in a separate PR.

Love Go, BTW! Too bad I don't get to use it often. OT: there are some interesting building concurrency blocks in the pipeline for future Python releases[^1][^2].

[^1]: https://peps.python.org/pep-0734/
[^2]: https://pyfound.blogspot.com/2024/06/python-language-summit-2024-lightning-talks.html (see Yuri's lightning talk; super interesting!)

acolomb (approved):

erlend-aasland:
Thanks for the review!

Issue #487, opened 2024-07-02
Consider bumping the `python-can` minimum required version to 4.0


erlend-aasland:
I tried running the test-suite with `python-can` 3.7, but it failed. Starting with `python-can` 4.0, it succeeds. Consider bumping the version requirement in pyproject.toml.

sveinse:
Is there a minimum python requirement in `python-can` 4.0? If there is, we might need to follow suit on minimum python version. Alternatively we must make canopen work with older python-can.

erlend-aasland:
`python-can` current `main`[^1] requires Python 3.8 or newer. That's the same Python requirement as `canopen`.

`python-can` v4.2.0 requires Python 3.7. Older versions of `python-can` has no specific requirement encoded in the `pyproject.toml` file.

IOW, the Python requirement is not a blocker.

[^1]: HEAD at `174d5600aaf3377bd44f1eab9cb53ae066d606ca`

erlend-aasland:
Seems like `python-can` v4.3.0 started requiring Python 3.8. Since that harmonises with the `canopen` Python requirement, it may make sense to bump the `python-can` requirement to 4.3.0.

acolomb:
> I tried running the test-suite with `python-can` 3.7, but it failed. Starting with `python-can` 4.0, it succeeds. Consider bumping the version requirement in pyproject.toml.

Where did you get python-can version 3.7? https://pypi.org/project/python-can/#history does not show such a version.  I'm interested in the actual failure mode, maybe we can then work around it.  Do you still have logs from your trial?

erlend-aasland:
> Where did you get python-can version 3.7? https://pypi.org/project/python-can/#history does not show such a version. I'm interested in the actual failure mode, maybe we can then work around it. Do you still have logs from your trial?

That should've been version 3.0, not 3.7; sorry 'bout that.

```sh
$ python3 -m pip install python-can==3
[...]
Successfully installed python-can-3.0.0 typing-3.7.4.3
```

Nonetheless, the test failure is:

```sh
$ python -m unittest discover test
[...]
======================================================================
FAIL: test_send_perodic (test_network.TestNetwork.test_send_perodic)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/erlend.aasland/src/canopen.git/test/test_network.py", line 65, in test_send_perodic
    self.assertTrue(9 <= bus.queue.qsize() <= 13)
[...]

PR #489, opened 2024-07-03
Adapt to renamed "bustype" parameter on can.Bus

acolomb:
The bustype keyword has long been replaced by "interface" and the former is deprecated in python-can.  The Network class passes it through without caring for the name, but documentation and tests referred to the older name regularly.

Based on discussion in #487.


erlend-aasland (approved):
LGTM. Confirmed there are no other occurrences of *bustype* remaining:

```sh
$ git grep bustype
$
```

PR #490, opened 2024-07-03
Restrict python-can to major versions 3 and 4

### LINKED ISSUE #487: Consider bumping the `python-can` minimum required version to 4.0

erlend-aasland:
Resolves #487

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 dependencies = [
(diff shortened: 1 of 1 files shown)


acolomb:
Do we still need this after #489?  I don't think there is an actual incompatibility anymore?

erlend-aasland:
Correct; this is not needed anymore.

PR #492, opened 2024-07-03
Gracefully handle missing Baudrate setting in DCF

acolomb:
The CiA306 spec is not quite clear on which fields are mandatory within the DeviceComissioning section.  Thus it should be assumed that the two options handled here could be missing.  Instead of raising an exception, ignore the absence of the Baudrate setting.


erlend-aasland (approved):

erlend-aasland:
I added tests for this:

- #495

PR #493, opened 2024-07-03
Add back PdoBase.export() dependency (canmatrix)

### LINKED ISSUE #488: `PdoBase.export()` is broken bco. broken dependencies
`PdoBase.export()` depends on the third-party module `canmatrix`, which was removed from the `canopen` requirements with commit c46228f9cf2d2661166d68e7175f3e8b99064194, three years ago. This implies it has been de facto broken since that commit, unless the user accidentally had a compatible `canmatrix` library installed.

There's a couple of ways to mitigate this:

1. add back the `canmatrix` dependency, fix `.export()` (if it needs fixing), and add unit tests in order to prevent future regressions like this
2. acknowledge that the API is de facto broken and purge it from `canopen`

FWIW, my personal preference would be 2).

Discovered while working on increasing PDO coverage to 100%.

erlend-aasland:
The canmatrix optional dependency was removed on Oct 10, 2021 with commit
c46228f. It is now added back as an optional dependency, using the same
name as previously: db_export.

To install the dependency:

$ python3 -m pip install 'canopen[db_export]'

Resolves #488

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -87,7 +87,7 @@ class PdoBase(Mapping):
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
         :raises ImportError:
-            When the ``db_export`` is not installed.
+            When the ``db_export`` feature is not installed.
 
         :return: The CanMatrix object created
         :rtype: canmatrix.canmatrix.CanMatrix
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -87,7 +87,7 @@ class PdoBase(Mapping):
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
         :raises ImportError:
-            When the ``db_feature`` is not installed.
+            When the ``db_export`` is not installed.
 
         :return: The CanMatrix object created
         :rtype: canmatrix.canmatrix.CanMatrix
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-import functools
 import threading
 import math
 from typing import Callable, Dict, Iterator, List, Optional, Union, TYPE_CHECKING
@@ -10,27 +9,12 @@ import binascii
 from canopen.sdo import SdoAbortedError
 from canopen import objectdictionary
 from canopen import variable
-try:
-    import canmatrix
-except ImportError:
-    canmatrix = None
 
 if TYPE_CHECKING:
     from canopen.network import Network
     from canopen import LocalNode, RemoteNode
     from canopen.pdo import RPDO, TPDO
     from canopen.sdo import SdoRecord
-
-
-def _disable_if(condition):
-    """Conditionally disable given function/method."""
-    def deco(func):
-        @functools.wraps(func)
-        def wrapper(*args, **kwds):
-            if not condition:
-                return func(*args, **kwds)
-        return wrapper
-    return deco
 
 PDO_NOT_VALID = 1 << 31
 RTR_NOT_ALLOWED = 1 << 30
@@ -91,7 +75,6 @@ class PdoBase(Mapping):
         for pdo_map in self.map.values():
             pdo_map.subscribe()
 
-    @_disable_if(canmatrix is None)
     def export(self, filename):
         """Export current configuration to a database file.
 
@@ -103,6 +86,8 @@ class PdoBase(Mapping):
 
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
+        :raises ImportError:
+            When the ``db_feature`` is not installed.
 
         :return: The CanMatrix object created
         :rtype: canmatrix.canmatrix.CanMatrix
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import functools
 import threading
 import math
 from typing import Callable, Dict, Iterator, List, Optional, Union, TYPE_CHECKING
@@ -9,12 +10,27 @@ import binascii
 from canopen.sdo import SdoAbortedError
 from canopen import objectdictionary
 from canopen import variable
+try:
+    import canmatrix
+except ImportError:
+    canmatrix = None
 
 if TYPE_CHECKING:
     from canopen.network import Network
     from canopen import LocalNode, RemoteNode
     from canopen.pdo import RPDO, TPDO
     from canopen.sdo import SdoRecord
+
+
+def _disable_if(condition):
+    """Conditionally disable given function/method."""
+    def deco(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwds):
+            if not condition:
+                return func(*args, **kwds)
+        return wrapper
+    return deco
 
 PDO_NOT_VALID = 1 << 31
 RTR_NOT_ALLOWED = 1 << 30
@@ -75,8 +91,15 @@ class PdoBase(Mapping):
         for pdo_map in self.map.values():
             pdo_map.subscribe()
 
+    @_disable_if(canmatrix is None)
     def export(self, filename):
         """Export current configuration to a database file.
+
+        .. note::
+           This API depends on the :mod:`!canmatrix` optional dependency,
+           which can be installed by requesting the ``db_export`` feature::
+
+              python3 -m pip install 'canopen[db_export]'
 
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -79,21 +79,23 @@ class PdoBase(Mapping):
         """Export current configuration to a database file.
 
         .. note::
-           This API depends on the :mod:`!canmatrix` optional dependency,
-           which can be installed by requesting the ``db_export`` feature::
+           This API requires the ``db_export`` feature to be installed::
 
               python3 -m pip install 'canopen[db_export]'
 
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
-        :raises ImportError:
-            When the ``db_export`` feature is not installed.
+        :raises NotImplementedError:
+            When the ``canopen[db_export]`` feature is not installed.
 
         :return: The CanMatrix object created
         :rtype: canmatrix.canmatrix.CanMatrix
         """
-        from canmatrix import canmatrix
-        from canmatrix import formats
+        try:
+            from canmatrix import canmatrix
+            from canmatrix import formats
+        except ImportError:
+            raise NotImplementedError("This feature requires the 'canopen[db_export]' feature")
 
         db = canmatrix.CanMatrix()
         for pdo_map in self.map.values():
--- a/test/test_pdo.py
+++ b/test/test_pdo.py
@@ -1,41 +1,51 @@
 import os.path
 import unittest
+
 import canopen
+try:
+    import canmatrix
+except ImportError:
+    canmatrix = None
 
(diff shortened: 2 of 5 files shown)


canopen/pdo/base.py
```diff
@@ -9,13 +10,28 @@
...
+
+def _disable_if(condition):
+    """Conditionally disable given function/method."""
+    def deco(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwds):
+            if not condition:
+                return func(*args, **kwds)
+        return wrapper
+    return deco
```
erlend-aasland, line 33: Adding a decorator for this may seem overly verbose, but conditionally adding the method in other ways may mess up the documentation. I believe this is the best approach, but I'm open for other suggestions.

erlend-aasland:
FTR, I reused the same feature specifier as before: `db_export`.

test/test_pdo.py
```diff
@@ -55,10 +65,23 @@ def test_bit_mapping(self):
...
+        self.node.rpdo.save()
+
+    @unittest.skipIf(canmatrix is None, "Requires canmatrix dependency")
+    def test_pdo_export(self):
+        import tempfile
+
+        for pdo in "tpdo", "rpdo":
+            with tempfile.NamedTemporaryFile(suffix=".csv") as tmp:
+                fn = tmp.name
+                with self.subTest(pdo=pdo, filename=fn):
```
erlend-aasland, line 79: Perhaps a little more easy on the eyes when debugging a failed test:

```suggestion
                with self.subTest(filename=fn, pdo=pdo):
```

canopen/pdo/base.py
acolomb: Do we really need to disable (remove) the method?  My simple approach would be to just throw an exception when the import fails.  Currently that's an `ImportError`, but it could be wrapped with a better exception message explaining the optional dependency.

erlend-aasland: Well, it's a convention to leave the method undefined if it's not supported, which is why I went for that, but other solutions are fine as long as they are documented.

acolomb: I'd favor a less intrusive solution. In fact, just letting the ImportError propagate, but documenting that it is possibly raised, would be sufficient.

erlend-aasland: Sure, I'm fine with that.

erlend-aasland: Addressed in 0b2eae97e1e259a802cb18498c525b85f268c486. Marking as resolved. PTAL.

erlend-aasland: It _may_ be slightly harder to apply #358 to this API with the simpler approach. If the API is conditionally approached, the return annotation is pretty straight forward. Now, I'm not quite sure how to handle it (and I'm far from a typing expert). Perhaps a `-> "canmatrix.CanMatrix"` will work; I'm not sure about the best approach.

erlend-aasland: cc. @sveinse, who probably knows more about typing than me.

sveinse: Using `-> "canmatrix.CanMatrix"` works fine (when `from __future__ import annotations` is used: `-> canmatrix.CanMatrix`) works find even if canmatrix is not available. It does not interfer with the run-time operations if it is not present.

However, the type checker do not like this construct where a module becomes a variable. The better fix is simply `pass` here.
```py
try:
    import canmatrix
except ImportError:
    canmatrix = None
```

In order to avoid errors by the type checker, canmatrix must be installed for it to check its types. I think it's fine to get the reported errors when it is not present.

erlend-aasland: Thanks for chiming in. IIUC, the approach used as the PR stands now should not be too much of a hassle when typing is eventually added.

sveinse: My 2 cents on missing imports: Either
1. use conditional import in the top of the file and simply let it fail when calling the methods that require canmatrix. But help the user to understand that the raise exception is due to missing canmatrix
2. stick with importing inside the method. I know it's not pep8-like/pythonic, but the resulting `ImportError()` is very unambigous.

erlend-aasland: Yeah. As I already wrote, I prefer something similar to my first proposal. Regarding the exception type, I don't have a strong opinion. I defer to the library maintainers :)

erlend-aasland: > However, the type checker do not like this construct where a module becomes a variable. The better fix is simply pass here.

... or simply to set a (boolean) guard to be used, for example, in a decorator variant.

sveinse: Use of any other variable as sentinel is fine. Just not the same name as the module (`canmatrix`).

canopen/pdo/base.py
```diff
@@ -78,8 +78,16 @@ def subscribe(self):
...
+        .. note::
+           This API depends on the :mod:`!canmatrix` optional dependency,
+           which can be installed by requesting the ``db_export`` feature::
+
+              python3 -m pip install 'canopen[db_export]'
+
         :param str filename:
             Filename to save to (e.g. DBC, DBF, ARXML, KCD etc)
+        :raises ImportError:
+            When the ``db_export`` feature is not installed.
```
acolomb, line 90: Moving the imports to the top is certainly a good idea.  I suggest to try the import at the top, but pass (with a comment) when it fails.  Down here in the method, let's be pythonic and ask for forgiveness rather than permission:  Just `try` the first line and if the import was not successful, catch the `NameError` and re-raise an appropriate `RuntimeError` or further specialized exception explaining what needs to be done.

I'd like to avoid any special constructs such as sentinel variables or dummy values for the `canmatrix` name.

test/test_pdo.py
```diff
@@ -1,41 +1,51 @@
 import os.path
 import unittest
+
 import canopen
+try:
+    import canmatrix
+except ImportError:
+    canmatrix = None
```
acolomb, line 8: Here again: Let's just `pass` on the import error and skip the test when access to any needed names in `canmatrix` gives a `NameError`.

test/test_pdo.py
```diff
@@ -55,10 +65,23 @@ def test_bit_mapping(self):
...
 
-    def test_save_pdo(self):
-        node = canopen.Node(1, EDS_PATH)
-        node.tpdo.save()
-        node.rpdo.save()
+    def test_pdo_save(self):
+        self.node.tpdo.save()
+        self.node.rpdo.save()
+
+    @unittest.skipIf(canmatrix is None, "Requires canmatrix dependency")
```
acolomb, line 72: Can we do a simple runtime check and catch the exception instead of this decorator using a sentinel value (`None`)?  Something like (from the top of my head, no clue about the unittest skipping function):

    try:
        _ = canmatrix.canmatrix
    except NameError:
        unittest.skip()

If a simple boolean condition is the only way to cleanly skip the test, let's do that in the import exception block above, but rather not with the `canmatrix` name.

acolomb (changes requested):

test/test_pdo.py
erlend-aasland: Sure!

canopen/pdo/base.py
erlend-aasland: Sure, that's a fine approach!

canopen/pdo/base.py
acolomb: Please do #2.  If you think a `RuntimeError` is not appropriate, we can manually raise an `ImportError`, which is very concise.  Who says an `ImportError` can only be raised at the time of importing? ;-)

But I'm not supposed to adding a new exception type either, if you think those ideas are too hacky.

erlend-aasland: > Who says an ImportError can only be raised at the time of importing? ;-)

https://peps.python.org/pep-0690/ (too bad it didn't happen!) 😁 

None of the proposed ideas are too hacky; all is fine. It's just a matter of deciding on an approach and documenting it for the users convenience :)

erlend-aasland: Regarding the exception type; the `TypeError` docs says this:

> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be. If an object is meant to support a given operation but has not yet provided an implementation, [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) is the proper exception to raise.

I guess we could use both `TypeError` and `NotImplementedError`, even though it might be a little bit unconventional. Perhaps `NotImplementedError` is the better choice.

acolomb: Then let's go for `NotImplementedError`.  Good idea!

erlend-aasland: Let's continue the discussion in your more recent remarks. Marking this conversation as resolved.

canopen/pdo/base.py
erlend-aasland: I choose NotImplementedError. See 013ff93cacb4261dd9b5c95dd7bb139056abab00. Let me know if that's fine.

acolomb: Thought we wanted to use import at the top though?  I meant wrapping the actual `CanMatrix` construction in the `try... except`.

erlend-aasland: > Thought we wanted to use import at the top though?

Why?

erlend-aasland: To expand on my "why?" comment: Doing an `.export()` with `canmatrix` installed works; doing it without `canmatrix` installed raises `NotImplementedError` with an explanatory error message. What's the missing behaviour?

acolomb: My understanding was that local imports are frowned upon.  Should we even introduce additional functionality based on `canmatrix`, then doing it once at the top seems the better approach.  So why not start with it right away?

Basically, have all dependencies indicated at the top, even the optional one (with an appropriate comment why `pass`ing on the failed import is OK).  In the method, as the first thing to do, check explicitly if the required optional dependency is there (by catching `NameError`), so in that place it's also easily spotted that the option is a prerequisite.

erlend-aasland: > My understanding was that local imports are frowned upon.

IMO, that depends on the API in question, and the kind of code base. For tests, I see it all the time. For optional APIs like this, I think it is a valid use case.

> Should we even introduce additional functionality based on `canmatrix`, then doing it once at the top seems the better approach. So why not start with it right away?

I'm not convinced by that argument, but I'll abide :)

erlend-aasland: FTR; I won't be back on my computer until earliest tomorrow, so don't expect any immediate updates. In the mean time, consider if you really want to go this route; I'm not sure the added complexity is really worth it. The current solution is compact and accurate. I'm not sure the proposed alternative is a better solution.

acolomb: See https://github.com/acolomb/canopen/tree/fix-pdo-export for what I had in mind.  Passes tests and I don't get any related mypy errors either.

erlend-aasland: > See https://github.com/acolomb/canopen/tree/fix-pdo-export for what I had in mind. Passes tests and I don't get any related mypy errors either.

Sure, I knew what you meant, but thanks; I just think 013ff93cacb4261dd9b5c95dd7bb139056abab00 is a nicer approach 😁

sveinse: If my opinion carries any weight I'd vote for the latter proposal by @erlend-aasland . I think its compact and to the point. and the simplest of the two. The downside to it is that linters will complain about import inside functions and have to be muted. Conditional run-time dependencies requires some special trickeries regardless how its solved.

erlend-aasland: > The downside to it is that linters will complain about import inside functions and have to be muted. Conditional run-time dependencies requires some special trickeries regardless how its solved.

This PR does not add new imports; not at module level, nor from within methods; **the imports are already there**.

You linter can be configured to not complain about imports from within a function/method.

acolomb (approved):

acolomb:
Please fix the conflicts, otherwise ready to merge.

acolomb (approved):

PR #494, opened 2024-07-03
README: Add formatting and amend doc build instructions

erlend-aasland:
Additionally:
- add formatting to Git branch names
- add formatting to Python module names
- ~~consistently recommend 'python3 -m pip' iso. 'pip'~~
- ~~consistently use 'python3' iso. 'python' for command line examples~~
- spell out Python version support

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@ The aim of the project is to support the most common parts of the CiA 301
 standard in a simple Pythonic interface. It is mainly targeted for testing and
 automation tasks rather than a standard compliant master implementation.
 
-The library supports Python 3.8+.
+The library supports Python 3.8 or newer.
 
 
 Features
@@ -36,25 +36,29 @@ Incomplete support for creating slave nodes also exists.
 Installation
 ------------
 
-Install from PyPI_ using pip::
+Install from PyPI_ using :program:`pip`::
 
-    $ pip install canopen
+    $ python3 -m pip install canopen
 
-Install from latest master on GitHub::
+Install from latest ``master`` on GitHub::
 
-    $ pip install https://github.com/christiansandberg/canopen/archive/master.zip
+    $ python3 -m pip install https://github.com/christiansandberg/canopen/archive/master.zip
 
 If you want to be able to change the code while using it, clone it then install
 it in `develop mode`_::
 
     $ git clone https://github.com/christiansandberg/canopen.git
     $ cd canopen
-    $ pip install -e .
+    $ python3 -m pip install -e .
 
 Unit tests can be run using the pytest_ framework::
 
-    $ pip install pytest
+    $ python3 -m pip install pytest
     $ pytest -v
+
+You can also use :mod:`unittest` standard library module::
+
+    $ python3 -m unittest discover test -v
 
 Documentation
 -------------
@@ -65,7 +69,8 @@ http://canopen.readthedocs.io/en/latest/
 
 It can also be generated from a local clone using Sphinx_::
 
-    $ python setup.py build_sphinx
+    $ python3 -m pip install -r doc/requirements.txt
+    $ make -C doc html
 
 
 Hardware support
--- a/README.rst
+++ b/README.rst
@@ -38,22 +38,22 @@ Installation
 
 Install from PyPI_ using :program:`pip`::
 
-    $ python3 -m pip install canopen
+    $ pip install canopen
 
 Install from latest ``master`` on GitHub::
 
-    $ python3 -m pip install https://github.com/christiansandberg/canopen/archive/master.zip
+    $ pip install https://github.com/christiansandberg/canopen/archive/master.zip
 
 If you want to be able to change the code while using it, clone it then install
 it in `develop mode`_::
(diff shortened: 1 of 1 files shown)


erlend-aasland:
BTW; CI runs should be disabled for PRs like this.

acolomb:
What's wrong with using `pip`?

acolomb (approved):

christiansandberg:
I don't think `python3` even works on Windows.

README.rst
```diff
@@ -36,26 +36,30 @@ Incomplete support for creating slave nodes also exists.
...
 
 Unit tests can be run using the pytest_ framework::
 
-    $ pip install pytest
+    $ python3 -m pip install pytest
     $ pytest -v
 
+You can also use :mod:`unittest` standard library module::
+
+    $ python3 -m unittest discover test -v
```
christiansandberg, line 61: I dont think we should rely on unittest anymore. It's fine to write new tests using pytest.

erlend-aasland:
André:
> What's wrong with using `pip` as a command?

IIRC, because it's easy to end up using the "wrong" `pip`. Explicitly executing `pip` as a module using your preferred `python3` command, ensures you're installing packages for the correct Python installation.

See also the `pip` docs: https://pip.pypa.io/en/stable/user_guide/

Christian:
> I don't think `python3` even works on Windows.

I don't use Windows, but yes, I believe you'll have to use `py -m pip`. IIRC, there are some differences depending on which shell you use (PowerShell vs. CMD.EXE vs. Git Bash, etc.), though.

André:
> Sorry, I have next to no experience with configuring GitHub actions. How would we exclude CI runs for certain changes? Is it worthwhile investigating that?

I also realised there is no docs build CI. I'll create an issue with some suggested CI improvements.

---

Thanks both of you for reviewing.

README.rst
erlend-aasland: I'm not sure how to interpret this. Do you mean that you don't want to promote running tests using `unittest`, or do you propose to do a blanket change and rewrite all tests using `pytest` APIs?

sveinse:
> What's wrong with using `pip` as a command?

On windows, running `pip` runs `pip.exe`. Any operations that involves update of pip itself will fail, as Windows unfortunately blocks the executable `pip.exe`. So it is common practice to run `python -m pip` (if venv) or `py -m pip` (if global) for that reason.

> I don't think `python3` even works on Windows.

`python3` is a non-Windows thing, and `py.exe` has been established as the go-to for Windows. However, all (three?) targets have `python` as executable when using a virtual environment [^1]. I find that as good middle ground for writing documentation. 

[^1]: &lt;python rant&gt; For some unfathomable reason, all OS _except_ Windows use `bin/` as the executable directory in virtual environments, while Windows use `Scripts\`. Its an irritating obstacle for making OS-independent interactions and scripts against virtual environments and extremely annoying. </python rant>

acolomb:
Well the Windows problem of `pip install -U pip` isn't relevant here, as that's not what the listed command is doing.

Please go back to `pip` and `python` as the most basic commands that illustrate what needs to be done.  Different environments may need adjustments, but there's no copy-paste-ability guarantee anywhere in the document. :-)

sveinse:
> If the `pip` maintainers think it's too dangerous to use the installed `pip[.exe]` binary, then they should remove it. People can always just do `alias pip="python -m pip"` if they frequently need it. But the intention here is to show what needs to be installed, not to introduce correct pip command line usage.

FYI: [Setuptools docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/) and [pip docs](https://pip.pypa.io/en/stable/getting-started/) docs consequently use `-m pip` in their documentation. I don't know why thou.

erlend-aasland:
> But the intention here is to show what needs to be installed, not to introduce correct pip command line usage.

Sure, but IMO those are not conflicting goals. Well, I updated the PR according to your remarks. PTAL.

acolomb (approved):

README.rst
acolomb: I understand it as better keep recommending `pytest` only and for *new* tests, target that API.  If someone has the time to port over the existing tests, go ahead.  I don't have a strong opinion here, as long as the tests can be invoked easily.

erlend-aasland: What's the rationale? `pytest` understands `unittest` perfectly well. If you want to run the test suite using `pytest`, go ahead. Ditto for `unittest`. Using `unittest` to write the tests makes for greater freedom when running the test suite. Anyway, this is better discussed in a separate issue.

acolomb: :man_shrugging:

Let's not hold this up by such a minor detail.  If anyone has strong feelings about it, let's handle in a new issue.

acolomb:
@erlend-aasland Hm... Unfortunately this broke our PyPI uploads, as the Sphinx roles in README.rst are not acceptable for the parsing on PyPI.  Will have to revert some of the formatting and try again with a v2.4.1 release.

See https://github.com/canopen-python/canopen/actions/runs/16746953346/job/47407670531

PR #498, opened 2024-07-07
CI: Build docs in CI

### LINKED ISSUE #497: Test docs build in CI
Currently, the documentation build is not tested by `pythonpackage.yml`. I suggest adding a separate workflow for docs changes only, and to run `pythonpackage.yml` for code changes only.

Moreover, consider testing multiple Python versions in the CI matrix.

erlend-aasland:
Resolves #497

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,3 @@
-sphinx==7.3
-sphinx-autodoc-typehints==2.2
-furo==2024.5
+sphinx~=7.3
+sphinx-autodoc-typehints~=2.2
+furo~=2024.5
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,3 @@
-sphinx
-sphinx-autodoc-typehints
-furo
+sphinx==7.3.7
+sphinx-autodoc-typehints==2.2.2
+furo==2024.5.6
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,3 @@
-sphinx==7.3.7
-sphinx-autodoc-typehints==2.2.2
-furo==2024.5.6
+sphinx==7.3
+sphinx-autodoc-typehints==2.2
+furo==2024.5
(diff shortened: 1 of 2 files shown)


doc/requirements.txt
```diff
@@ -1,3 +1,3 @@
-sphinx
-sphinx-autodoc-typehints
-furo
+sphinx==7.3.7
+sphinx-autodoc-typehints==2.2.2
+furo==2024.5.6
```
acolomb, line 3: Why are you pinning the versions here?  At the least, I would expect to list only minimum, but not specific versions.

acolomb:
Our "CI" for the docs actually lives on readthedocs.org, as each commit on `master` is built there under the `latest` version.  Testing here may actually lead to less PRs getting merged that break the online docs, so that's positive.

This doesn't make an effort though to separate action runs between docs-only and code-only changes, as mentioned in #497?  Since any code change also affects the docs, and docs-only fixes are rare, I can live with that.

erlend-aasland:
> This doesn't make an effort though to separate action runs between docs-only and code-only changes, as mentioned in https://github.com/christiansandberg/canopen/issues/497? Since any code change also affects the docs, and docs-only fixes are rare, I can live with that.

Yes. I'm not used to the docstrings being so tightly coupled with the rendered docs; I simply forgot that it's not possible to do that separation with the current setup.

doc/requirements.txt
erlend-aasland: Hm, yes this may be a tad too specific. Using `~=major.minor` or `~=major` would be a better option.

acolomb: I'd suggest leaving them unspecified or give only a minimum. What do we gain by forbidding newer versions?  It seems to have worked all this time.

erlend-aasland: I prefer to ping dependencies at some level, often at the minor version (assuming SemVer). Having a fairly reproducible test environment is a nice thing.

acolomb:
I just hope Someone(TM) will be there to bump these versioned dependencies when they have bit-rotted enough ;-)

acolomb:
So it seems there are no artifacts?  Wouldn't it make sense to build some kind of HTML bundle while we're already generating the docs?

erlend-aasland:
> So it seems there are no artifacts? Wouldn't it make sense to build some kind of HTML bundle while we're already generating the docs?

That's a very good idea.

PR #499, opened 2024-07-07
Git: correctly ignore docs build directory

erlend-aasland:


.gitignore
```diff
@@ -27,6 +27,10 @@ var/
 *.egg
 build-deb/
 _version.py
+venv*/
```
sveinse, line 30: I attempted such change 2 weeks ago, but it was discouraged, see https://github.com/christiansandberg/canopen/pull/468/files/5c2343afd7af3e225d1101285d0562b7969b46e0#r1648161548

erlend-aasland: That's too bad. Might as well just close this, then 🤷

erlend-aasland: Well, the `docs` fix is still relevant.

erlend-aasland:
FTR; here's the current variant of the gitignore template that this project's `.gitignore` was based on:

https://github.com/github/gitignore/blob/main/Python.gitignore

Perhaps it is time to sync with the modern variant.

acolomb:
Wow, that template really has a lot of stuff listed.  I hope I'm not alone with my opinion though - what's ignored in a project's Git configuration should be based on *objective* decisions *relevant to the code-base*.  Thus I don't want any IDE state crap (e.g. `.vscode`) there, nor temporary files used by random tools (`*.swp`) or operating systems (`.DS_Store`).  In my view, all these are specific to the individual developer's environment and should therefore be added as "local" or "system-wide" ignores in Git (see [GitHub's docs ](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)).

The version-tracked ignore patterns should target files that are expected to be generated by the build process(es) and should be hidden to avoid accidently adding generated files to the versioning.  Thus this (remaining) change is fine, as the docs will land at the specified location.

Putting a `venv` directory inside the source worktree is a concious choice by the developer, thus their individual responsibility.  It can be ignored locally via `.git/info/exclude`.

Temporary files and system caches can be handled the same way, or even globally via `~/.config/git/ignore`.

If anything, we should clean out the current `.gitignore` to get rid of all this template stuff that someone imagined could ever be there.  For adding new stuff, my current policy is outlined above, and tries to avoid piling up stuff from random trending IDEs and such, because there will be no end to the proliferation.

Issue #500, opened 2024-07-08
`NmtBase.state` should always return a `str`


erlend-aasland:
## Problem

`NmtBase.state` is documented and annotated as always returning a string, however there is a code path that may return an `int`:

https://github.com/christiansandberg/canopen/blob/62e9c1f851cb8721355d22c8571f2b41599c1b36/canopen/nmt.py#L74-L92

The code path in question is highly unlikely; the `NmtBase._state` member is set in the following places:

- ✅ `NmtBase.__init__`: explicitly set to 0
- ✅ `NmtBase.on_command`: set _iff_ the new state is a valid/known state
- ✅ `NmtBase.send_command`: set _iff_ the new state is a valid/known state
- ❓ `NmtMaster.on_heartbeat`: set to whatever was decoded from the heartbeat message, which _should_ be fine for any compliant device

## Possible solutions

In `NmtMaster.on_heartbeat`, only set `_state` if it is a valid state; if it's not, `log.error(...)` or raise an exception.

In `NmtBase.state`, either raise an exception if `._state` is not a valid state, or return something a la `f"UNKNOWN STATE {self._state}"`.

<details>
<summary>Draft diff</summary>

```diff
diff --git a/canopen/nmt.py b/canopen/nmt.py
index 401ad15..1316a5c 100644
--- a/canopen/nmt.py
+++ b/canopen/nmt.py
@@ -86,10 +86,8 @@ class NmtBase:
         - 'RESET'
         - 'RESET COMMUNICATION'
         """
-        if self._state in NMT_STATES:
-            return NMT_STATES[self._state]
-        else:
-            return self._state
+        assert self._state in NMT_STATES:
+        return NMT_STATES[self._state]
 
     @state.setter
     def state(self, new_state: str):
@@ -122,6 +120,12 @@ class NmtMaster(NmtBase):
             logger.debug("Received heartbeat can-id %d, state is %d", can_id, new_state)
             for callback in self._callbacks:
                 callback(new_state)
+            if new_state not in NMT_STATES:
+                log.error(
+                    "Received heartbeat can-id %d with invalid state %d",
+                    can_id, new_state
+                )
+                return
             if new_state == 0:
                 # Boot-up, will go to PRE-OPERATIONAL automatically
                 self._state = 127
```

</details>

sveinse:
Awesome. I've found the same thing while type annotating, so this discussion is on my list 👍 

[...]

PR #503, opened 2024-07-08
Tests: establish test/util.py and test/test_nmt.py

erlend-aasland:
Give NMT tests their own file, and establish a util.py for the test
suite for common stuff like the location of sample EDS files.

### DIFF
--- a/test/util.py
+++ b/test/util.py
@@ -3,4 +3,3 @@ import os
 
 DATATYPES_EDS = os.path.join(os.path.dirname(__file__), "datatypes.eds")
 SAMPLE_EDS = os.path.join(os.path.dirname(__file__), "sample.eds")
-EDS_PATH = SAMPLE_EDS
--- a/test/util.py
+++ b/test/util.py
@@ -1,0 +1,6 @@
+import os
+
+
+DATATYPES_EDS = os.path.join(os.path.dirname(__file__), "datatypes.eds")
+SAMPLE_EDS = os.path.join(os.path.dirname(__file__), "sample.eds")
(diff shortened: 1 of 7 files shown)


erlend-aasland:
Yak-shave in preparation for #500

erlend-aasland:
Note that the NMT slave test now uses `setUp`/`tearDown` iso. their class equivalents, in order to provide a saner environment for the test cases.

erlend-aasland:
Moreover, I removed some dead code.

test/util.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+
+DATATYPES_EDS = os.path.join(os.path.dirname(__file__), "datatypes.eds")
+SAMPLE_EDS = os.path.join(os.path.dirname(__file__), "sample.eds")
+EDS_PATH = SAMPLE_EDS
```
acolomb, line 6: Shouldn't we get rid of this old name everywhere?  Might just as well adjust the other usages to the new name.

acolomb (approved):

test/util.py
erlend-aasland: Sure; I just wanted to keep the diff to a minimum. I'll update the PR to consistently use the new name only.

erlend-aasland: Addressed with 0905d017c69ee9f3727dc3480591ddb073fc746b.

acolomb (approved):

erlend-aasland:
Thanks for the review!

acolomb:
Thanks for your continued contributions!

PR #504, opened 2024-07-08
Add tests for NMT base and NMT master

erlend-aasland:


test/test_nmt.py
```diff
@@ -1,21 +1,137 @@
...
+            "PRE-OPERATIONAL",
+            "SLEEP",
+            "STANDBY",
+            "STOPPED",
+        )
+        for cmd in dataset:
+            with self.subTest(cmd=cmd):
+                code = NMT_COMMANDS[cmd]
+                self.nmt.send_command(code)
+                self.assertNotEqual(self.nmt.state, "INITIALISING")
```
erlend-aasland, line 28: Alternatively something more precise, kind of like this:

```suggestion
                self.assertEqual(self.nmt.state,
                                 NMT_STATES[COMMAND_TO_STATE[code]])
```

erlend-aasland:
More yak shave for #500.

test/test_nmt.py
acolomb: Sure, seems simple enough and if it breaks, then something in the tested code must have gone wrong.

test/test_nmt.py
```diff
@@ -1,21 +1,136 @@
...
+            self.node.nmt.wait_for_bootup(self.TIMEOUT)
+
+    def test_nmt_master_on_heartbeat(self):
+        # Skip the special INITIALISING case.
+        for code in [st for st in NMT_STATES if st != 0]:
+            with self.subTest(code=code):
+                task = self.net.send_periodic(self.COB_ID, [code], self.PERIOD)
+                self.addCleanup(task.stop)
+                actual = self.node.nmt.wait_for_heartbeat(self.TIMEOUT)
+                task.stop()
```
acolomb, line 78: Is this safe to be called twice?  You added it as a cleanup callback above.

test/test_nmt.py
```diff
@@ -1,21 +1,136 @@
...
+        self.node.nmt.wait_for_bootup(self.TIMEOUT)
+        state = self.node.nmt.wait_for_heartbeat(self.TIMEOUT)
+        self.assertEqual(state, "PRE-OPERATIONAL")
+
+    def test_nmt_master_on_heartbeat_unknown_state(self):
+        task = self.net.send_periodic(self.COB_ID, [0xcb], self.PERIOD)
+        self.addCleanup(task.stop)
+        state = self.node.nmt.wait_for_heartbeat(self.TIMEOUT)
+        # Expect the high bit to be masked out, and the resulting integer
+        # returned as it is. See gh-500 for the data type inconsistency.
```
acolomb, line 94: What is a gh-500?  I don't really like references to a development platform within the code itself.  As it stands, this should be considered undefined behavior, better not test for a value that we know should be fixed.

acolomb (approved):
Wow, this is impressive.  Thank you very much for your diligence.  I could never have come up with all these test cases...

test/test_nmt.py
erlend-aasland: > Sure, seems simple enough [...]

The PR as it stands, or the suggestion? :)

test/test_nmt.py
erlend-aasland: Yeah, it's safe. I added the cleanup task in the case where `wait_for_heartbeat` fails, _but_ the cleanup tasks are not run until the actual test method completes (that is: after the loop ends), so you'll be starting a lot of periodic tasks on the bus, creating completely havoc in the test environment. OTOH, I could just as well use a `try...except NmtError...finally` instead. Let me rewrite this part; I think it will be clearer.

erlend-aasland: I think c2ff667bbb06a4fb2b6a10d5703ce5df211d29f9 may be a little bit easier to read and reason about.

test/test_nmt.py
erlend-aasland: Rewritten with c2ff667bbb06a4fb2b6a10d5703ce5df211d29f9 to expect the upcoming change in decided in gh-500.

test/test_nmt.py
```diff
@@ -1,21 +1,140 @@
...
+
+
+class TestNmtMaster(unittest.TestCase):
+    NODE_ID = 2
+    COB_ID = 0x700 + NODE_ID
+    PERIOD = 0.01
+    TIMEOUT = PERIOD * 2
+
+    def setUp(self):
+        bus = can.ThreadSafeBus(
```
erlend-aasland, line 48: Actually, this does not need to be a thread-safe bus; OTOH, there's no harm in using one.

erlend-aasland:
FTR, I will follow up with a PR for hardening the NMT slave tests.

test/test_nmt.py
acolomb: This was related only to the suggestion.

test/test_nmt.py
```diff
@@ -1,21 +1,140 @@
...
+
+    def test_nmt_master_on_heartbeat(self):
+        # Skip the special INITIALISING case.
+        for code in [st for st in NMT_STATES if st != 0]:
+            with self.subTest(code=code):
+                task = self.net.send_periodic(self.COB_ID, [code], self.PERIOD)
+                try:
+                    actual = self.node.nmt.wait_for_heartbeat(self.TIMEOUT)
+                except NmtError:
+                    self.fail("Timed out waiting for heartbeat")
```
acolomb, line 79: Will this include the exception message itself?

erlend-aasland: Yes, but actually, we don't need the `except` clause; we can just use `try...finally` instead.

erlend-aasland: Simplified with 6fa2986d6253b47315e96cf8f7cee93e5dd84298. Marking as resolved.

acolomb (approved):

Issue #501, opened 2024-07-08
pytest fails on master due to TestEDS.test_export_eds_to_file


sveinse:
I'm running Windows and the tests fails:

```
TestEDS.test_export_eds_to_file
TestEDS.test_export_eds_to_file_unknown_extension
```

Both fails with:

```
test\test_eds.py:235:    # The second is test fails on line 245
canopen\objectdictionary\__init__.py:58: PermissionError
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\sveinse\\AppData\\Local\\Temp\\tmp00_selbk.eds'
```

I'm not sure why this specific test doesn't work in Windows while the others works fine.

erlend-aasland:
Sorry, this my fault; I used `tempfile.NamedTemporaryFile`, which returns a writable and opened file. `canopen` tries to open it again, which works fine on Linux and macOS, but fails on Windows.

PR #505, opened 2024-07-08
Tests: harden TestNetwork

erlend-aasland:
- make tests deterministic
- use cleanup hooks

### DIFF
--- a/test/test_network.py
+++ b/test/test_network.py
@@ -1,4 +1,3 @@
-import time
 import unittest
 
 import canopen
@@ -10,7 +9,8 @@ class TestNetwork(unittest.TestCase):
 
     def setUp(self):
         network = canopen.Network()
-        network.add_node(2, SAMPLE_EDS)
+        with self.assertLogs():
+            network.add_node(2, SAMPLE_EDS)
         network.add_node(3, network[2].object_dictionary)
         self.network = network
 
@@ -31,7 +31,10 @@ class TestNetwork(unittest.TestCase):
 
     def test_send(self):
         bus = can.interface.Bus(interface="virtual", channel=1)
+        self.addCleanup(bus.shutdown)
+
         self.network.connect(interface="virtual", channel=1)
+        self.addCleanup(self.network.disconnect)
 
         # Send standard ID
         self.network.send_message(0x123, [1, 2, 3, 4, 5, 6, 7, 8])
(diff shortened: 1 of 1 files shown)


test/test_network.py
```diff
@@ -48,33 +51,42 @@ def test_send(self):
...
-        while msg is not None and msg.data == b'\x01\x02\x03':
-            msg = bus.recv(0)
-        self.assertIsNotNone(msg)
-        self.assertSequenceEqual(msg.data, [4, 5, 6])
-        task.stop()
+        acc = []
+        event = Event()
+
+        def hook(id_, data, ts):
+            acc.append(data)
```
erlend-aasland, line 72: We *could* also add an accumulator for the timestamps and do an `assertAlmostEqual` to validate the interval.

erlend-aasland:
I did some experiments with a full-blown CI matrix[^1] and discovered that `test_send_periodic` failed 99% of the times on the macOS CI. I think it would be valuable to run the CI on different platforms. See also #501 for a Windows specific (?) error.

[^1]: all supported Python versions x [windows, macos, linux]

test/test_network.py
```diff
@@ -48,33 +51,42 @@ def test_send(self):
         self.assertEqual(msg.arbitration_id, 0x12345)
         self.assertTrue(msg.is_extended_id)
 
-        bus.shutdown()
-        self.network.disconnect()
+    def test_send_periodic(self):
+        from threading import Event
```
acolomb, line 55: This changes test methodology in two ways AFAICT: Not checking the amount of packets received, and using `network.subscribe()` instead of `bus.recv()`.  Is there a good reason for the latter?  In #504 you do use a separate `ThreadSafeBus` to receive the messages from the virtual interface.  Shouldn't the same apply here?

As for the import, I see no necessity to do that locally instead of at the top of the module.

test/test_network.py
acolomb: Sounds good, although a longer interval will probably help to get better accuracy here.

acolomb:
> I think it would be valuable to run the CI on different platforms.

As long as GitHub covers the additional computation costs, sure why not.  Although I think Python 3.8 (oldest supported) and the two most recent Python releases (3.11, 3.12 currently) should be fine.

test/test_network.py
erlend-aasland: I agree we should still check the number of messages received. Let me work a little bit more on this; marking it as draft for now.

erlend-aasland: I think 0af94f7ed1b35266ac9a6b1344467f96f09ab4c6 should work pretty well. It checks the number of packets received, and roughly checks the period. Should be pretty deterministic.

test/test_network.py
erlend-aasland: Resolved with 0af94f7ed1b35266ac9a6b1344467f96f09ab4c6.

erlend-aasland:
> > I think it would be valuable to run the CI on different platforms.
> 
> As long as GitHub covers the additional computation costs, sure why not. Although I think Python 3.8 (oldest supported) and the two most recent Python releases (3.11, 3.12 currently) should be sufficient.

Sounds good. I'll pair it with a matrix of operating systems, post landing this PR.

test/test_network.py
acolomb: How certain are we that it won't catch two periodic messages?  The new check depends on exactly two packages being received.

acolomb: And what about the first comment, using `network.subscribe()` instead of `bus.recv()`?

erlend-aasland: > How certain are we that it won't catch two periodic messages? The new check depends on exactly two packages being received.

Pretty certain, but I guess with a severely overloaded runner, you could possibly end up in a place where the background periodic task gets to shoot twice before we get to the check.

> And what about the first comment, using `network.subscribe()` instead of `bus.recv()`?

Sure, we can use `recv()` instead, I just thought `subscribe()` might be easier to read. It does not really matter which tool we use, as long as we're testing the right thing. I can draft up a proposal using `recv` instead.

acolomb: > Pretty certain, but I guess with a severely overloaded runner, you could possibly end up in a place where the background periodic task gets to shoot twice before we get to the check.

Then let's just wait and see how often that actually happens.

> I can draft up a proposal using `recv` instead.

My concern was mainly that we now implicitly test `.subscribe()` as well, which we didn't before.

erlend-aasland: Got it.

I've got the current PR running in an infinite loop in the background; no flakiness observed (yet).

All of the experiments I've done with `.recv()` have been extremely flaky, so if you're fine with using `.subscribe()`, I'd prefer that (at least for now). It's a little bit verbose, but it's deterministic.

acolomb: Sure, there's nothing wrong with it.  Just wanted to raise awareness of using the additional method in the test.  I'm not strict about that though, any failures will have to be investigated by looking at the whole stack anyway.

acolomb (approved):

erlend-aasland:
I really appreciate the thoroughness of your reviews; thanks a lot!

acolomb:
Oh, just noticed my commit message erroneously referred to heartbeats in the periodicity stuff.  Though this is lower level, I just got confused.

erlend-aasland:
> Oh, just noticed my commit message erroneously referred to heartbeats in the periodicity stuff. Though this is lower level, I just got confused.

No sweat, you're fine; as you said yourself, no Git log is perfect :)

acolomb:
Well, apparently only the `socketcan` interface in python-can implements the required `modify_data()` method.  Otherwise, it falls back to stopping and restarting the task, which the comment in `update()` explains as messing up the period.

> BTW, I still can't get reproducible tests using `.recv()`. Perhaps it is buggy?

Haven't really looked at the implementation, but I thought it was just a lower level call in the underlying python-can library.  It's probably a difference between waiting / polling messages and installing a notifier callback (which `network.subscribe()` does.

erlend-aasland:
> It's probably a difference between waiting / polling messages and installing a notifier callback (which `network.subscribe()` does.

Yep, sounds plausible.

PR #508, opened 2024-07-09
Tests: harden TestNetwork and TestNmt even more

erlend-aasland:
- test task.stop()
- use more lenient timeout
- ~~make sure we always collect two frames~~

See commit messages for details.

### DIFF
--- a/test/test_network.py
+++ b/test/test_network.py
@@ -1,5 +1,5 @@
-import time
 import unittest
+from threading import Event
 
 import canopen
 import can
@@ -10,7 +10,8 @@ class TestNetwork(unittest.TestCase):
 
     def setUp(self):
         network = canopen.Network()
-        network.add_node(2, SAMPLE_EDS)
+        with self.assertLogs():
+            network.add_node(2, SAMPLE_EDS)
         network.add_node(3, network[2].object_dictionary)
         self.network = network
 
@@ -31,7 +32,10 @@ class TestNetwork(unittest.TestCase):
 
     def test_send(self):
         bus = can.interface.Bus(interface="virtual", channel=1)
+        self.addCleanup(bus.shutdown)
+
         self.network.connect(interface="virtual", channel=1)
+        self.addCleanup(self.network.disconnect)
 
         # Send standard ID
         self.network.send_message(0x123, [1, 2, 3, 4, 5, 6, 7, 8])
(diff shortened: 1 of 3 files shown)


erlend-aasland:
Follow-up for shortcomings in #505.

erlend-aasland:
@acolomb, some rationale for the changes:

- During tests with macOS runners, I found that they very often timed out[^1], hence I've increased the timeout considerably.
- Since `.update()` restarts the timer, we can't use a frame from dataset 1 and a frame from dataset 2 to validate the periodicity; we need to collect exactly two frames from each "session" in order to test this deterministically.
- Added missing test of `task.stop()` functionality.
- Introduced a `check()` helper for the sake of readability.

I think this should be fine. I'll mark it as ready for review tomorrow morning CET ☕

[^1]: I've only seen this on macOS runners, never on Ubuntu nor Windows.

test/test_network.py
```diff
@@ -65,30 +66,52 @@ def test_send_periodic(self):
...
         self.addCleanup(self.network.unsubscribe, COB_ID)
 
         task = self.network.send_periodic(COB_ID, DATA1, PERIOD)
         self.addCleanup(task.stop)
 
-        event.wait(PERIOD*2)
-
-        # Update task data.
+        def periodicity():
+            # Check if periodicity is established; flakiness have been observed
```
erlend-aasland, line 83:
```suggestion
            # Check if periodicity is established; flakiness has been observed
```

test/test_network.py
```diff
@@ -65,30 +66,52 @@ def test_send_periodic(self):
...
+        idx = data.index(DATA2)
+        self.assertTrue(all(v[0] == DATA2 for v in acc[idx:]))
+
+        # Stop the task.
+        task.stop()
+        # A message may have been in flight when we stopped the timer,
+        # so allow a single failure.
+        bus = self.network.bus
+        msg = bus.recv(TIMEOUT)
+        if msg:
```
erlend-aasland, line 113: Be slightly more specific.

```suggestion
        if msg is not None:
```

test/test_nmt.py
```diff
@@ -117,7 +117,11 @@ def test_nmt_master_node_guarding(self):
         self.assertEqual(msg.dlc, 0)
 
         self.node.nmt.stop_node_guarding()
-        self.assertIsNone(self.bus.recv(self.TIMEOUT))
+        # A message may have been in flight when we stopped the timer,
+        # so allow a single failure.
+        msg = self.bus.recv(self.TIMEOUT)
+        if msg:
```
erlend-aasland, line 123:
```suggestion
        if msg is not None:
```

erlend-aasland:
After this PR has landed, I'll follow-up with a small PR to expand the CI matrix so we can run the test suite on macOS and Windows, as well as on Linux. That should prevent failures like #501.

erlend-aasland:
> Are we on a free tier or something paid for? And are we well away from any spending limits on actions? -- Adding new CI matrixes is great, but it triples build minutes, so I think maybe we should give the owner a chance to ok this.

Ordinary user accounts with public repos are given a pretty large batch of CI minutes per month for free. Anyway, let's discuss this in the following-up PR; it's not related to _this_ change.

PR #510, opened 2024-07-09
Purge import hack from network.py

### LINKED ISSUE #509: Purge workaround for docs build
`canopen/network.py` contains the following docs build workaround:

https://github.com/christiansandberg/canopen/blob/3aa509db9fea8094516e5eb717be3df182f0751a/canopen/network.py#L8-L17

I suggest to purge this hack from `network.py`; the RTD config already installs `canopen` as a package, which implicitly installs `can`. Moreover, purging this workaround will make it easier to add type annotations for `network.py`.

erlend-aasland:
Always require dependencies to be installed.

Resolves #509

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -5,16 +5,9 @@ import logging
 import threading
 from typing import Callable, Dict, Iterator, List, Optional, Union
 
-try:
-    import can
-    from can import Listener
-    from can import CanError
-except ImportError:
-    # Do not fail if python-can is not installed
-    can = None
-    CanError = Exception
-    class Listener:
-        """ Dummy listener """
+import can
+from can import Listener
+from can import CanError
 
 from canopen.node import RemoteNode, LocalNode
 from canopen.sync import SyncProducer


friederschueler (approved):

sveinse (approved):
Approved with joy. Thank you @erlend-aasland

erlend-aasland:
Thanks for the review!

acolomb:
Sorry that was too quick for me. Would you please leave me some chance to actually review? I haven't grasped at all what this was about and wanted to ask questions. Just too many things at once, and now it moves by too fast...

sveinse:
Should all PRs run through you @acolomb? If they should, I didn't know about it, I'm sorry. I assumed that two collaborators approving were enough, but I might have been mistaking. Do we have some rules of engagement anywhere in this project so we can avoid running into similar situations?

acolomb:
No rules that I know of and in general, no I don't insist on approving everything. But in this case I did have some questions. And the other discussions running in parallel took away enough time so I hadn't had a chance to ask them. Just one day to react is a bit short if you're working on this voluntarily, beside $dayjob.

sveinse:
I understand. And please do not read this as being flippant or self-imposing, but as being rather curious: I count several recently closed PRs that have lifetimes well shorter than 24hrs from opening to merge, so this is not the first. If collaborators have somewhat equal rights then there will be some PR that runs through while not all are available -- even ones where one might have questions. That's a part of running a open source project for us all. So that's why I asked specifically if you are appointed lead and need to explicitly approve PRs.

acolomb:
The difference lies in the nature of the change.  Several other PRs lately had been discussed previously or had little impact and were pretty straightforward or purely additions.  This one OTOH has the potential to make the module fail to import when it previously was working.  I haven't grasped the justification of the workaround (or the scenario where it's needed) in the first place.  But before potentially breaking some user's application with the next update, I'd like to understand fully where this came from and why it's no longer necessary.

And no, "this makes type annotations easier" is not enough of an explanation :-)

Issue #511, opened 2024-07-10
Network instance is optional in many classes, which requires lot of checking


sveinse:
With reference to #509 - lots of classes depend on an instance of `class Network` and have it as an attribute in order to communicate with the CAN bus. Examples are `SdoBase()`, `PdoBase()`, `NmtBase()` and others. Their constructor does not require a `Network` argument so the formal type of the attribute is `network: Optional[Network] = None`.

These attributes are set when a `RemoteNode()` is created and the reference to the network is injected from `RemoteNode.associate_network()`.

The effect of this is that we'll need a lot of checks all over the code to ensure network is properly set. E.g. as here from `class EmcyProducer`:

```python
    def send(self, code: int, register: int = 0, data: bytes = b""):
        payload = EMCY_STRUCT.pack(code, register, data)
        if self.network is None:   # This is needed to not fail the send_message line
            raise RuntimeError("A Network is required")
        self.network.send_message(self.cob_id, payload)
```

Is is a requirement to be able to create these class instances without a functional Network? We could make a default Network-like object that will emulate network operations if we need it to work without an actual CAN bus network.

What do you think we should do here?

erlend-aasland:
IMO, the better thing to do would be to improve the documentation to put emphasis on always using `Network.add_node` and `Network.create_node`[^1] to create nodes.

BTW, are there use-cases for subclassing the \*Base classes you mention?

> We could make a default Network-like object that will emulate network operations if we need it to work without an actual CAN bus network.

I would instead prefer keeping the status quo.

[^1]: It is slightly confusing that `add_node` and `create_node` are used for creating remote and local nodes respectively. Something like `create_remote_node` and `create_local_node` would be clearer.

sveinse:
> I would instead prefer keeping the status quo.

I've added all the runtime checks in PR #513 to ensure the `None` case of `Optional[Network]` is properly handled. We _could_ avoid all that if `Network` weren't `Optional`. But in many ways, I agree because having as it is today requires less intrusive changes to the codebase.

acolomb:
[...]

PR #512, opened 2024-07-10
Add mypy support and fixup project to give no errors

sveinse:
With reference to #358 this PR is the starting point of adding typing with mypy support in canopen. It sets a recommended permissive starting point for mypy and it contains the absolute _minimal_ code change required to pass all mypy errors.

It does contain behavioral changes that were required to pass the tests. The majority of these fixes are more graceful handling of errors that would occur anyways, so the behavioral change is what error they produce.

* Add runtime test for `self.network` before using the network
* `PeriodicMessageTask.update()` don't stop the task unless it's running
* `Variable.desc` ensure that the object is int
* `Variable.read()` fail with ValueError unless a valid fmt is used
* `Variable.write()` ensure the description is a string
* `BaseNode.__init__()` fail if no node_id is provided
* `ObjectDictionary.__getitem__()` when splitting "." only return if the object is not an `ODVariable`
* `ODRecord.__eq__()`, `ODArray.__eq__()` and `ODVariable.__eq__()` test type of other before comparing
* `ODVariable.encode_raw()`, `.decode_phys()`, `.encode_phys()` add type tests of ensure the input is of correct type
* `PdoMap` various methods: ensure necessary attributes are not None

### DIFF
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import struct
 import logging
 import threading
@@ -52,7 +53,7 @@ class EmcyConsumer:
 
     def wait(
         self, emcy_code: Optional[int] = None, timeout: float = 10
-    ) -> "EmcyError":
+    ) -> Optional[EmcyError]:
         """Wait for a new EMCY to arrive.
 
         :param emcy_code: EMCY code to wait for
@@ -86,10 +87,14 @@ class EmcyProducer:
         self.cob_id = cob_id
 
     def send(self, code: int, register: int = 0, data: bytes = b""):
+        if self.network is None:
+            raise RuntimeError("A Network is required")
         payload = EMCY_STRUCT.pack(code, register, data)
         self.network.send_message(self.cob_id, payload)
 
     def reset(self, register: int = 0, data: bytes = b""):
+        if self.network is None:
+            raise RuntimeError("A Network is required")
         payload = EMCY_STRUCT.pack(0, register, data)
         self.network.send_message(self.cob_id, payload)
 
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -90,14 +90,10 @@ class EmcyProducer:
         self.cob_id = cob_id
 
     def send(self, code: int, register: int = 0, data: bytes = b""):
-        if self.network is None:
-            raise RuntimeError("A Network is required")
         payload = EMCY_STRUCT.pack(code, register, data)
         self.network.send_message(self.cob_id, payload)
 
     def reset(self, register: int = 0, data: bytes = b""):
-        if self.network is None:
-            raise RuntimeError("A Network is required")
         payload = EMCY_STRUCT.pack(0, register, data)
         self.network.send_message(self.cob_id, payload)
 
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -3,18 +3,21 @@ from __future__ import annotations
 from collections.abc import MutableMapping
 import logging
 import threading
-from typing import Callable, Dict, Iterator, List, Optional, Union
+from typing import Callable, Dict, Iterator, List, Optional, Union, TYPE_CHECKING, TextIO
 
 try:
     import can
     from can import Listener
     from can import CanError
 except ImportError:
-    # Do not fail if python-can is not installed
-    can = None
-    CanError = Exception
-    class Listener:
-        """ Dummy listener """
+    # Type checkers don't like this conditional logic, so it is only run when
+    # not type checking
+    if not TYPE_CHECKING:
+        # Do not fail if python-can is not installed
+        can = None
+        CanError = Exception
+        class Listener:
+            """ Dummy listener """
 
 from canopen.node import RemoteNode, LocalNode
 from canopen.sync import SyncProducer
@@ -23,6 +26,9 @@ from canopen.nmt import NmtMaster
 from canopen.lss import LssMaster
 from canopen.objectdictionary.eds import import_from_node
 from canopen.objectdictionary import ObjectDictionary
+
+if TYPE_CHECKING:
+    from can.typechecking import CanData
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


erlend-aasland:
This is great! Can you separate out all the behavioural changes in separate PRs accompanied with tests?

erlend-aasland:
Just to be explicit: in #358 there is an agreement on not mixing type annotations with other types of changes, including bug fixes and behavioural changes.

sveinse:
Yes, its possible to separate it out. While I see the intentions of separating out the type annotations from the behavioral changes, I'm not sure I like where this is going. Type annotating an existing project and behavioral changes goes iteratively hand in hand. Annotate something and an error pop up. Fix an error, and another annotation error appears. And so on. The sheer volume of PRs per behavioral change will be too great for me to handle. I won't be able to dedicate that much time to this. This PR is just a minor extract of what is to come ....at least if we're aiming for strict mypy. So perhaps we need to go back to #358 and really discuss what ambitions we have for type annotations to begin with.

erlend-aasland:
Yes, it is time-consuming, and it requires multiple PRs. IMO it is definitely worth it in the long run.

acolomb:
Yes, please let's first reach consensus on that discussion about the ambitions and the roadmap.  Will chime in there soon, but I can't afford looking at it every day right now.  Please have a little patience :-)

sveinse:
How many pieces must this PR be split into?

erlend-aasland:
> How many pieces must this PR be split into?

From a _quick_ glance at these changes, ISTM something like this is a sane approach[^1]:

- #511
- issue: discuss and agree on `add_node` / `create_node` changes
- issue (similar to gh-511): what to do in BaseNode when `node_id` is `None`
- issue (similar to the above): how to handle `cob_id` is `None` in PdoBase
- PR: add `WriteCallback` / `ReadCallback` protocols (I believe these should not be public APIs, no?)
- PR: harden OD encoding/decoding
- PR: harden OD equality operator

I might have missed some; there's _a lot_ of different types of changes here.

[^1]: For _each_ issue, time is best spent on reaching a consensus for a desired solution _before_ proposing a PR. Anything else is often a waste of reviewer time.

erlend-aasland:
When the above preparations have landed, it will be time to introduce annotations, which should then be a much easier PR to review, as it would do only one thing: introduce annotations.

erlend-aasland:
> The sheer volume of PRs per behavioral change will be too great for me to handle. I won't be able to dedicate that much time to this.

That's fine; this is open source. Let me know if you want me to help our with any of the preliminary tasks.

sveinse:
Please go ahead

canopen/emcy.py
```diff
@@ -52,7 +53,7 @@ def reset(self):
 
     def wait(
         self, emcy_code: Optional[int] = None, timeout: float = 10
-    ) -> "EmcyError":
+    ) -> Optional[EmcyError]:
```
acolomb, line 56: Cosmetics, let's keep these separate from the mypy error fixing.  There are lots of things like this to do, but better to keep a PR focused and revisit such cleanups later. IMHO.

canopen/emcy.py
```diff
@@ -86,10 +87,14 @@ def __init__(self, cob_id: int):
         self.cob_id = cob_id
 
     def send(self, code: int, register: int = 0, data: bytes = b""):
+        if self.network is None:
+            raise RuntimeError("A Network is required")
```
acolomb, line 91: Superseded by #525?

canopen/network.py
```diff
@@ -3,18 +3,21 @@
...
-    class Listener:
-        """ Dummy listener """
+    # Type checkers don't like this conditional logic, so it is only run when
+    # not type checking
+    if not TYPE_CHECKING:
+        # Do not fail if python-can is not installed
+        can = None
+        CanError = Exception
+        class Listener:
+            """ Dummy listener """
```
acolomb, line 20: Fixed in master, I believe?

canopen/network.py
```diff
@@ -45,7 +51,7 @@ def __init__(self, bus: Optional[can.BusABC] = None):
         #: List of :class:`can.Listener` objects.
         #: Includes at least MessageListener.
         self.listeners = [MessageListener(self)]
-        self.notifier = None
+        self.notifier: Optional[can.Notifier] = None
```
acolomb, line 54: Factored out to #534

canopen/network.py
```diff
@@ -138,15 +144,15 @@ def __exit__(self, type, value, traceback):
...
+        node: Union[int, RemoteNode],
+        object_dictionary: Union[str, ObjectDictionary, TextIO, None] = None,
         upload_eds: bool = False,
     ) -> RemoteNode:
         """Add a remote node to the network.
 
         :param node:
             Can be either an integer representing the node ID, a
-            :class:`canopen.RemoteNode` or :class:`canopen.LocalNode` object.
+            :class:`canopen.RemoteNode` object.
```
acolomb, line 155: I think what is wrong here is the return type annotation.  It's perfectly alright to add an existing `LocalNode` to a network, isn't it?

canopen/network.py
```diff
@@ -335,10 +345,12 @@ def update(self, data: bytes) -> None:
         old_data = self.msg.data
         self.msg.data = new_data
         if hasattr(self._task, "modify_data"):
+            assert self._task is not None  # This will never be None, but mypy needs this
```
acolomb, line 348: Superseded by #534

canopen/network.py
```diff
@@ -335,10 +345,12 @@ def update(self, data: bytes) -> None:
...
             self._task.modify_data(self.msg)
         elif new_data != old_data:
             # Stop and start (will mess up period unfortunately)
-            self._task.stop()
+            if self._task is not None:
+                self._task.stop()
```
acolomb, line 353: Superseded by #534

acolomb (changes requested):
Gradually working through this.  I had some of these fixes made locally as well, so trying to reconcile the bits and factor out into smaller PRs.  I haven't checked all touched files yet, will do so as time permits.

This should be merged with master, especially after #525 is sanctioned and merged.  It will reduce the diff a lot, making it easier to reason about.

sveinse:
I think we concluded on that this PR is not attractive, and that many of its sub-features are superseded by more specific PRs. I have no intentions on merging this PR, so I'm setting it to DRAFT. Yet, as a curious experiment, I updated the branch to reflect the current master to see how far the branch is from master. Perhaps we can get some inspirations for some new smaller PRs from it. Reminder of its goal: Get default mypy to pass without any error.

PR #513, opened 2024-07-10
Annotate Network and check usage of it runtime

sveinse:
With reference to #511 this PR properly annotates all uses of `Network` and adds runtime checks for the usage of `Network` instances. In very many classes `self.network` is `Optional[Network]` with default `None` and thus a runtime check is needed for it being set properly.

### DIFF
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -2,7 +2,10 @@ import struct
 import logging
 import threading
 import time
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from canopen.network import Network
 
 # Error code, error register, vendor specific data
 EMCY_STRUCT = struct.Struct("<HB5s")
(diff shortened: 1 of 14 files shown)


erlend-aasland:
I'm not sure if sprinkling the code with a lot of `if`s is the best way forward. Some of these changes are in functions/methods that can be assumed to be part of hot code, so it will potentially have a performance impact. Moreover, if it is decided to move forward with this PR, the exception messages should be rewritten to provide the user with a solution to the problem. Something like _"This node is not associated with a network; did you forget to add it using `add_node()`?"_ or similar could be an improvement.

acolomb:
I agree that all these conditionals and RuntimeErrors are a bit distracting and verbose.  Adding an assertion seems less intrusive and can be switched off to save runtime costs.

Other than that, the most basic question is, why should the `network` attribute be Optional?  We could require it wherever possible, especially on classes that are not really a part of the public API.  An alternative is to initialize to a dummy implementation instance that satisfies the required attributes.  Haven't really investigated how involved that would be.  Maybe a Protocol (PEP 544) could be useful for this purpose?

acolomb:
Superseded by #525.

PR #521, opened 2024-07-12
Docs: remove unused Sphinx config 'html_static_path'

erlend-aasland:


acolomb:
Does it hurt?  I carried it over while modernizing the config file.  There are certainly no files in there right now, but it does indicate where we should put them once required.

erlend-aasland:
> Does it hurt?

It produces a warning when building the docs. Other than that, no.

acolomb (approved):
We can always go back in history should we ever need it again...

Issue #328, opened 2022-09-20
LocalNode not emitting TPDOs


nick-wlink:
Great library. Generally it has been nice to use, but I'm having some trouble with PDOs.

I would like to use the canopen library to simulate a number of slave devices. Configuration of the slaves will be read from an eds/dcf file and thereafter the bus master may make further configuration changes to the slave devices PDOs. In the final application the bus master will be a separate physical device (_not_ the canopen library).

To begin I am trying to set up a single slave and get it to transmit a PDO. For convenience I am using the canopen library to also simulate my bus master. I have two python repls open. Top left is the bus master, bottom left is the slave and right is a dump of the virtual CAN bus they are communicating over.

My expectation was that a slave with event-driven asynchronous TPDOs configured would start transmitting the TPDO as soon as it was put into 'OPERATIONAL' mode. I'm new to CANOpen, so I'm not sure if this expectation is correct. But this is the case with an actual CANOpen sensor (encoder) that I've been experimenting with. You can see in the screen capture that I've tried to configure the slave PDO (most of the config and mappings are already done in the DCF file) and put it into operational mode, but I don't get any PDOs in the candump.

![log](https://user-images.githubusercontent.com/113959584/191161662-8df7dd8c-d824-457e-91aa-328228039711.png)

I notice that there is a `tpdo.start(interval)` function in the API, but I don't quite understand how this works. Is this supposed to be called on the device that is emitting the TPDO (the slave device in my case)? How do I use this if I don't know in advance which PDOs the master will configure? Is there some kind of callback that I should invoke it from when the slave is put into operational mode?

Also, when I call the `start()` function I am getting an error:
```
>>> node.tpdo[1].start(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nick/.local/lib/python3.10/site-packages/canopen/pdo/base.py", line 498, in start
    self._task = self.pdo_node.network.send_periodic(
  File "/home/nick/.local/lib/python3.10/site-packages/canopen/network.py", line 233, in send_periodic
    return PeriodicMessageTask(can_id, data, period, self.bus, remote)
  File "/home/nick/.local/lib/python3.10/site-packages/canopen/network.py", line 311, in __init__
    self.msg = can.Message(is_extended_id=can_id > 0x7FF,
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

windelbouwman:
[...]

PR #524, opened 2024-07-29
LocalNode improvements for TPDOs

grantweiss-RaymondCorp:
Local Node improvements:
- Implement sending of TPDOs on transition to NMT operational state. #328 
- Allow runtime (pre-op) configuration of TPDO COB-ID

### DIFF
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -34,7 +34,7 @@ class LocalNode(BaseNode):
         self.add_write_callback(self.nmt.on_write)
         self.emcy = EmcyProducer(0x80 + self.id)
 
-        self.nmt.add_state_change_callback(self.nmt_state_changed)
+        self.nmt.add_state_change_callback(self._nmt_state_changed)
 
     def associate_network(self, network):
         self.network = network
@@ -130,7 +130,7 @@ class LocalNode(BaseNode):
             obj = obj[subindex]
         return obj
 
-    def nmt_state_changed(self, old_state, new_state):
+    def _nmt_state_changed(self, old_state, new_state):
         if new_state == "OPERATIONAL":
             for i, pdo in self.tpdo.map.items():
                 if pdo.enabled:
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -34,6 +34,8 @@ class LocalNode(BaseNode):
         self.add_write_callback(self.nmt.on_write)
         self.emcy = EmcyProducer(0x80 + self.id)
 
+        self.nmt.add_state_change_callback(self.nmt_state_changed)
+
     def associate_network(self, network):
         self.network = network
         self.sdo.network = network
@@ -127,3 +129,19 @@ class LocalNode(BaseNode):
                 raise SdoAbortedError(0x06090011)
             obj = obj[subindex]
         return obj
+
+    def nmt_state_changed(self, old_state, new_state):
+        if new_state == "OPERATIONAL":
+            for i, pdo in self.tpdo.map.items():
+                if pdo.enabled:
+                    try:
+                        pdo.start()
+                        logger.info(f"Successfully started TPDO {i}")
+                    except ValueError:
+                        logger.warning(f"Failed to start TPDO {i} due to missing period")
+                    except Exception as e:
+                        logger.error(f"Unknown error starting TPDO {i}: {str(e)}")
+                else:
+                    logger.info(f"TPDO {i} not enabled")
+        elif old_state == "OPERATIONAL":
+            self.tpdo.stop()
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -35,6 +35,7 @@ class LocalNode(BaseNode):
         self.emcy = EmcyProducer(0x80 + self.id)
 
         self.nmt.add_state_change_callback(self._nmt_state_changed)
+        self.add_write_callback(self._tpdo_configuration_write)
 
     def associate_network(self, network):
         self.network = network
@@ -145,3 +146,23 @@ class LocalNode(BaseNode):
                     logger.info(f"TPDO {i} not enabled")
         elif old_state == "OPERATIONAL":
             self.tpdo.stop()
+
+    def _tpdo_configuration_write(self, index, subindex, od, data):
+        if 0x1800 <= index <= 0x19FF:
+            # Only allowed to edit pdo configuration in pre-op
+            if self.nmt.state != "PRE-OPERATIONAL":
+                logger.warning("Tried to configure tpdo when not in pre-op")
+                return
+            
+            if subindex == 0x01:
+                if len(data) == 4:
+                    tpdoNum = (index - 0x1800) + 1
+                    if tpdoNum in self.tpdo.map.keys():
+                        PDO_NOT_VALID = 1 << 31
+                        RTR_NOT_ALLOWED = 1 << 30
+
+                        cob_id = int.from_bytes(data, 'little') & 0x7FF
+
+                        self.tpdo.map[tpdoNum].cob_id = cob_id & 0x1FFFFFFF
+                        self.tpdo.map[tpdoNum].enabled = cob_id & PDO_NOT_VALID == 0
+                        self.tpdo.map[tpdoNum].rtr_allowed = cob_id & RTR_NOT_ALLOWED == 0
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -35,7 +35,6 @@ class LocalNode(BaseNode):
         self.emcy = EmcyProducer(0x80 + self.id)
 
         self.nmt.add_state_change_callback(self._nmt_state_changed)
-        self.add_write_callback(self._tpdo_configuration_write)
 
     def associate_network(self, network):
         self.network = network
@@ -118,6 +117,15 @@ class LocalNode(BaseNode):
         self.data_store.setdefault(index, {})
         self.data_store[index][subindex] = bytes(data)
 
+        if 0x1800 <= index <= 0x19FF:
+            # TPDO Communication parameter changed
+            tpdoNum = (index - 0x1800) + 1
+            self._tpdo_configuration_write(tpdoNum)
+        elif 0x1A00 <= index <= 0x1BFF:
+            # TPDO Mapping parameter changed
+            tpdoNum = (index - 0x1A00) + 1
+            self._tpdo_configuration_write(tpdoNum)
+
     def _find_object(self, index, subindex):
         if index not in self.object_dictionary:
             # Index does not exist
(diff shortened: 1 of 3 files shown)


grantweiss-RaymondCorp:
I can also pull in 2305314b199f52037c1c9d3a81efbf888e8d5921 if someone wants to confirm this is correct canopen behavior.

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
             obj = obj[subindex]
         return obj
+
+    def _nmt_state_changed(self, old_state, new_state):
+        if new_state == "OPERATIONAL":
+            for i, pdo in self.tpdo.map.items():
+                if pdo.enabled:
+                    try:
+                        pdo.start()
+                        logger.info(f"Successfully started TPDO {i}")
```
acolomb, line 140: This should use `pdo.name` to disambiguate between nodes.

And please stick to the formatting style used in the rest of the code.  For logger calls, we use `%`-style and pass the substitutions as arguments (where formatters see them as structured data).  For other uses, f-strings are preferred.

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
+                    except ValueError:
+                        logger.warning(f"Failed to start TPDO {i} due to missing period")
+                    except Exception as e:
+                        logger.error(f"Unknown error starting TPDO {i}: {str(e)}")
```
acolomb, line 144: Wouldn't `logger.exception()` be more helpful?

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
+                else:
+                    logger.info(f"TPDO {i} not enabled")
+        elif old_state == "OPERATIONAL":
```
acolomb, line 147: Only when coming *from* OPERATIONAL?  I'd say all states beside that one forbid PDO transmission, thus we can unconditionally disable them when not in OPERATIONAL.  Stopping if already stopped should work alright, if not then please fix that.

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
+            self.tpdo.stop()
+
+    def _tpdo_configuration_write(self, index, subindex, od, data):
+        if 0x1800 <= index <= 0x19FF:
```
acolomb, line 151: This could be negated (with a comment) to `return` early and reduce the nested indentation levels.

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
+            # Only allowed to edit pdo configuration in pre-op
+            if self.nmt.state != "PRE-OPERATIONAL":
```
acolomb, line 153: That's not true in general, but we could make it a requirement for this library's implementation.  Please indicate clearly if it should be the latter.

canopen/node/local.py
```diff
@@ -127,3 +130,39 @@ def _find_object(self, index, subindex):
...
+                    tpdoNum = (index - 0x1800) + 1
+                    if tpdoNum in self.tpdo.map.keys():
+                        PDO_NOT_VALID = 1 << 31
+                        RTR_NOT_ALLOWED = 1 << 30
+
+                        cob_id = int.from_bytes(data, 'little') & 0x7FF
+
+                        self.tpdo.map[tpdoNum].cob_id = cob_id & 0x1FFFFFFF
+                        self.tpdo.map[tpdoNum].enabled = cob_id & PDO_NOT_VALID == 0
+                        self.tpdo.map[tpdoNum].rtr_allowed = cob_id & RTR_NOT_ALLOWED == 0
```
acolomb, line 168: I think this whole approach should rather try to manipulate our `PdoMap` members instead of adding separate parsing code here.  Adjust the related objects via SDO, then call `pdo.read(from_od=True)`.  Haven't checked in detail whether that works, but it's the reason for having that `from_od` parameter.

Nit-pick: Please fix up your editor to emit a trailing newline at the end of the file, or add one manually.

acolomb (changes requested):
Nice addition, but I think the implementation could be simplified by re-using more existing code.  And some style nit-picks for consistency, see below.

canopen/node/local.py
grantweiss-RaymondCorp: I almost had it, depending on what level we want to support (variable vs dynamic) it can be changed in pre-op or operational. [https://www.can-cia.org/can-knowledge/pdo-protocol-1](https://www.can-cia.org/can-knowledge/pdo-protocol-1#:~:text=in%20CANopen%C2%A0FD.-,PDO%20mapping,-CANopen%20FD%20distinguishes). I expanded it to allow while in operational. This still leaves out the PDO mapping procedure...

grantweiss-RaymondCorp:
@acolomb Thank you for looking over everything, I think I addressed it all.

canopen/node/local.py
acolomb: The only restriction I found in the standard is that subindex 03 "shall not be changed while the PDO exists (bit 31 ofsub-index 01h is set to 0b)."  So as long as the library's implementation does support dynamic reconfiguration (which I think it does?) we have no reason to restrict its usage IMHO.

The mentioned restriction as well as the PDO mapping procedure should be handled in the PDO code, but that's definitely not a blocker for this PR.

canopen/node/local.py
```diff
@@ -115,6 +117,15 @@ def set_data(
...
         self.data_store[index][subindex] = bytes(data)
 
+        if 0x1800 <= index <= 0x19FF:
+            # TPDO Communication parameter changed
+            tpdoNum = (index - 0x1800) + 1
+            self._tpdo_configuration_write(tpdoNum)
+        elif 0x1A00 <= index <= 0x1BFF:
+            # TPDO Mapping parameter changed
+            tpdoNum = (index - 0x1A00) + 1
+            self._tpdo_configuration_write(tpdoNum)
```
acolomb, line 127: This could be shortened a bit because the ranges are adjacent.  You only need one range check and one modulo to find the TPDO number.  But feel free to ignore this if you think it's not an improvement.

canopen/node/local.py
```diff
@@ -34,6 +34,8 @@ def __init__(
         self.add_write_callback(self.nmt.on_write)
         self.emcy = EmcyProducer(0x80 + self.id)
 
+        self.nmt.add_state_change_callback(self._nmt_state_changed)
```
acolomb, line 37: I'm wondering whether this should be made opt-in.  There is certainly existing code using `LocalNode` which already manages the TPDO transmissions externally, because the library didn't use to do it.  With this change, these users could end up with duplicated transmissions, or at least things stepping on each others' toes.

canopen/node/local.py
```diff
@@ -127,3 +138,32 @@ def _find_object(self, index, subindex):
                 raise SdoAbortedError(0x06090011)
             obj = obj[subindex]
         return obj
+
+    def _nmt_state_changed(self, old_state, new_state):
+        if new_state == "OPERATIONAL":
+            for pdo in self.tpdo.map.values():
+                if pdo.enabled:
```
acolomb, line 145: Maybe a matter of taste, but I would prefer an inverted check and a `continue` here.  Makes for less indentation and doesn't stretch the else branch possibly out of sight.

canopen/node/local.py
```diff
@@ -127,3 +138,32 @@ def _find_object(self, index, subindex):
...
+                    try:
+                        pdo.start()
+                        logger.info("Successfully started %s", pdo.name)
+                    except ValueError:
+                        logger.warning("Failed to start %s due to missing period", pdo.name)
+                    except Exception:
+                        logger.exception("Unknown error starting %s", pdo.name)
+                else:
+                    logger.info("%s not enabled", pdo.name)
```
acolomb, line 154: Now this logging is probably too much.  Worst case, it will be emitted for all 512 PDOs.  That's too much even on DEBUG level.  I'd just drop this log call.

canopen/node/local.py
```diff
@@ -127,3 +138,32 @@ def _find_object(self, index, subindex):
...
+
+        # Only allowed to edit pdo configuration in pre-op or operational
+        if self.nmt.state not in ("PRE-OPERATIONAL", "OPERATIONAL"):
+            logger.warning("Tried to configure %s when not in pre-op or operational", pdo.name)
+            return
+        
+        try:
+            pdo.read(from_od=True)
+        except:
+            pass
```
acolomb, line 169: Why would we want to silently swallow any and all exceptions here?

test/test_local.py
```diff
@@ -200,6 +200,37 @@ def test_save(self):
...
+        # Disable tpdo 1
+        self.local_node.tpdo[1].enabled = False
+        self.local_node.tpdo[1].cob_id = 0
+        self.local_node.tpdo[1].period = 0.5 # manually assign a period
+
+        self.local_node.nmt.state = 'INITIALISING'
+        self.local_node.nmt.state = 'PRE-OPERATIONAL'
+
+        # Attempt to re-enable tpdo 1 via sdo writing
+        PDO_NOT_VALID = 1 << 31
```
acolomb, line 222: Possible to re-use `canopen.pdo.base.PDO_NOT_VALID`?

test/test_local.py
```diff
@@ -200,6 +200,37 @@ def test_save(self):
...
+        self.local_node.nmt.state = 'PRE-OPERATIONAL'
+        self.local_node.nmt.state = 'OPERATIONAL'
+
+        self.assertNotEqual(self.local_node.tpdo[1]._task, None)
+
+    def test_config_pdo(self):
+        # Disable tpdo 1
+        self.local_node.tpdo[1].enabled = False
+        self.local_node.tpdo[1].cob_id = 0
+        self.local_node.tpdo[1].period = 0.5 # manually assign a period
```
acolomb, line 216: I don't understand that comment.  Just above, you do the same without a comment.  Is this intended to test mix & match between direct assignment and SDO based changes?

Isn't it using the value from the inhibit time object anyway?

test/test_local.py
```diff
@@ -200,6 +200,37 @@ def test_save(self):
...
+
+        self.local_node.nmt.state = 'INITIALISING'
+        self.local_node.nmt.state = 'PRE-OPERATIONAL'
+
+        # Attempt to re-enable tpdo 1 via sdo writing
+        PDO_NOT_VALID = 1 << 31
+        odDefaultVal = self.local_node.object_dictionary["Transmit PDO 0 communication parameters.COB-ID use by TPDO 1"].default
```
acolomb, line 223: I'd use the numeric index here, as it's less fragile and shorter.  There's already a typo in that description, and fixing it would break the test :wink:

test/test_local.py
```diff
@@ -200,6 +200,37 @@ def test_save(self):
...
+        enabledCobId = odDefaultVal & ~PDO_NOT_VALID # Ensure invalid bit is not set
+
+        self.remote_node.sdo["Transmit PDO 0 communication parameters.COB-ID use by TPDO 1"].raw = enabledCobId
+
+        # Transition to operational
```
acolomb, line 228: These two comments don't really help.  The test code should be simple enough to not need an explanation.

acolomb (changes requested):

test/test_local.py
grantweiss-RaymondCorp: It doesn't use inhibit time currently. I have a different branch for this, I wasn't sure if I should try to pull it in with this stuff.  https://github.com/grantweiss-RaymondCorp/canopen/commit/0a9f50da74e1e3ed3c9a18a13e747a6a1a484772

canopen/node/local.py
grantweiss-RaymondCorp: What would be the preferred way to achieve this? A keyword argument in the `LocalNode` constructor?

acolomb: Yes, I think an additional argument `auto_tpdo: bool = False` would be fine, with appropriate documentation (docstring) of course.

And while we're at it, I think the `_nmt_state_changed` name does not convey what the function does.  It should rather be named `_tpdo_auto_start_stop`, then this line will clearly express what the connection is (NMT state --> TPDO start).

test/test_local.py
acolomb: The approach of falling back to `event_timer` looks reasonable.  But I agree we don't have to do it all in this PR, but can iterate with more improvements / features once this is merged.

canopen/node/local.py
acolomb: And this will generate warnings for any TPDO not configured as periodic, right?  Those should probably rather be skipped silently.  Have a look at the `is_periodic` property, which could be used for filtering.

canopen/node/local.py
acolomb: > Would the `_tpdo_configuration_write` call inside `set_data` also be opt-in?

No, I think if someone writes to the local node's objects, it is quite expected to affect the PDO settings.  Actually even for RPDOs.  Of course our processing of these parameters is quite rudimentary now and not really standard compliant, but that can be improved step by step.

PR #525, opened 2024-08-07
Add a dummy object to designate "uninitialized" networks (fixes #511)

### LINKED ISSUE #511: Network instance is optional in many classes, which requires lot of checking

acolomb:
This class can replace the dummy "None" value for attribute initializations, which can then be properly typed as Network to avoid static type checking errors.

Switch to that new initialization in the NmtBase class as an example.

This has the benefit of not needing `self.network is not None` checks at run-time wherever a method or attribute access is used, but still satisfies static type checking.  When hitting such a code path at run-time, of course it will lead to an exception because the attributes required in the Network methods are not set.  But that is a case of wrong API usage (accessing a network without associating it first), which a static checker cannot detect reliably.  The dummy class could be modified to provide a better exception message if desired, but this is just a minimal proof of concept so far.

This is an alternative approach to #513.  If the concept looks acceptable, it will be extended to other places with the same type checker errors.

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -279,7 +279,7 @@ class Network(MutableMapping):
         return len(self.nodes)
 
 
-class _DummyNetwork(Network):
+class _UninitializedNetwork(Network):
     """Empty network implementation as a placeholder before actual initialization."""
 
     def __init__(self, bus: Optional[can.BusABC] = None):
@@ -288,6 +288,10 @@ class _DummyNetwork(Network):
     def __getattribute__(self, name):
         raise RuntimeError("No actual Network object was assigned, "
                            "try associating to a real network first.")
+
+
+#: Singleton instance
+_UNINITIALIZED_NETWORK = _UninitializedNetwork()
 
 
 class PeriodicMessageTask:
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from collections.abc import MutableMapping
 import logging
 import threading
-from typing import Callable, Dict, Iterator, List, Optional, Union
+from typing import Callable, Dict, Final, Iterator, List, Optional, Union
 
 import can
 from can import Listener
@@ -291,7 +291,7 @@ class _UninitializedNetwork(Network):
 
 
 #: Singleton instance
-_UNINITIALIZED_NETWORK = _UninitializedNetwork()
+_UNINITIALIZED_NETWORK: Final[Network] = _UninitializedNetwork()
 
 
 class PeriodicMessageTask:
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -285,6 +285,10 @@ class _DummyNetwork(Network):
     def __init__(self, bus: Optional[can.BusABC] = None):
         """Do not initialize attributes, by skipping the parent constructor."""
 
+    def __getattribute__(self, name):
+        raise RuntimeError("No actual Network object was assigned, "
+                           "try associating to a real network first.")
+
 
 class PeriodicMessageTask:
     """
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -279,6 +279,13 @@ class Network(MutableMapping):
         return len(self.nodes)
 
 
+class _DummyNetwork(Network):
+    """Empty network implementation as a placeholder before actual initialization."""
+
+    def __init__(self, bus: Optional[can.BusABC] = None):
+        """Do not initialize attributes, by skipping the parent constructor."""
+
+
 class PeriodicMessageTask:
     """
     Task object to transmit a message periodically using python-can's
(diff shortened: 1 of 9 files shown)


erlend-aasland:
> @erlend-aasland What do you think of this approach, before I start fleshing it out in more detail?

I'm not totally convinced, but I like it better than sprinkling the code with conditionals.

canopen/network.py
```diff
@@ -279,6 +279,17 @@ def __len__(self) -> int:
         return len(self.nodes)
 
 
+class _DummyNetwork(Network):
```
erlend-aasland, line 282: I'm not sure about the naming. I'd prefer if the naming more closely reflected it's purpose:

- `_NetworkPlaceholder`
- `_UninitializedNetwork`
- `_EmptyNetwork`

canopen/node/local.py
```diff
@@ -44,15 +47,15 @@ def associate_network(self, network):
...
-        self.tpdo.network = None
-        self.rpdo.network = None
-        self.nmt.network = None
-        self.emcy.network = None
+        self.network = canopen.network._DummyNetwork()
+        self.sdo.network = self.network
+        self.tpdo.network = self.network
+        self.rpdo.network = self.network
+        self.nmt.network = self.network
+        self.emcy.network = self.network
```
erlend-aasland, line 58: Consider to create a singleton placeholder in `canopen/network.py`, import the singleton here, instead of creating new objects all the time.

Ditto for `canopen/remote.py`, `canopen/pdo/base.py`, and `canopen/sdo/base.py`.

Then you can do something like this:

```suggestion
        self.network = _NetworkPlaceholder
        self.sdo.network = _NetworkPlaceholder
        self.tpdo.network = _NetworkPlaceholder
        self.rpdo.network = _NetworkPlaceholder
        self.nmt.network = _NetworkPlaceholder
        self.emcy.network = _NetworkPlaceholder
```

acolomb: What to name the singleton then?  I'm leaning towards: `_EmptyNetwork = _UninitializedNetwork()`.

erlend-aasland: How about one of these?

- `_UNINITIALIZED_NETWORK`
- `_EMPTY_NETWORK`
- `_NETWORK_PLACEHOLDER`

canopen/network.py
```diff
@@ -279,6 +279,21 @@ def __len__(self) -> int:
...
+    def __init__(self, bus: Optional[can.BusABC] = None):
+        """Do not initialize attributes, by skipping the parent constructor."""
+
+    def __getattribute__(self, name):
+        raise RuntimeError("No actual Network object was assigned, "
+                           "try associating to a real network first.")
+
+
+#: Singleton instance
+_UNINITIALIZED_NETWORK = _UninitializedNetwork()
```
erlend-aasland, line 294:
```suggestion
_UNINITIALIZED_NETWORK: Final[Network] = _UninitializedNetwork()
```

Requires `from typing import Final`.

erlend-aasland (approved):
I like this!

erlend-aasland:
(You need to amend the PR title to reflect the name changes.)

acolomb:
@sveinse Are you satisfied with this solution instead of #513?

acolomb:
Gentle ping @sveinse, I'm still waiting on your opinion.  If you're currently not able to invest much time, please say so.  Two contributors with a preference to accept this PR is enough for me, just didn't want to pass you over.

sveinse (approved):
@acolomb thank you for the ping. I think this solution is the best of the non-perfect solutions. The alternative would be to pepper with `network: Optional[Network] = None` and having to deal with the `None` to instruct the linter what it should be. Using this object shows intent and it will fail if it's tried used without initialization. PR Approved.

Sorry for not responding, I'm in a terrible busy phase @work so I have limited time. Thanks for the involvement thou.

PR #517, opened 2024-07-11
Tests: improve coverage of NodeScanner

erlend-aasland:

PR #526, opened 2024-08-09
Make node scanner tests more deterministic

erlend-aasland:
It seems sending and receiving on the same virtual bus[^1] is flaky. My initial assumption that the flakiness was caused by unordered messages was thus incorrect[^2].

After fiddling around with this a couple of evenings, I get stable CI runs by using two separate virtual `can.Bus` instances; one for the object we're testing ("TX bus"), and one for validating what's being sent ("RX bus"). We might want to follow this strategy for other parts of the test suite as well. I vaguely remember earlier issues when using `receive_own_messages=True` and repeated `bus.recv` polling.

[^1]: `can.Bus(interface="virtual", receive_own_messages=True)`
[^2]: ... which is pretty obvious when you browse through the `python-can` source code

### DIFF
--- a/test/test_network.py
+++ b/test/test_network.py
@@ -321,8 +321,13 @@ class TestScanner(unittest.TestCase):
             self.scanner.search()
 
     def test_scanner_search(self):
-        bus = can.Bus(interface="virtual", receive_own_messages=True)
-        net = canopen.Network(bus)
+        rxbus = can.Bus(interface="virtual")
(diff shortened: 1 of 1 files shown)


erlend-aasland:
FTR, this fixes up the tests added with commit d1c28e51b93c7d45fce40187672e7ab7a0616cac (gh-517).

acolomb (approved):

PR #535, opened 2024-08-14
Tests: force disconnect from virtual busses on tearDown() (fixes #530)

### LINKED ISSUE #530: Reduce network shutdown latency
Currently, when we shut down a network, we (among other things) call `notifier.stop()`, if there's a notifier. The `can.Notifier` has a pretty long timeout by default; is this needed when we call `network.shutdown()`?

erlend-aasland:


erlend-aasland:
Closed in favour of #536.

PR #536, opened 2024-08-14
Make Network.notifier timeouts customizable (fixes #530)

### LINKED ISSUE #530: Reduce network shutdown latency

acolomb:
Move the hard-coded 1 second maximum cycle duration passed during
`can.Notifier` construction to a class variable `NOTIFIER_CYCLE`.
Explicitly specify the timeout to stop the notifier in another class
variable `NOTIFIER_SHUTDOWN_TIMEOUT`, matching the default used in
python-can.  These can be overridden per instance if needed.

Do not wait for notifier thread shutdown during tests by explicitly setting the `NOTIFIER_SHUTDOWN_TIMEOUT` to zero in all `Network` instances created for unit tests.

This reduces total test running time from ~30 to ~9 seconds when run locally.

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -25,8 +25,8 @@ Callback = Callable[[int, bytearray, float], None]
 class Network(MutableMapping):
     """Representation of one CAN bus containing one or more nodes."""
 
-    NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration
-    NOTIFIER_SHUTDOWN_TIMEOUT: float = 5.0  #: Maximum waiting time to stop notifiers
+    NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration.
+    NOTIFIER_SHUTDOWN_TIMEOUT: float = 5.0  #: Maximum waiting time to stop notifiers.
 
     def __init__(self, bus: Optional[can.BusABC] = None):
         """
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -25,6 +25,9 @@ Callback = Callable[[int, bytearray, float], None]
 class Network(MutableMapping):
     """Representation of one CAN bus containing one or more nodes."""
 
+    NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration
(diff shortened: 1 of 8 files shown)


acolomb:
@erlend-aasland this should replace #535.  Note that the actual time for the tests alone varies between 9 and 47 seconds on my machine.  It seems the NMT tests timings are not very reproducible mainly.  That also explains why this PR is a bit quicker for one check (34 s), but slower for the other (50 s) than the alternative #535.

erlend-aasland:
You beat me; I was working on a similar PR based on your suggestion 😄 This looks nice!

canopen/network.py
```diff
@@ -25,6 +25,9 @@
 class Network(MutableMapping):
     """Representation of one CAN bus containing one or more nodes."""
 
+    NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration
+    NOTIFIER_SHUTDOWN_TIMEOUT: float = 5.0  #: Maximum waiting time to stop notifiers
```
erlend-aasland, line 29: Make sure the docs are rendered with correct punctuation:

```suggestion
    NOTIFIER_CYCLE: float = 1.0  #: Maximum waiting time for one notifier iteration.
    NOTIFIER_SHUTDOWN_TIMEOUT: float = 5.0  #: Maximum waiting time to stop notifiers.
```

erlend-aasland (approved):

canopen/network.py
acolomb: Hm, I thought that was part of the Sphinx rendering process?  Might be confusing it with Doxygen, IIRC there the full-stops get added automatically.  Pretty annoying for me to see them in the code, actually.

erlend-aasland: Nope, it's never been part of Sphinx AFAIK. Perhaps Doxygen, but I haven't used that in ages, so I don't remember how it behaves.

Correct punctuation in comments is a small prize to pay for nicely rendered docs, no?

PR #538, opened 2024-08-16
Skip the "highest subindex" entry when iterating over SdoRecord

acolomb:
The count entry itself is not part of the data (according to CiA 301), thus should not be yielded from an iterator.  That matches the behavior of SdoArray, which also yields only the array contents.

Note that the basis of returned record sub-objects is still the subset described by the OD, which might be smaller than the actual entries accessible on the node, and less than indicated by the record's subindex 0.  Thus the count (and iteration set) is reduced by one element only if the subindex 0 was actually part of this subset.

This supersedes PR #430, with the alternative approach justified in the [last comment there](https://github.com/christiansandberg/canopen/pull/430#issuecomment-2292416333).

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -5,9 +5,17 @@ name: Python package
 
 on:
   push:
-    branches: [ "master" ]
+    branches:
+    - 'master'
+    paths-ignore:
+    - 'README.rst'
+    - 'LICENSE.txt'
   pull_request:
-    branches: [ "master" ]
+    branches:
+    - 'master'
+    paths-ignore:
+    - 'README.rst'
+    - 'LICENSE.txt'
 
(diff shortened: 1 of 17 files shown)


erlend-aasland (commented):
Consider to also update `doc/sdo.rst` with a sentence or two similar to the added code comments.

Also, it would be nice with test for this, and by that I mean preferably to add those tests first (in a separate PR), followed by the behavioural changes (this PR) coupled with amendments to the (presumably broken) tests.

acolomb:
All comments addressed, see #539.

erlend-aasland (approved):
Thanks! I can confirm that the added tests cover the affected lines, and that `test_record_iter_length` expectedly fails if applied on `master`.

erlend-aasland (approved):
🚀

PR #542, opened 2024-09-02
TPDO mappings may be read-only

### LINKED ISSUE #541: `PdoMap.save()` does not handle cases where 1A00h through 1A03h are read-only
I've got a sensor device where the TPDO mapping parameters are read-only. AFAICS from the CiA 301 standard, TPDO mappings _can_ be read-only, so it seems to me this case should be supported. A fix should be trivial; I'll propose a patch shortly.

erlend-aasland:
Resolves #541

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -427,15 +427,19 @@ class PdoMap:
         for var in self.map:
             logger.info("Writing %s (0x%04X:%02X, %d bits) to PDO map",
                         var.name, var.index, var.subindex, var.length)
-            if getattr(self.pdo_node.node, "curtis_hack", False):
-                # Curtis HACK: mixed up field order
-                self.map_array[subindex].raw = (var.index |
-                                                var.subindex << 16 |
-                                                var.length << 24)
-            else:
-                self.map_array[subindex].raw = (var.index << 16 |
-                                                var.subindex << 8 |
-                                                var.length)
+            try:
+                if getattr(self.pdo_node.node, "curtis_hack", False):
+                    # Curtis HACK: mixed up field order
+                    self.map_array[subindex].raw = (var.index |
+                                                    var.subindex << 16 |
+                                                    var.length << 24)
+                else:
+                    self.map_array[subindex].raw = (var.index << 16 |
+                                                    var.subindex << 8 |
+                                                    var.length)
+            except SdoAbortedError as e:
+                if e.code != 0x06010002:
+                    raise
             subindex += 1
         try:
             self.map_array[0].raw = len(self.map)
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -423,24 +423,20 @@ class PdoMap:
             # mappings for an invalid object 0x0000:00 to overwrite any
             # excess entries with all-zeros.
             self._fill_map(self.map_array[0].raw)
-        subindex = 1
-        for var in self.map:
+        for subindex, var in enumerate(self.map, start=1):
+            if not var.od.writable:
+                continue
             logger.info("Writing %s (0x%04X:%02X, %d bits) to PDO map",
                         var.name, var.index, var.subindex, var.length)
-            try:
-                if getattr(self.pdo_node.node, "curtis_hack", False):
-                    # Curtis HACK: mixed up field order
-                    self.map_array[subindex].raw = (var.index |
-                                                    var.subindex << 16 |
-                                                    var.length << 24)
-                else:
-                    self.map_array[subindex].raw = (var.index << 16 |
-                                                    var.subindex << 8 |
-                                                    var.length)
-            except SdoAbortedError as e:
-                if e.code != 0x06010002:
-                    raise
-            subindex += 1
+            if getattr(self.pdo_node.node, "curtis_hack", False):
+                # Curtis HACK: mixed up field order
+                self.map_array[subindex].raw = (var.index |
+                                                var.subindex << 16 |
+                                                var.length << 24)
+            else:
+                self.map_array[subindex].raw = (var.index << 16 |
+                                                var.subindex << 8 |
+                                                var.length)
         try:
             self.map_array[0].raw = len(self.map)
         except SdoAbortedError as e:


canopen/pdo/base.py
```diff
@@ -424,18 +424,23 @@ def save(self) -> None:
             # excess entries with all-zeros.
             self._fill_map(self.map_array[0].raw)
         subindex = 1
+
```
erlend-aasland, line 427:
```suggestion
```

erlend-aasland:
FTR, the `.save()` method could benefit from some refactoring. Nothing critical, though.

canopen/pdo/base.py
```diff
@@ -427,15 +427,19 @@ def save(self) -> None:
         for var in self.map:
```
erlend-aasland, line 427: If you want to save some lines and make it slightly more pythonic:

```diff
diff --git a/canopen/pdo/base.py b/canopen/pdo/base.py
index 04bc28b..8c52644 100644
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -423,8 +423,7 @@ class PdoMap:
             # mappings for an invalid object 0x0000:00 to overwrite any
             # excess entries with all-zeros.
             self._fill_map(self.map_array[0].raw)
-        subindex = 1
-        for var in self.map:
+        for subindex, var in enumerate(self.map, start=1):
             logger.info("Writing %s (0x%04X:%02X, %d bits) to PDO map",
                         var.name, var.index, var.subindex, var.length)
             try:
@@ -440,7 +439,6 @@ class PdoMap:
             except SdoAbortedError as e:
                 if e.code != 0x06010002:
                     raise
-            subindex += 1
         try:
             self.map_array[0].raw = len(self.map)
         except SdoAbortedError as e:
```

acolomb:
Hm.  I'm not sure ignoring the errors like this is the right thing to do.  After all, the mappings will then be out of sync.  If the device doesn't support dynamic re-mapping, then calling `.save()` isn't the right thing to do, but rather `.read()`.  A better approach to catch this in advance would be to look at the `AccessType` attribute (via `.writable` property) in the Object Dictionary before even trying to save, or at least while saving, before attempting the actual SDO access.  If the OD is wrong on that account, then fix your EDS :-)

> FTR, the `.save()` method could benefit from some refactoring. Nothing critical, though.

What do you have in mind exactly?  Sure, it ain't pretty, but I wouldn't want to introduce lots of internal methods without a clear benefit (like reusability).

erlend-aasland:
> Hm. I'm not sure ignoring the errors like this is the right thing to do. After all, the mappings will then be out of sync. If the device doesn't support dynamic re-mapping, then calling `.save()` isn't the right thing to do, but rather `.read()`. A better approach to catch this in advance would be to look at the `AccessType` attribute (via `.writable` property) in the Object Dictionary before even trying to save, or at least while saving, before attempting the actual SDO access. If the OD is wrong on that account, then fix your EDS :-)

Thanks for your insight; yes, using the the OD is a better approach. For my case, I see that the supplied EDS file has subindices of 1A00 marked as read-only. I'll rewrite the PR asap.

> What do you have in mind exactly? Sure, it ain't pretty, but I wouldn't want to introduce lots of internal methods without a clear benefit (like reusability).

Nothing in mind; just a drive-by observation :)

acolomb:
Just to be clear about the motivation for this change. You're trying to save the PDO settings because you changed something besides the mapping parameters. The communication parameters are writable, but mappings are not. Therefore you want to save only the writable parameters.

I wonder why we should do this only for the mappings. If some other setting is read-only, it may benefit from the same treatment. And I don't feel really comfortable with just silently skipping the entries. There should be some logging at least to indicate the operation was only partially carried out.

erlend-aasland:
> Just to be clear about the motivation for this change. You're trying to save the PDO settings because you changed something besides the mapping parameters. The communication parameters are writable, but mappings are not. Therefore you want to save only the writable parameters.

Correct; specifically in my case, I want to disable some TPDOs and modify the SYNC behaviour of others.

> I wonder why we should do this only for the mappings. If some other setting is read-only, it may benefit from the same treatment.

That makes sense.

> And I don't feel really comfortable with just silently skipping the entries. There should be some logging at least to indicate the operation was only partially carried out.

I wondered about this. In this case I don't think it will be too annoying with more, rather than less, logging. I guess INFO would be an appropriate level if the values are not changed, and WARNING if a local configuration change was made and scheduled for download. (Can we even detect this, though? I did not look yet.)

> This still leaves the problem that the read-only settings are then likely out of sync, because if we couldn't write the value, we don't read it back either. So in this case, a read operation is in order after the partial saving, to sync the read-only entries. Or they must be initialized correctly from the OD first with the recorded constant value.

How about raising an exception if the user tries to write to a read-only variable, so we'd prevent this before we reach `save()`?

```python
node.tpdo.read()
node.tpdo[1].trans_type = 0  # BOOM! Trying to set a read-only field
node.tpdo.save()
```

acolomb:
> I guess INFO would be an appropriate level if the values are not changed, and WARNING if a local configuration change was made and scheduled for download. (Can we even detect this, though? I did not look yet.)

That's precisely the problem. We cannot know without either uploading or downloading the objects. The correct approach IMHO would be to read the settings, modify some, write back and then it's okay for some updates to be skipped.

> How about raising an exception if the user tries to write to a read-only variable, so we'd prevent this before we reach save()?

That's a nice idea, but not always possible. First, assigning the variable happens when the user does it, but also when the read method gets the uploaded value. And then there might be cases where the PDO configuration is done "offline": You know the device's settings and just configure the library to have the same picture, without actually asking the device. That saves a whole bunch of SDO exchanges and wouldn't work well if assigning the attribute is blocked just because the hypothetical SDO write access would fail.

erlend-aasland:
> That's a nice idea, but not always possible. First, assigning the variable happens when the user does it, but also when the read method gets the uploaded value.

We can solve the public API using properties if the OD provides what's needed to determine writability.

acolomb:
That doesn't solve the second issue with the approach though.  I think it complicates things.  Currently some use cases may run into an exception, but we shouldn't prevent other use cases while fixing that.  The only ways to know the library state reflects the device configuration is to either read or write PDO config.  Splitting up the writing into communication and mapping parameters (e.g. via optional boolean `skip_*` parameters or separate methods) may be a solution for your specific case.  I don't see us restricting the API use though, based on some assumptions about a value being "changed".  Downloading (saving) a read-only parameter is simply the wrong thing to do and should be skipped.  And to see whether it was changed from some other (which?) value, you need to do the right thing: upload (read) it.

erlend-aasland:
Perhaps this is best solved as a documentation update. It seems pleasing all use cases is too much hassle for the win.

acolomb:
Logging failed entries instead of raising an SDO exception is still an improvement.  If OD says it's not writable, then log at INFO level and continue with the next entry / block.  That should get noticed during development.

The documentation update should then advise what to do if this type of INFO message appears:

- The device only partially supports changing PDO configuration.  To make sure the library's view of the PDO configuration is in sync with the device, try calling `.read()` first, then modify the entries which were *not* mentioned as being skipped (the others cannot be changed).  Finally, after calling `.save()` all entries should match up.  If the device actually does changing an entry, make sure the Object Dictionary description correctly reflects that write access capability.

Something along those lines...

sveinse:
> How about raising an exception if the user tries to write to a read-only variable, so we'd prevent this before we reach `save()`?
> 
> ```python
> node.tpdo.read()
> node.tpdo[1].trans_type = 0  # BOOM! Trying to set a read-only field
> node.tpdo.save()
> ```

I think it should be an error to update mapping parameters of a RO mapping, regardless if `.read()` were read "offline" from the OD or from the actual device (`read(from_od=False)`). This reduces the risk for getting out of sync with the remote node.

Since canopen have no downloadable "RW bit" for the mapping access, the OD is the single truth to determine if a PDO mapping parameter can be updated or not.

I think the concept of the code changes of this PR are fine. However it will need to check RW-flags on every object it sets, not just the mapping parameter. All sets of the communication parameter objects must be checked too. -- This conditional logic is required to save _some_ objects that are RW, but it will leave the RO ones untouched, which is expected.

> > FTR, the `.save()` method could benefit from some refactoring. Nothing critical, though.
> 
> What do you have in mind exactly? Sure, it ain't pretty, but I wouldn't want to introduce lots of internal methods without a clear benefit (like reusability).

FTR When we delve into the asyncio adoption, we will have to refactor this and the `.load()` methods. They are not implementable with asyncio as there is hidden blocking IO within them. E.g. `self.com_record[1].raw = ...` is blocking and non-portable. Since we don't want two parallel implementations (regular and async) of each, they must be changed into something portable. We'll discuss this later in another issue.

acolomb (changes requested):
As discussed, the writability check should apply to more parameters than just the mappings.  And in the current form, the conditional is looking at the wrong thing.

I'll try to come up with a different approach in another PR (#593).

erlend-aasland:
Closed in favour of #593. Thanks for your guidance and review, André; sorry for leaving this hanging.

PR #545, opened 2024-10-29
Add api for tpdo and rpdo

khauersp:
I was trying to use a local node and interacting with the setup for the tpdo and rpdo objects wasn't the most clear on how to set them up and get them working properly so I added some more functions to help with that process.


acolomb:
Why did you close this @khauersp?  I couldn't look at it in detail yet, but was going to review as time permits.

PR #559, opened 2025-02-18
Server sdo block upload (based on v2.3.0)

H-Grobben:
Based of version v2.3.0, added the block transfer for SDO Server, upload.

I've worked on this for version 2.1.0, but never got around to make a PR for this. Now I've updated to 2.3.0 lately it was time to do a PR for it.

Did add a test to the SDO unit tester. Don't know if it needs to be tested more.

### DIFF
--- a/canopen/sdo/constants.py
+++ b/canopen/sdo/constants.py
@@ -4,7 +4,6 @@ from typing import Final
 
 # Command, index, subindex
 SDO_STRUCT = struct.Struct("<BHB")
-SDO_BLOCKINIT_STRUCT = struct.Struct("<BHBI")  # Command + seqno, index, subindex, size
 SDO_BLOCKACK_STRUCT = struct.Struct("<BBB") # c + ackseq + new blocksize
 SDO_BLOCKEND_STRUCT = struct.Struct("<BH") # c + CRC
 SDO_ABORT_STRUCT = struct.Struct("<BHBI") # c +i + si + Abort code
--- a/canopen/sdo/constants.py
+++ b/canopen/sdo/constants.py
@@ -4,9 +4,10 @@ from typing import Final
 
 # Command, index, subindex
 SDO_STRUCT = struct.Struct("<BHB")
-SDO_BLOCK_STRUCT = struct.Struct("<B") # c + seqno
+SDO_BLOCKINIT_STRUCT = "<BHBI"  # Command + seqno, index, subindex, size
 SDO_BLOCKACK_STRUCT = struct.Struct("<BBB") # c + ackseq + new blocksize
 SDO_BLOCKEND_STRUCT = struct.Struct("<BH") # c + CRC
+SDO_ABORT_STRUCT = struct.Struct("<BHBI") # c +i + si + Abort code
 
 # Command[5-7]
 REQUEST_SEGMENT_DOWNLOAD = 0 << 5
--- a/canopen/sdo/constants.py
+++ b/canopen/sdo/constants.py
@@ -4,7 +4,7 @@ from typing import Final
 
 # Command, index, subindex
 SDO_STRUCT = struct.Struct("<BHB")
-SDO_BLOCKINIT_STRUCT = "<BHBI"  # Command + seqno, index, subindex, size
+SDO_BLOCKINIT_STRUCT = struct.Struct("<BHBI")  # Command + seqno, index, subindex, size
 SDO_BLOCKACK_STRUCT = struct.Struct("<BBB") # c + ackseq + new blocksize
 SDO_BLOCKEND_STRUCT = struct.Struct("<BH") # c + CRC
 SDO_ABORT_STRUCT = struct.Struct("<BHBI") # c +i + si + Abort code
--- a/canopen/sdo/constants.py
+++ b/canopen/sdo/constants.py
@@ -4,7 +4,11 @@ from typing import Final
 
 # Command, index, subindex
 SDO_STRUCT = struct.Struct("<BHB")
+SDO_BLOCK_STRUCT = struct.Struct("<B") # c + seqno
+SDO_BLOCKACK_STRUCT = struct.Struct("<BBB") # c + ackseq + new blocksize
+SDO_BLOCKEND_STRUCT = struct.Struct("<BH") # c + CRC
 
+# Command[5-7]
 REQUEST_SEGMENT_DOWNLOAD = 0 << 5
 REQUEST_DOWNLOAD = 1 << 5
 REQUEST_UPLOAD = 2 << 5
@@ -21,18 +25,31 @@ RESPONSE_ABORTED = 4 << 5
 RESPONSE_BLOCK_DOWNLOAD = 5 << 5
 RESPONSE_BLOCK_UPLOAD = 6 << 5
 
+# Block transfer sub-commands, Command[0-1]
+SUB_COMMAND_MASK = 0x3
 INITIATE_BLOCK_TRANSFER = 0
 END_BLOCK_TRANSFER = 1
 BLOCK_TRANSFER_RESPONSE = 2
 START_BLOCK_UPLOAD = 3
 
-EXPEDITED = 0x2
-SIZE_SPECIFIED = 0x1
-BLOCK_SIZE_SPECIFIED = 0x2
-CRC_SUPPORTED = 0x4
-NO_MORE_DATA = 0x1
-NO_MORE_BLOCKS = 0x80
-TOGGLE_BIT = 0x10
+EXPEDITED = 0x2             # Expedited and segmented
+SIZE_SPECIFIED = 0x1        # All transfers
+BLOCK_SIZE_SPECIFIED = 0x2  # Block transfer: size specified in message
+CRC_SUPPORTED = 0x4         # client/server CRC capable
+NO_MORE_DATA = 0x1          # Segmented: last segment
+NO_MORE_BLOCKS = 0x80       # Block transfer: last segment
+TOGGLE_BIT = 0x10           # segmented toggle mask
+
+# Block states
+BLOCK_STATE_NONE = -1
+BLOCK_STATE_INIT = 0        # state when entering
+BLOCK_STATE_UPLOAD = 0x10   # delimiter, used for block type check
+BLOCK_STATE_UP_INIT_RESP = 0x11 # state when entering, response during upload
+BLOCK_STATE_UP_DATA = 0x12     # Upload Data transfer state
+BLOCK_STATE_UP_END = 0x13      # End of Upload block transfers
+BLOCK_STATE_DOWNLOAD = 0x20 # delimiter, used for block type check
+BLOCK_STATE_DL_DATA = 0x24     # Download Data transfer state
+BLOCK_STATE_DL_END = 0x25      # End of Download block transfers
 
 # Abort codes
 ABORT_TOGGLE_NOT_ALTERNATED: Final[int] = 0x0503_0000
--- a/canopen/sdo/exceptions.py
+++ b/canopen/sdo/exceptions.py
@@ -1,5 +1,3 @@
-
-
 class SdoError(Exception):
     pass
 
@@ -35,12 +33,18 @@ class SdoAbortedError(SdoError):
         0x060A0023: "Resource not available",
         0x08000000: "General error",
         0x08000020: "Data cannot be transferred or stored to the application",
-        0x08000021: ("Data can not be transferred or stored to the application "
-                     "because of local control"),
-        0x08000022: ("Data can not be transferred or stored to the application "
-                     "because of the present device state"),
-        0x08000023: ("Object dictionary dynamic generation fails or no object "
-                     "dictionary is present"),
+        0x08000021: (
+            "Data can not be transferred or stored to the application "
+            "because of local control"
+        ),
+        0x08000022: (
+            "Data can not be transferred or stored to the application "
+            "because of the present device state"
+        ),
+        0x08000023: (
+            "Object dictionary dynamic generation fails or no object "
+            "dictionary is present"
+        ),
         0x08000024: "No data available",
     }
 
@@ -57,6 +61,13 @@ class SdoAbortedError(SdoError):
     def __eq__(self, other):
         """Compare two exception objects based on SDO abort code."""
         return self.code == other.code
+
+    @staticmethod
+    def from_string(text):
+        code = list(SdoAbortedError.CODES.keys())[
+            list(SdoAbortedError.CODES.values()).index(text)
+        ]
+        return code
 
 
 class SdoCommunicationError(SdoError):
--- a/canopen/sdo/exceptions.py
+++ b/canopen/sdo/exceptions.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 class SdoError(Exception):
     pass
 
@@ -33,23 +35,23 @@ class SdoAbortedError(SdoError):
         0x060A0023: "Resource not available",
         0x08000000: "General error",
         0x08000020: "Data cannot be transferred or stored to the application",
-        0x08000021: (
-            "Data can not be transferred or stored to the application "
-            "because of local control"
-        ),
-        0x08000022: (
-            "Data can not be transferred or stored to the application "
-            "because of the present device state"
-        ),
-        0x08000023: (
-            "Object dictionary dynamic generation fails or no object "
-            "dictionary is present"
-        ),
+        0x08000021: ("Data can not be transferred or stored to the application "
+                     "because of local control"),
+        0x08000022: ("Data can not be transferred or stored to the application "
+                     "because of the present device state"),
+        0x08000023: ("Object dictionary dynamic generation fails or no object "
+                     "dictionary is present"),
         0x08000024: "No data available",
     }
 
-    def __init__(self, code: int):
+    def __init__(self, code: Union[int, str]):
         #: Abort code
+        if isinstance(code, str):
+            try:
+                self.code = self.from_string(code)
+            except ValueError as e:
+                raise ValueError(f"Unknown SDO abort description: {code}") from e
+        else:
         self.code = code
 
     def __str__(self):
@@ -62,8 +64,7 @@ class SdoAbortedError(SdoError):
         """Compare two exception objects based on SDO abort code."""
         return self.code == other.code
 
-    @staticmethod
-    def from_string(text):
+    def from_string(self, text):
         code = list(SdoAbortedError.CODES.keys())[
             list(SdoAbortedError.CODES.values()).index(text)
         ]
--- a/canopen/sdo/exceptions.py
+++ b/canopen/sdo/exceptions.py
@@ -52,7 +52,7 @@ class SdoAbortedError(SdoError):
             except ValueError as e:
                 raise ValueError(f"Unknown SDO abort description: {code}") from e
         else:
-        self.code = code
+            self.code = code
 
     def __str__(self):
         text = f"Code 0x{self.code:08X}"
--- a/canopen/sdo/exceptions.py
+++ b/canopen/sdo/exceptions.py
@@ -60,21 +60,21 @@ class SdoAbortedError(SdoError):
             text = text + ", " + self.CODES[self.code]
         return text
 
-    def __eq__(self, other:Union['SdoAbortedError', int, str]):
+    def __eq__(self, other: object) -> bool:
         """Compare two exception objects based on SDO abort code."""
         if isinstance(other, int):
             other = SdoAbortedError(other)
         elif isinstance(other, str):
             other = SdoAbortedError(other)
         elif not isinstance(other, SdoAbortedError):
-            raise TypeError(f"Cannot compare SdoAbortedError with {type(other)}")
+            return NotImplemented
         return self.code == other.code
 
     def from_string(self, text: str) -> int:
         code = list(SdoAbortedError.CODES.keys())[
             list(SdoAbortedError.CODES.values()).index(text)
         ]
-        return SdoAbortedError(code)
+        return code
 
 
 class SdoCommunicationError(SdoError):
--- a/canopen/sdo/exceptions.py
+++ b/canopen/sdo/exceptions.py
@@ -60,15 +60,21 @@ class SdoAbortedError(SdoError):
             text = text + ", " + self.CODES[self.code]
         return text
 
-    def __eq__(self, other):
+    def __eq__(self, other:Union['SdoAbortedError', int, str]):
         """Compare two exception objects based on SDO abort code."""
+        if isinstance(other, int):
+            other = SdoAbortedError(other)
+        elif isinstance(other, str):
+            other = SdoAbortedError(other)
+        elif not isinstance(other, SdoAbortedError):
+            raise TypeError(f"Cannot compare SdoAbortedError with {type(other)}")
         return self.code == other.code
 
-    def from_string(self, text):
+    def from_string(self, text: str) -> int:
         code = list(SdoAbortedError.CODES.keys())[
             list(SdoAbortedError.CODES.values()).index(text)
         ]
-        return code
+        return SdoAbortedError(code)
 
 
 class SdoCommunicationError(SdoError):
--- a/canopen/sdo/server.py
+++ b/canopen/sdo/server.py
@@ -9,7 +9,8 @@ logger = logging.getLogger(__name__)
 
 
 class SdoBlockException(SdoAbortedError):
-    """ Dedicated SDO Block exception. """
+    """Dedicated SDO Block exception."""
+
 
 class SdoServer(SdoBase):
     """Creates an SDO server."""
@@ -33,20 +34,21 @@ class SdoServer(SdoBase):
         self.sdo_block = None
 
     def on_request(self, can_id, data, timestamp):
-        logger.debug('on_request')
+        logger.debug("on_request")
         if self.sdo_block and self.sdo_block.state != BLOCK_STATE_NONE:
             try:
                 self.process_block(data)
             except SdoAbortedError as exc:
                 self.sdo_block = None
                 self.abort(exc.code)
-            except Exception as exc:
+                raise
+            except Exception:
                 self.sdo_block = None
                 self.abort()
-                logger.exception(exc)
+                raise
             return
 
-        command, = struct.unpack_from("B", data, 0)
+        (command,) = struct.unpack_from("B", data, 0)
         ccs = command & 0xE0
 
         try:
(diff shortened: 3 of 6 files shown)


test/test_sdo.py
```diff
@@ -66,6 +67,7 @@ def _send_message(self, can_id, data, remote=False):
         self.assertSequenceEqual(data, next_data[1])
         self.assertEqual(can_id, 0x602)
         while self.data and self.data[0][0] == RX:
+            # print(" < 0x%03X:%s" % (0x580 + RX, binascii.hexlify(self.data[0][1])))
```
sveinse, line 70: Please remove this comment

test/test_sdo.py
```diff
@@ -82,25 +84,25 @@ def setUp(self):
 
     def test_expedited_upload(self):
         self.data = [
-            (TX, b'\x40\x18\x10\x01\x00\x00\x00\x00'),
-            (RX, b'\x43\x18\x10\x01\x04\x00\x00\x00')
+            (TX, b"\x40\x18\x10\x01\x00\x00\x00\x00"),
```
sveinse, line 87: This file contains a lot of clean code changes. E.g. change from ' -> " and other black improvements. We generally want these improvements, but this is mixed with the functional improvements, making the PR hard to read. Can you please create another PR containing only the formatting changes in order for this PR to become purely functional. Thanks.

canopen/sdo/exceptions.py
```diff
@@ -35,12 +33,18 @@ class SdoAbortedError(SdoError):
         0x060A0023: "Resource not available",
         0x08000000: "General error",
         0x08000020: "Data cannot be transferred or stored to the application",
-        0x08000021: ("Data can not be transferred or stored to the application "
-                     "because of local control"),
-        0x08000022: ("Data can not be transferred or stored to the application "
-                     "because of the present device state"),
-        0x08000023: ("Object dictionary dynamic generation fails or no object "
-                     "dictionary is present"),
+        0x08000021: (
```
sveinse, line 36: Formatting change. Put in separate PR, please. (See below for more info).

canopen/sdo/server.py
```diff
@@ -8,6 +8,10 @@
 logger = logging.getLogger(__name__)
 
 
+class SdoBlockException(SdoAbortedError):
+    def __init__(self, code: int):
+        super.__init__(self, code)
+
```
sveinse, line 14: Can't we simply do:

```py
class SdoBlockException(SdoAbortedError):
    """Put description here"""
```

canopen/sdo/server.py
```diff
@@ -217,3 +312,83 @@ def download(
             When node responds with an error.
         """
         return self._node.set_data(index, subindex, data)
+
+class SdoBlock():
```
sveinse, line 316: Since this is a new class, please provide class and methods docstrings. I would also strongly recommend that we add type hints to all method interfaces and to class attributes.

canopen/sdo/exceptions.py
```diff
@@ -58,6 +62,13 @@ def __eq__(self, other):
         """Compare two exception objects based on SDO abort code."""
         return self.code == other.code
 
+    @staticmethod
+    def from_string(text):
+        code = list(SdoAbortedError.CODES.keys())[
+            list(SdoAbortedError.CODES.values()).index(text)
+        ]
+        return code
+
```
sveinse, line 71: Instead of implementing this staticmethod, could a better solution be that if the `SdoAbortedError()` is instantiated with a `code` which is str, it looks up the numeric value for the error?

That being said, I know some linters (e.g. sonarqube) is giving errors if using string literals as identifiers in the code and suggest its better to create a constant.

canopen/sdo/server.py
```diff
@@ -217,3 +312,83 @@ def download(
...
+    def __init__(self, node, request, docrc=False):
+
+        command, index, subindex = SDO_STRUCT.unpack_from(request)
+        # only do crc if crccheck lib is available _and_ if requested
+        _req_crc = (command & CRC_SUPPORTED) == CRC_SUPPORTED
+
+        if (command & SUB_COMMAND_MASK) == INITIATE_BLOCK_TRANSFER:
+            self.state = BLOCK_STATE_INIT
+        else:
+            raise SdoBlockException(SdoAbortedError.from_string("Unknown SDO command specified"))
```
sveinse, line 334: Use the string directly in `SdoBlockException("Unknown SDO command specified")` as discussed above.

canopen/sdo/server.py
```diff
@@ -57,6 +67,77 @@ def on_request(self, can_id, data, timestamp):
             self.abort()
             logger.exception(exc)
 
+    def process_block(self, request):
```
sveinse, line 70: Please add function docstring and type hints.

canopen/sdo/server.py
```diff
@@ -57,6 +67,77 @@ def on_request(self, can_id, data, timestamp):
...
+            command, _, _= SDO_STRUCT.unpack_from(request)
+            # in upload state
+            if self.sdo_block.state == BLOCK_STATE_UP_INIT_RESP:
+                logger.debug('BLOCK_STATE_UP_INIT_RESP')
+                #init response was sent, client required to send new request
+                if (command & REQUEST_BLOCK_UPLOAD) != REQUEST_BLOCK_UPLOAD:
+                    raise SdoBlockException(0x05040001)
+                if (command & START_BLOCK_UPLOAD) != START_BLOCK_UPLOAD:
+                    raise SdoBlockException(0x05040001)
+                # self.sdo_block.update_state(BLOCK_STATE_UP_DATA)
```
sveinse, line 90: Please do not put code in comments (as well as the rest of the instances in this file).

canopen/sdo/server.py
```diff
@@ -57,6 +67,77 @@ def on_request(self, can_id, data, timestamp):
...
+            elif self.sdo_block.state == BLOCK_STATE_UP_DATA:
+                logger.debug('BLOCK_STATE_UP_DATA')
+                command, ackseq, newblk = SDO_BLOCKACK_STRUCT.unpack_from(request)
+                if (command & REQUEST_BLOCK_UPLOAD) != REQUEST_BLOCK_UPLOAD:
+                    raise SdoBlockException(0x05040001)
+                elif (command & BLOCK_TRANSFER_RESPONSE) != BLOCK_TRANSFER_RESPONSE:
+                    raise SdoBlockException(0x05040001)
+                elif (ackseq != self.sdo_block.last_seqno):
+                    self.sdo_block.data_uploaded = self.sdo_block.data_succesfull_upload
+
```
sveinse, line 109: Very minor, but please be consistent on formatting. Number of lines between code blocks.

canopen/sdo/server.py
```diff
@@ -57,6 +67,77 @@ def on_request(self, can_id, data, timestamp):
...
+                    for block in blocks:
+                        self.send_response(block)
+
+            elif self.sdo_block.state == BLOCK_STATE_UP_END:
+                self.sdo_block = None
+
+        elif BLOCK_STATE_DOWNLOAD < self.sdo_block.state:
+            logger.debug('BLOCK_STATE_DOWNLOAD')
+            # in download state
+            pass
```
sveinse, line 136: `pass` is not needed when `logger.debug()` exists.

canopen/sdo/server.py
```diff
@@ -57,6 +67,77 @@ def on_request(self, can_id, data, timestamp):
...
+
+                blocks = self.sdo_block.get_upload_blocks()
+                for block in blocks:
+                    self.send_response(block)
+
+            elif self.sdo_block.state == BLOCK_STATE_UP_DATA:
+                logger.debug('BLOCK_STATE_UP_DATA')
+                command, ackseq, newblk = SDO_BLOCKACK_STRUCT.unpack_from(request)
+                if (command & REQUEST_BLOCK_UPLOAD) != REQUEST_BLOCK_UPLOAD:
+                    raise SdoBlockException(0x05040001)
```
sveinse, line 104: Further down the new code is using string literals as exceptions, while here it's using numerical values. I think maybe this should be consistent.

canopen/sdo/server.py
```diff
@@ -106,12 +187,26 @@ def segmented_upload(self, command):
         response[1:1 + size] = data
         self.send_response(response)
 
-    def block_upload(self, data):
-        # We currently don't support BLOCK UPLOAD
-        # according to CIA301 the server is allowed
-        # to switch to regular upload
-        logger.info("Received block upload, switch to regular SDO upload")
-        self.init_upload(data)
+    def block_upload(self, request):
```
sveinse, line 190: Function docstring and type hints.

sveinse (changes requested):
Thank you for the PR! We really appreciate it.

I've mostly been looking at the coding aspects of the PR. As mentioned in the comments, there are many formatting changes, making it hard to evaluate the functional change (SDO block upload). I recommend doing the formatting improvement separately and first.

Some of these comments might seem like nitpicking, but we are generally working on improving the general quality of the project. It helps a lot if all contributions do the same. Just ask if there are any questions.

test/test_sdo.py
H-Grobben: I'll revert the black formatting and check if it all still works.

H-Grobben:
I updated the PR with the review points. Thanks for the input, and sorry for the black formatting. Company has switched to it just this year, so I need some getting used to the automatic formatting.

acolomb:
@H-Grobben you have an indentation error in `exceptions.py` and an unrelated breaking change in `test_sdo.py`.  Please make sure to run the test suite locally (just call `pytest` from the top-level of your clone) to avoid such mistakes.

If you had given permission for the project maintainters to commit to your PR branch, I would have fixed those two already.  It's just a small checkmark to set when creating the PR.

H-Grobben:
> @H-Grobben you have an indentation error in `exceptions.py` and an unrelated breaking change in `test_sdo.py`. Please make sure to run the test suite locally (just call `pytest` from the top-level of your clone) to avoid such mistakes.
> 
> If you had given permission for the project maintainters to commit to your PR branch, I would have fixed those two already. It's just a small checkmark to set when creating the PR.

Sorry for the indent, don't know why I missed that. Plain stupid.. But it's fixed.

For the access to the branch: external access is not allowed by the company (allowing external push/updates to the company repository) so i don't have the checkbox, and I can't find any other option to share it as editable for you, sorry.

acolomb (changes requested):
Sorry for not looking at the implementation more closely so far.  I've just checked the latest changes (see inline comment), then went on looking at the feature as a whole.

I still can't say much about the overall architecture, whether the extra `class SdoBlock` is needed at all, or should better be integrated directly.  What I did notice is that the code overall shows some sloppiness, not adhering to PEP-8 and other code quality standards for example.  As previously mentioned, complete typing info would be much desired for new code.  Some things are added but never actually used.  The logging bypasses the module-wide `logger` object sometimes, and gives very terse debug message contents, which doesn't blend in with the rest of the library.  The code has some more remaining whitespace issues, although not functionally relevant, but introducing needless noise.

I just want to know whether you're motivated to fix these issues and beat the code into good shape.  Surely someone can help you with these code styling details.  That would be easier if we had push access, but I understand why that's impossible in your case.

If you don't have the time for such details, would you still be willing to thoroughly test an implementation proposed in another PR which takes your approach and possibly does some details differently?

test/test_local.py
```diff
@@ -37,13 +37,14 @@ def test_expedited_upload(self):
...
-        # from block upload to expedite upload
-        self.assertEqual("Unexpected response 0x41", str(context.exception))
+    # Remove this test, as Block upload is now supported:
+    # def test_block_upload_switch_to_expedite_upload(self):
+    #     with self.assertRaises(canopen.SdoCommunicationError) as context:
+    #         with self.remote_node.sdo[0x1008].open('r', block_transfer=True) as fp:
+    #             pass
+    #     # We get this since the sdo client don't support the switch
+    #     # from block upload to expedite upload
+    #     self.assertEqual("Unexpected response 0x41", str(context.exception))
```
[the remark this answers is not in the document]
H-Grobben: Done.

sveinse:
@acolomb looking at the comments from codecov in the review, there are many (new) lines which are not covered by any tests. Where are with that on new contributions like this?

acolomb:
@sveinse frankly I haven't understood what the Codecov reports are trying to tell me.  Just not that much into automatic code metrics measurement, but my focus is more on my own logical judgment of the code.

That is not to say I'm satisfied with adding code that is not covered by unit tests.  But I think we have more urgent matters to tackle than reaching 100 % coverage.  For new code, of course the best time to add tests is now, while we're looking at it.  In that regard, I see it as a helpful indicator to judge how much effort the contributor put into (automated) testing.  Real-world testing might however be even more important for this library.

sveinse:
@acolomb on a general note (and somewhat a derailment in the PR): When migrating an existing project with no or little test coverage, its always good to state "from now on, we require 100% or >90% test coverage on commits". Now is the time that contributors have their head in this particular code, so its less effort to have it done now. Going back later to make test code for it is very hard to do.

Codecov is telling us the code coverage of the changed files in the PR, limited to the PR. That's what's great about it, because it can be an instrument to steer towards increasing the test coverage on contributions. -- But for some reason, this particular branch/PR isn't showing up right. It returns "No report found to compare against", so I don't really know what goes on here.

bizfsc:
I would like to help to make this PR merge ready. (Btw. I am @friederschueler, I am just forced to use my corporate account from now on)

acolomb:
Hi @bizfsc, I figured that was probably the background :-)

Happy if you can pick this up.  Real-world testing was one of the important points above IIRC, so are you up to any of that currently?

bizfsc:
Yes, I have some real world devices and software stacks I can use to test this. Also I will use my end-of-month tokens to generate some tests :)

H-Grobben:
We had the need more server block transfer. Busy with coding, maybe I (or my colleague) have some time to fix the tests and linters.

H-Grobben:
Rebased to latest HEAD, added tests and checked coverage reports. Please check if it needs some improvements.

canopen/sdo/constants.py
```diff
@@ -2,7 +2,12 @@
 
 # Command, index, subindex
 SDO_STRUCT = struct.Struct("<BHB")
+SDO_BLOCKINIT_STRUCT = "<BHBI"  # Command + seqno, index, subindex, size
```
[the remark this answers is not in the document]
H-Grobben: Added the sturct.Struct, but it isn't used, indeed. 

The code is using the default SDO_STRUCT, not the BLOCK INIT. So can be removed.

bizfsc:
Currently reviewing this, but a lot of changes.

bizfsc (changes requested):
Thanks for the continued work on this PR. The block transfer implementation is solid and the test coverage has improved significantly. I've identified a few issues that need to be addressed before merge.

## Critical Issues

### 1. Bug: `from_string()` returns `SdoAbortedError` instead of `int`

`from_string()` declares `-> int` but returns `SdoAbortedError(code)`. This means `self.code` ends up being an `SdoAbortedError` object when initialized from a string. mypy catches this. See inline suggestion.

### 2. `__eq__` violates Liskov Substitution Principle

The `__eq__` override should accept `object` as the parameter type and return `NotImplemented` for unsupported types (Python convention for comparison operators), rather than raising `TypeError`. See inline suggestion.

### 3. Breaking change: commenting out `readinto()`, `readable()`, `tell()` in `ReadableStream`

These methods are part of the `io.RawIOBase` interface. Commenting them out is a breaking change for any code that calls `readable()`, `tell()`, or `readinto()` on a `ReadableStream`. If they're not needed, remove them entirely — but consider that `tell()` and `readable()` are standard stream methods that consumers may rely on. At minimum, keep `readable()` and `tell()` since they're trivial and expected on readable streams.

Also, @acolomb's earlier review applies here too: don't leave commented-out code in the tree.

### 4. `logging.debug()` vs `logger.debug()` — inconsistent

`block_upload()` and `_SdoBlock.update_state()` use `logging.debug(...)` directly instead of the module-level `logger`. This bypasses the logger hierarchy and prevents per-module log configuration. See inline suggestion.

## Minor Issues

- `databyte != None` should be `databyte is not None` (PEP 8, identity check)
- `from canopen.sdo.constants import *` in `test/test_sdo.py` — wildcard imports in test code pollute the namespace; prefer explicit imports
- `self.crc = CRC_SUPPORTED if (docrc & _req_crc) else 0` — `docrc` is always `False` (default param), so CRC is never actually enabled. Is this intentional? If so, a TODO or comment clarifying the plan would help.
- `SdoBlockException` as a bare subclass of `SdoAbortedError` adds no value if it's always caught as `SdoAbortedError` in `on_request()`. Consider removing it and using `SdoAbortedError` directly with appropriate codes.

canopen/sdo/exceptions.py
```diff
@@ -44,20 +44,38 @@ class SdoAbortedError(SdoError):
         0x08000024: "No data available",
     }
 
-    def __init__(self, code: int):
+    def __init__(self, code: Union[int, str]):
```
acolomb, line 47: I'm not happy with this usage pattern.  We should stick to one style for using the SDO abort exceptions.  The established way is construction via explicit numeric IDs.  We have defined constants for all of them, so please use those instead of hard-coding strings or using magic number literals.  I'm aware that the constants were added after this PR was initially proposed.

This will make the PR more focused, as the changes to `SdoAbortedError` are not needed at all.

canopen/sdo/server.py
```diff
@@ -27,9 +27,24 @@ def __init__(self, rx_cobid, tx_cobid, node):
         self._index = None
         self._subindex = None
         self.last_received_error = 0x00000000
+        self.sdo_block = None
```
acolomb, line 30: Type hints for newly added code, please.  Since there is already a `BLOCK_STATE_NONE` value, do we really need to add another possible state `.sdo_block = None` instead of resetting the existing object?

canopen/sdo/server.py
```diff
@@ -27,9 +27,24 @@ def __init__(self, rx_cobid, tx_cobid, node):
...
 
     def on_request(self, can_id, data, timestamp):
-        command, = struct.unpack_from("B", data, 0)
+        logger.debug("on_request")
```
acolomb, line 33: This debug message is very generic and thus unlikely to help any user.

canopen/sdo/server.py
```diff
@@ -27,9 +27,24 @@ def __init__(self, rx_cobid, tx_cobid, node):
...
+            try:
+                self.process_block(data)
+            except SdoAbortedError as exc:
+                self.sdo_block = None
+                self.abort(exc.code)
+                raise
+            except Exception:
+                self.sdo_block = None
+                self.abort()
+                raise
```
acolomb, line 44: This appears to be bad API design, if the possible exception subclasses from `.process_block()` are not clearly documented and need such a catch-all.

The `self.sdo_block = None` could be moved to `self.abort()` to make sure we never miss resetting the state when a transfer is aborted.

Overall, this should be integrated into the ccs matching below, not open up its own SDO handling subculture.  The catch-all for not crashing the receiver thread is already implemented on that level.

canopen/sdo/server.py
```diff
@@ -185,8 +343,10 @@ def send_response(self, response):
 
     def abort(self, abort_code=ABORT_GENERAL_ERROR):
         """Abort current transfer."""
-        data = struct.pack("<BHBL", RESPONSE_ABORTED,
-                           self._index, self._subindex, abort_code)
+        if isinstance(abort_code, SdoAbortedError):
+            abort_code = abort_code.code
```
acolomb, line 347: Although there are no type hints on this method, it's rather clear that the argument should be an abort code.  We don't need more flexibility in using internal helper methods, but new code adhering to the established calling conventions.  As is, this has a small runtime overhead without any added benefit in code clarity.  Exceptions should be raised, not passed around, in most cases.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
             When node responds with an error.
         """
         return self._node.set_data(index, subindex, data)
+
+
+class _SdoBlock:
```
acolomb, line 390: I suppose this intends to mirror the block up/download stream classes in the SDO client implementation?  Of course, encapsulating the sequence number tracking and CRC calculation makes sense in a common class for upload and download.  But right now, there is no clear separation between functionality in this class and the `process_block()` method in the server.  Thus it should be stripped down to a simple provider / receiver of block segments, handling CRC and counters internally, or dissolved into state attributes directly in the `SdoServer` class.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+        sub_cmd_mask = 0x1 if is_download else SUB_COMMAND_MASK
+        if (command & sub_cmd_mask) == INITIATE_BLOCK_TRANSFER:
+            self.state = BLOCK_STATE_INIT
+        else:
+            # Realistically shouldnt happen since this is only called after receiving an initiate command, but check anyway
+            raise SdoAbortedError("Unknown SDO command specified") # pragma: no cover
+
+        # TODO: CRC of data if requested
+        self.crc = CRC_SUPPORTED if (docrc & _req_crc) else 0
+        self._node = node
```
acolomb, line 434: No need to store the associated node as an attribute, it is used only once, right in the constructor.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+            if command & BLOCK_SIZE_SPECIFIED:
+                (self.size,) = struct.unpack_from("<L", request, 4)
+            else:
+                self.size = None  # pragma: no cover
+        else:
+            self.req_blocksize = request[4]
+            if not 1 <= self.req_blocksize <= 127:
+                raise SdoAbortedError("Invalid block size")
+            self.data = self._node.get_data(index, subindex, check_readable=True)
+            self.size = len(self.data)
```
acolomb, line 453: Instead of passing a boolean to distinguish data flow direction, these attributes could be specified directly in the respective calling method of `SdoServer`.  Ideally, there should be a clear separation of concerns, not multiple places dissecting the SDO message structure separately and splitting the logic over several functions / classes.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+                    break
+            msgs.append(response)
+            self.last_seqno = seqno
+
+            if self.size == self.data_uploaded:
+                break
+        logger.debug(msgs)
+        return msgs
+
+    def get_data_byte(self):
```
acolomb, line 503: This is vastly inefficient, reading a single byte at a time into a buffer, after the whole data was previously already buffered.  At the very least, slicing should be applied to the existing buffer, copying a chunk as large as needed.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+            self._data_buffer = bytearray()
+            if command & BLOCK_SIZE_SPECIFIED:
+                (self.size,) = struct.unpack_from("<L", request, 4)
+            else:
+                self.size = None  # pragma: no cover
+        else:
+            self.req_blocksize = request[4]
+            if not 1 <= self.req_blocksize <= 127:
+                raise SdoAbortedError("Invalid block size")
+            self.data = self._node.get_data(index, subindex, check_readable=True)
```
acolomb, line 452: The whole point of SDO block transfer is to support rather large objects.  Buffering the complete data in memory is an avoidable waste of resources.  We already fail to properly support streaming access in the existing SDO segmented transfer code / storage in `LocalNode`.  But duplicating the data for a whole block here is going in the wrong direction.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+            CAN message containing SDO request.
+        :param docrc:
+            If True, CRC is calculated and checked.
+            TODO: implement CRC for block transfters, now always false.
+        :param is_download:
+            If True, initialise for block download (server receives data).
+            If False (default), initialise for block upload (server sends data).
+        """
+        command, index, subindex = SDO_STRUCT.unpack_from(request)
+        # only do crc if crccheck lib is available _and_ if requested
```
acolomb, line 418: Where is this "crccheck lib" actually searched for or used?

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+        self.subindex = subindex
+        self.seqno = 0
+
+        if is_download:
+            # Server defines the block size for download (client sends this many
+            # segments per block before waiting for an acknowledgement)
+            self.req_blocksize = 127
+            self._data_buffer = bytearray()
+            if command & BLOCK_SIZE_SPECIFIED:
+                (self.size,) = struct.unpack_from("<L", request, 4)
```
acolomb, line 445: Should use a predefined struct, like in other places.  Avoid hard-coding the offset.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+        self._node = node
+        self.index = index
+        self.subindex = subindex
+        self.seqno = 0
+
+        if is_download:
+            # Server defines the block size for download (client sends this many
+            # segments per block before waiting for an acknowledgement)
+            self.req_blocksize = 127
+            self._data_buffer = bytearray()
```
acolomb, line 443: The size extracted below (if available) should be used to pre-allocate a buffer, instead of always extending it with each new segment, possibly causing frequent reallocations.  A streaming API into the local data store object could mitigate this completely, whether or not a size is given in advance.

canopen/sdo/server.py
```diff
@@ -225,3 +385,146 @@ def download(
...
+        """
+        logger.debug("update_state %X -> %X", self.state, new_state)
+        if new_state >= self.state:
+            self.state = new_state
+        else:
+            raise SdoAbortedError(
+                "Data can not be transferred or stored to the application because of the present device state"
+            )
+
+    def get_upload_blocks(self):
```
acolomb, line 469: This definitely duplicates the data for one block in memory, including the protocol overhead.  A generator function or any other incremental processing would be in order.

acolomb (changes requested):
Thank you for investing more time to get this PR updated.  Sorry to say this, but I'm not satisfied with the overall code structure and quality.  The `LocalNode` stuff already feels like a tacked-on hack in some places.  But this protocol implementation largely doesn't even try to blend in with the rest of the code.  Instead adding helper code for doing things differently or circumventing existing solutions, which enlarges the diff significantly.  It exists in its own world.  Where efficient APIs for accessing larger objects are missing, it makes another copy of the data instead of trying to fix the deficiency (which is solved elegantly on the client side).

We need some substantial restructuring and architectural rethinking here, before this can be included in the library.  Not to mention the current expectations regarding code quality for added features, having clear type annotations as a minimum for robust development.  Coding style / formatting is not worthwhile to address now, given the bigger picture issues.  Some inline comments added to highlight where I see problems, hope that helps as a bit of guidance on expected style as well.

I haven't looked at the unit tests at all.  They may be valuable on their own, and could be included independently.  Marked as expected failure of course, until the actual block transfer implementation is included.

PR #567, opened 2025-04-04
Fix debian package generation to include subdirectories

civerachb-cpr:
The existing .deb generation does not include subdirectories, resulting in missing modules:

image

1. Move package information out of `makedeb` into `setup.py` as parameters
2. Explicitly list all python modules that should be included in the package
3. Replace the deb-generation command with `python3 setup.py --command-packages=stdeb.command bdist_deb`

Tested on Ubuntu 24.04; resulting debian file now includes the missing folders (see screenshot above)


acolomb:
Thanks for your contribution and for trying to improve the Debian package generation.

I don't think we want to proliferate stuff in `setup.py`.  The package has already adopted a more modern approach based on the declarative `pyproject.toml` file.  I have no experience with how Debian usually handles building Python packages, but I doubt that adding stuff to `setup.py` is state of the art.  Of course, neither is an elaborate `makedeb` script including logic to extract the version string, so there is definitely room for improvment.

Building the package by itself should be left to the PyPA build helper: `python3 -m build` in the distro's top level directory.

friederschueler:
Using debian as the main distro on many embedded devices, here is my take:

avoid debian packages for python at all. use instead a virtual env and be just happy.

create the **venv** yourself 

```
sudo apt install python3-full
python3 -m venv .venv
source .venv/bin/activate
pip install canopen
```
or use **pipx**

```
sudo apt install pipx
pipx ensurepath
pipx install canopen
```

because if you don't there is no simple way to create a fully compatible debian package.
we would have to pick some distros and releases we want to support. Debian and Ubuntu warn you about installing something into system directories not managed by an official package these days:

```
To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.
```

Although it would look kind of evil to remove the current setup.py and makedeb files, this is the direction I would take.

Issue #570, opened 2025-04-23
Integrate code quality tools (black, isort, ruff)


juliusz-t:
The tools allows to automatically detect (and fix) errors and style issues, so I suggest adding:

- [`black`](https://github.com/psf/black) for consistent code formatting,
- [`isort`](https://github.com/PyCQA/isort) for standardized import sorting,
- [`ruff`](https://github.com/astral-sh/ruff) for linting (error checking, style enforcement, etc.).

This will improve the consistency and readability of code. 

Tasks:
1. Add black, isort, and ruff as development dependencies.
2. Configure them via pyproject.toml.
3. Describe usage in README.
4. Integrate checks into the CI pipeline.

juliusz-t:
In [#569](https://github.com/canopen-python/canopen/issues/569#issuecomment-2822541100) a merge roadmap is proposed in terms of code quality (type hints and test coverage) and asyncio support. I will be happy to help you complete these two tasks. Could you evaluate the idea of ​​adding these tools in context of existing Pull Requests?

acolomb:
Please refrain from big "this fixes it all and reactors many things at the same time" pull requests. They are almost impossible to review and give little benefit for the time invested.

As with typing, I'm totally fine with adding the tool configuration, though we shouldn't have to specify many non-default options. Then anyone can run them locally or on the CI and send in smaller, self-contained pulls fixing individual (classes of) errors / warnings for a tool. When it's finally ready to run without errors, the linters can be configured as a mandatory check in the CI. Please be careful with functional changes / refactoring just because of some linter message. Plain reformatting usually has the lowest benefit / cost ratio when it comes to review time, and I do think thorough reviews are good practice for a stable library. 

Hope you're okay with that approach, as I think it's the only way to get some serious review during the process.

As for current development practice, I'm using mainly flake8 and have already pushed some smaller tidying up steps last year based on that.

PR #571, opened 2025-04-23
Code quality tools (#570)

juliusz-t:
PR related to issue #570

There are other issues regarding code quality and typing. It may be worth reviewing existing PRs and adding them to master before this change is accepted.

### DIFF
--- a/canopen/lss.py
+++ b/canopen/lss.py
@@ -19,16 +19,16 @@ CS_SWITCH_STATE_SELECTIVE_PRODUCT_CODE = 0x41
 CS_SWITCH_STATE_SELECTIVE_REVISION_NUMBER = 0x42
 CS_SWITCH_STATE_SELECTIVE_SERIAL_NUMBER = 0x43
 CS_SWITCH_STATE_SELECTIVE_RESPONSE = 0x44
-CS_IDENTIFY_REMOTE_SLAVE_VENDOR_ID = 0x46               # m -> s
-CS_IDENTIFY_REMOTE_SLAVE_PRODUCT_CODE = 0x47            # m -> s
-CS_IDENTIFY_REMOTE_SLAVE_REVISION_NUMBER_LOW = 0x48     # m -> s
-CS_IDENTIFY_REMOTE_SLAVE_REVISION_NUMBER_HIGH = 0x49    # m -> s
-CS_IDENTIFY_REMOTE_SLAVE_SERIAL_NUMBER_LOW = 0x4A       # m -> s
-CS_IDENTIFY_REMOTE_SLAVE_SERIAL_NUMBER_HIGH = 0x4B      # m -> s
-CS_IDENTIFY_NON_CONFIGURED_REMOTE_SLAVE = 0x4C          # m -> s
-CS_IDENTIFY_SLAVE = 0x4F                                # s -> m
-CS_IDENTIFY_NON_CONFIGURED_SLAVE = 0x50                 # s -> m
-CS_FAST_SCAN = 0x51                                     # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_VENDOR_ID = 0x46  # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_PRODUCT_CODE = 0x47  # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_REVISION_NUMBER_LOW = 0x48  # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_REVISION_NUMBER_HIGH = 0x49  # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_SERIAL_NUMBER_LOW = 0x4A  # m -> s
+CS_IDENTIFY_REMOTE_SLAVE_SERIAL_NUMBER_HIGH = 0x4B  # m -> s
+CS_IDENTIFY_NON_CONFIGURED_REMOTE_SLAVE = 0x4C  # m -> s
+CS_IDENTIFY_SLAVE = 0x4F  # s -> m
+CS_IDENTIFY_NON_CONFIGURED_SLAVE = 0x50  # s -> m
+CS_FAST_SCAN = 0x51  # m -> s
 CS_INQUIRE_VENDOR_ID = 0x5A
 CS_INQUIRE_PRODUCT_CODE = 0x5B
 CS_INQUIRE_REVISION_NUMBER = 0x5C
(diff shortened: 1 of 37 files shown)


juliusz-t:
I pushed the proposed tool configuration and the effect of their use. Please let me know if any changes have been made that are not acceptable to maintainers or parties using the project. The change is quite invasive from a code merge perspective. If I get the green light I can help implement compatible fixes in PRs that are waiting for approval for various reasons.

acolomb:
I'll need some time to judge this. And to be fair, actual feature work is still waiting in several PRs, so those should get attention first.

canopen/lss.py
```diff
@@ -49,7 +48,7 @@
 ERROR_STORE_NOT_SUPPORTED = 1
 ERROR_STORE_ACCESS_PROBLEM = 2
 
-ERROR_VENDOR_SPECIFIC = 0xff
+ERROR_VENDOR_SPECIFIC = 0xFF
```
acolomb, line 51: Sure.

canopen/lss.py
```diff
@@ -132,14 +132,15 @@ def send_switch_state_selective(self,
 
         self.__send_lss_address(CS_SWITCH_STATE_SELECTIVE_VENDOR_ID, vendorId)
         self.__send_lss_address(CS_SWITCH_STATE_SELECTIVE_PRODUCT_CODE, productCode)
-        self.__send_lss_address(CS_SWITCH_STATE_SELECTIVE_REVISION_NUMBER, revisionNumber)
-        response = self.__send_lss_address(CS_SWITCH_STATE_SELECTIVE_SERIAL_NUMBER, serialNumber)
+        self.__send_lss_address(
+            CS_SWITCH_STATE_SELECTIVE_REVISION_NUMBER, revisionNumber
```
acolomb, line 136: Definitely need to agree on a line length setting for Black.  Changes such as this one cause unnecessary churn and I don't think it's really solving a problem.  My usual compromise is to set a 96 character limit.  Not quite conservative (< 80), but also not as disruptive as allowing 120 chars (+50 %) and it leaves some margin for diff markings.

canopen/lss.py
```diff
@@ -132,14 +132,15 @@ def send_switch_state_selective(self,
...
+        response = self.__send_lss_address(
+            CS_SWITCH_STATE_SELECTIVE_SERIAL_NUMBER, serialNumber
+        )
 
         cs = struct.unpack_from("<B", response)[0]
-        if cs == CS_SWITCH_STATE_SELECTIVE_RESPONSE:
-            return True
-
-        return False
+        return cs == CS_SWITCH_STATE_SELECTIVE_RESPONSE
```
acolomb, line 143: Nice.

canopen/lss.py
```diff
@@ -197,19 +198,22 @@ def activate_bit_timing(self, switch_delay_ms):
         message = bytearray(8)
 
         message[0] = CS_ACTIVATE_BIT_TIMING
-        message[1:3] = struct.pack('<H', switch_delay_ms)
+        message[1:3] = struct.pack("<H", switch_delay_ms)
```
acolomb, line 201: We shouldn't get started with this, it has potential for endless discussions and no real benefit.  I actually configure Black with `skip-string-normalization = true` locally, to avoid such noise.

canopen/lss.py
```diff
@@ -197,19 +198,22 @@ def activate_bit_timing(self, switch_delay_ms):
...
         self.__send_command(message)
 
     def store_configuration(self):
-        """Store node id and baud rate.
-        """
+        """Store node id and baud rate."""
```
acolomb, line 205: Nice.  PEP8 if I remember correctly?  I use flake8 to help me with such things.

canopen/network.py
```diff
@@ -41,9 +40,9 @@ def __init__(self, bus: Optional[can.BusABC] = None):
         #: List of :class:`can.Listener` objects.
         #: Includes at least MessageListener.
         self.listeners = [MessageListener(self)]
-        self.notifier: Optional[can.Notifier] = None
-        self.nodes: Dict[int, Union[RemoteNode, LocalNode]] = {}
-        self.subscribers: Dict[int, List[Callback]] = {}
+        self.notifier: can.Notifier | None = None
+        self.nodes: dict[int, RemoteNode | LocalNode] = {}
+        self.subscribers: dict[int, list[Callback]] = {}
```
acolomb, line 45: What Python version is the minimum to support this?  We shouldn't break compatibility because of such minor changes.

pyproject.toml
```diff
@@ -48,3 +48,62 @@ testpaths = [
...
+"canopen/sdo/server.py" = [
+    "F403",  # unable to detect undefined names
+    "F405",  # variable defined from star imports
+]
+"test/test_eds.py" = [
+    "SIM117",  # nested with statements
+]
+"test/test_local.py" = [
+    "SIM117",  # nested with statements
+]
```
acolomb, line 103: Definitely don't get started collecting exemptions like this.  Either fix it or accept that a warning will pop up unless somebody steps up to cleanly refactor and get rid of the root cause.

acolomb (changes requested):
I skimmed over the changes now and must say sorry, I'm not happy with the result and the current settings.

Tried to add individual comments on some changes, to give you a feel for what is likely to be accepted and what is not.

In general, I think the classes of changes should be discussed and handled separately.  This "all things fixed (or silenced)" approach would take very long to review thoroughly, and discussion will be painful because it touches on so many opinionated subjects.

I'm open for code style improvements in general.  But rather by applying a consistent style to newly added code and stuff that's touched anyway as part of feature work.  As it stands, this PR will waste a lot of time for very little benefit, which can be better spent looking at the open feature PRs.  Sorry.

sveinse:
In @juliusz-t defense, most of these code changes are result of using black. And black is a) very opinionated on formatting, and b) has become the defacto standard in many projects. We should definitely take a stand if we want that in this project.

acolomb:
No need for defense, it's all being done in good faith and I hope that my reactions are not taken as an offense :-)

I definitely think that opting into rigorous "black" code styling is too much.  There are various reasons why deviating from it makes sense, such as the aligned comments I pointed out above.  It's certainly useful to consider what `python -m black --diff` suggests, and that is usually the best choice.  But I reject throwing away different but sensible formatting that was chosen consciously for good reasons.

On that ground, blindly applying all changes that black would make will not happen.  And I'm also reluctant to add specific silencing options just to make it shut up (except for the quotes thing which is just very noisy).  Take the tool as a tool, which can help finding decisions, but not as the single source of truth what's best for the project :-)

PR #573, opened 2025-04-26
Fix multiple associations and removals of networks

### LINKED ISSUE #562: Behavior and failure for uninitialized network
https://github.com/christiansandberg/canopen/blob/5382f245d8ede61b113c6a648cb99b199667f0cd/canopen/node/local.py#L51-L59

Calling `LocalNode.remove_network()` or `RemoteNode.remove_network()` on an already closed network will fail, due to `self.network` no longer have the `.unsubscribe()` method (which is expected). Depending on what behavior is wanted, there lacks some functionality with regards to the logic of uninitialized `LocalNode`. Related to #525

### Option 1 - multiple calls to `.remove_network()` should be ok
If it should be acceptable to call `.remove_network()` multiple times without failing -> then these methods should support that the network is uninitialized

```py
    def remove_network(self) -> None:
        if self.network is canopen.network._UNINITIALIZED_NETWORK:
            return
        # Rest of the function....
```

### Option 2 - multiple calls to `.remove_network()` should fail
If a repeated call to `.remove_network()` should fail or fail on uninitialized network, I think we're missing a function to test if the `LocalNode` object is uninitialized. 

`LocalNode` and friends it sets the network uninitialized on its `__init__()` and then later initialize it by `.associate_network()`. This means that there are code paths through the usage of `LocalNode` where a cleanup code might need to call `.remove_network()` conditionally depending on the state of its association with a network.

There is no current way to test if a `LocalNode` object is associated without using the private `canopen.network._UNINITIALIZED_NETWORK` object. Moreso, `_UNINITIALIZED_NETWORK` is inherited from `Network` so a negative `isinstance(node.network, Network)` does not help to determine if its initialized.

To prevent access to the private object, I suggest that the best approach would be to add a new `.has_network()` method to the `LocalNode` and `RemoteNode` implementation.

```py
    def has_network(self):
       """Return True if the node has an associated network."""
       return self.network is not canopen.network._UNINITIALIZED_NETWORK
```

sveinse:
This PR fixes the issue of not being able to call `associate_network()` and `remove_network()` multiple times. See #562.

Fixes #562

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -79,6 +79,9 @@ class Network(MutableMapping):
             del self.subscribers[can_id]
         else:
             self.subscribers[can_id].remove(callback)
+            # Remove empty list entry
+            if not self.subscribers[can_id]:
+                del self.subscribers[can_id]
 
     def connect(self, *args, **kwargs) -> Network:
         """Connect to CAN bus using python-can.
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -75,13 +75,10 @@ class Network(MutableMapping):
             If given, remove only this callback.  Otherwise all callbacks for
             the CAN ID.
         """
-        if callback is None:
-            del self.subscribers[can_id]
-        else:
+        if callback is not None:
             self.subscribers[can_id].remove(callback)
-            # Remove empty list entry
-            if not self.subscribers[can_id]:
-                del self.subscribers[can_id]
+        if not self.subscribers[can_id] or callback is None:
+            del self.subscribers[can_id]
 
     def connect(self, *args, **kwargs) -> Network:
         """Connect to CAN bus using python-can.
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -51,7 +51,6 @@ class LocalNode(BaseNode):
         network.subscribe(0, self.nmt.on_command)
 
     def remove_network(self) -> None:
-        # Make it safe to call this method multiple times
         if not self.has_network():
             return
         self.network.unsubscribe(self.sdo.rx_cobid, self.sdo.on_request)
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -39,6 +39,9 @@ class LocalNode(BaseNode):
         self.emcy = EmcyProducer(0x80 + self.id)
 
     def associate_network(self, network: canopen.network.Network):
+        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
+            # Unsubscribe from old network (to prevent double subscription)
+            self.remove_network()
         self.network = network
         self.sdo.network = network
         self.tpdo.network = network
@@ -49,6 +52,9 @@ class LocalNode(BaseNode):
         network.subscribe(0, self.nmt.on_command)
 
     def remove_network(self) -> None:
+        # Make it safe to call this method multiple times
+        if self.network is canopen.network._UNINITIALIZED_NETWORK:
+            return
         self.network.unsubscribe(self.sdo.rx_cobid, self.sdo.on_request)
         self.network.unsubscribe(0, self.nmt.on_command)
         self.network = canopen.network._UNINITIALIZED_NETWORK
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -39,9 +39,8 @@ class LocalNode(BaseNode):
         self.emcy = EmcyProducer(0x80 + self.id)
 
     def associate_network(self, network: canopen.network.Network):
-        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
-            # Unsubscribe from old network (to prevent double subscription)
-            self.remove_network()
+        if self.has_network():
+            raise RuntimeError("Node is already associated with a network")
         self.network = network
         self.sdo.network = network
         self.tpdo.network = network
@@ -53,7 +52,7 @@ class LocalNode(BaseNode):
 
     def remove_network(self) -> None:
         # Make it safe to call this method multiple times
-        if self.network is canopen.network._UNINITIALIZED_NETWORK:
+        if not self.has_network():
             return
         self.network.unsubscribe(self.sdo.rx_cobid, self.sdo.on_request)
         self.network.unsubscribe(0, self.nmt.on_command)
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -51,6 +51,9 @@ class RemoteNode(BaseNode):
             self.load_configuration()
 
     def associate_network(self, network: canopen.network.Network):
+        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
+            # Unsubscribe from old network (to prevent double subscriptions)
+            self.remove_network()
         self.network = network
         self.sdo.network = network
         self.pdo.network = network
@@ -64,6 +67,9 @@ class RemoteNode(BaseNode):
         network.subscribe(0, self.nmt.on_command)
 
     def remove_network(self) -> None:
+        # Make it safe to call this method multiple times
+        if self.network is canopen.network._UNINITIALIZED_NETWORK:
+            return
         for sdo in self.sdo_channels:
             self.network.unsubscribe(sdo.tx_cobid, sdo.on_response)
         self.network.unsubscribe(0x700 + self.id, self.nmt.on_heartbeat)
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -66,7 +66,6 @@ class RemoteNode(BaseNode):
         network.subscribe(0, self.nmt.on_command)
 
     def remove_network(self) -> None:
-        # Make it safe to call this method multiple times
         if not self.has_network():
             return
         for sdo in self.sdo_channels:
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -51,9 +51,8 @@ class RemoteNode(BaseNode):
             self.load_configuration()
 
     def associate_network(self, network: canopen.network.Network):
-        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
-            # Unsubscribe from old network (to prevent double subscriptions)
-            self.remove_network()
+        if self.has_network():
+            raise RuntimeError("Node is already associated with a network")
         self.network = network
         self.sdo.network = network
         self.pdo.network = network
@@ -68,7 +67,7 @@ class RemoteNode(BaseNode):
 
     def remove_network(self) -> None:
         # Make it safe to call this method multiple times
-        if self.network is canopen.network._UNINITIALIZED_NETWORK:
+        if not self.has_network():
             return
         for sdo in self.sdo_channels:
             self.network.unsubscribe(sdo.tx_cobid, sdo.on_response)
--- a/test/test_node.py
+++ b/test/test_node.py
@@ -1,20 +1,14 @@
-"""Unit tests for the RemoteNode and LocalNode classes."""
 import unittest
 
 import canopen
-import canopen.network
-import canopen.objectdictionary
 
 
 def count_subscribers(network: canopen.Network) -> int:
     """Count the number of subscribers in the network."""
-    return sum(
-        len(n) for n in network.subscribers.values()
-    )
+    return sum(len(n) for n in network.subscribers.values())
 
 
 class TestLocalNode(unittest.TestCase):
-    """Unit tests for the LocalNode class."""
 
     @classmethod
     def setUpClass(cls):
@@ -29,7 +23,6 @@ class TestLocalNode(unittest.TestCase):
         cls.network.disconnect()
 
     def test_associate_network(self):
-
         # Need to store the number of subscribers before associating because the
         # network implementation automatically adds subscribers to the list
         n_subscribers = count_subscribers(self.network)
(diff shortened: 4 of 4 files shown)


test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+import canopen
+
+from .util import SAMPLE_EDS
+
+
+def count_subscribers(network: canopen.Network) -> int:
+    """Count the number of subscribers in the network."""
+    return sum(
+        len(n) for n in network.subscribers.values()
+    )
```
sveinse, line 13: This function exists because `Network.unsubscribe()` doesn't properly clean up. It leaves a dict of zero-length lists per `can_id` on unsubscription. Does this warrant a separate PR to fix the `Network.subsubscribe()` code?

https://github.com/canopen-python/canopen/blob/48cca7c655f8c932b1f761a8edc11d0702397fab/canopen/network.py#L69-L81

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+class TestLocalNode(unittest.TestCase):
+    """
+    Test local node.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.network = canopen.Network()
+        cls.network.NOTIFIER_SHUTDOWN_TIMEOUT = 0.0
+        cls.network.connect("test", interface="virtual")
```
acolomb, line 25:
```suggestion
        cls.network.connect(interface="virtual")
```

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.network.disconnect()
+
+    def test_associate_network(self):
+
+        # Need to store the number of subscribers before associating because
+        # the current network implementation automatically adds subscribers
+        # to the list
```
acolomb, line 37:
```suggestion
        # Need to store the number of subscribers before associating because the
        # network implementation automatically adds subscribers to the list
```

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
```
acolomb, line 92:
```suggestion
        # Need to store the number of subscribers before associating because the
        # network implementation automatically adds subscribers to the list
```

test/test_node.py
acolomb: Yes I think it would be good measure to add something like this here:

    if not len(self.subscribers[can_id]):
        del self.subscribers[can_id]

And probably refactor to have that deletion only once, combining the two conditionals.

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+        self.assertEqual(count_subscribers(self.network), n_subscribers)
+
+        # Test that its possible to deassociate the network multiple times
+        self.node.remove_network()
+
+
+class TestRemoteNode(unittest.TestCase):
+    """
+    Test remote node.
+    """
```
acolomb, line 74: Better comment.

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        cls.network = canopen.Network()
+        cls.network.NOTIFIER_SHUTDOWN_TIMEOUT = 0.0
+        cls.network.connect("test", interface="virtual")
```
acolomb, line 80:
```suggestion
        cls.network.connect(interface="virtual")
```

test/test_node.py
```diff
@@ -0,0 +1,121 @@
+"""Unit tests for the RemoteNode and LocalNode classes."""
+import unittest
+
+import canopen
+
+from .util import SAMPLE_EDS
```
acolomb, line 6: I think we can get away without an EDS here at all.  Just initialize the nodes with an empty `canopen.objectdictionary.ObjectDictionary()`.

canopen/node/local.py
```diff
@@ -39,6 +39,9 @@ def __init__(
         self.emcy = EmcyProducer(0x80 + self.id)
 
     def associate_network(self, network: canopen.network.Network):
+        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
+            # Unsubscribe from old network (to prevent double subscription)
+            self.remove_network()
```
acolomb, line 44: Is it really common to re-associate a node to a different network?  I would think raising a `RuntimeError` exception here is more appropriate.

acolomb (changes requested):
Nice.  Mainly nit-picks, and I do welcome to use the existing `.has_network()` method.  I don't quite think automatically re-associating to a different network is appropriate.  Associating twice without removing from the first network will more likely point toward a programming error rather than a conscious, common workflow.

canopen/node/local.py
```diff
@@ -39,6 +39,9 @@ def __init__(
         self.emcy = EmcyProducer(0x80 + self.id)
 
     def associate_network(self, network: canopen.network.Network):
+        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
```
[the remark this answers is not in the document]
sveinse: Done.

canopen/node/remote.py
```diff
@@ -51,6 +51,9 @@ def __init__(
             self.load_configuration()
 
     def associate_network(self, network: canopen.network.Network):
+        if self.network is not canopen.network._UNINITIALIZED_NETWORK:
```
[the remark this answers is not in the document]
sveinse: Done.

test/test_node.py
```diff
@@ -0,0 +1,121 @@
...
+def count_subscribers(network: canopen.Network) -> int:
+    """Count the number of subscribers in the network."""
+    return sum(
+        len(n) for n in network.subscribers.values()
+    )
+
+
+class TestLocalNode(unittest.TestCase):
+    """
+    Test local node.
```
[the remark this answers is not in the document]
sveinse: Full sentence added on one line. The intention was that these two classes contain other unittests for these two classes, not just this particular test.

test/test_node.py
sveinse: Changed it to using `canopen.objectdictionary.ObjectDictionary()`

canopen/network.py
```diff
@@ -79,6 +79,9 @@ def unsubscribe(self, can_id, callback=None) -> None:
             del self.subscribers[can_id]
         else:
             self.subscribers[can_id].remove(callback)
+            # Remove empty list entry
+            if not self.subscribers[can_id]:
+                del self.subscribers[can_id]
```
acolomb, line 84: I meant something like this instead, with a single `del` code-path.  I find that easier to reason about:

```python3
        if callback is not None:
            self.subscribers[can_id].remove(callback)
        if not self.subscribers[can_id] or callback is None:
            del self.subscribers[can_id]
```

test/test_node.py
```diff
@@ -0,0 +1,113 @@
+"""Unit tests for the RemoteNode and LocalNode classes."""
+import unittest
+
+import canopen
+import canopen.network
+import canopen.objectdictionary
```
acolomb, line 6: These seem to be not needed.
```suggestion
```

test/test_node.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def count_subscribers(network: canopen.Network) -> int:
+    """Count the number of subscribers in the network."""
+    return sum(
+        len(n) for n in network.subscribers.values()
+    )
```
acolomb, line 13:
```suggestion
    return sum(len(n) for n in network.subscribers.values())
```

(black)

test/test_node.py
```diff
@@ -0,0 +1,113 @@
+"""Unit tests for the RemoteNode and LocalNode classes."""
```
acolomb, line 1: I'm not really fond of having comments just for the sake of having a comment.  None of our test files so far have a module-level docstring.  There is no generated API documentation for them.  And `import unittest` says just as much as this comment, while the enumeration of classes is redundant and bound to get out of sync sometime.

In essence, don't feel obliged to write docstrings that add no value.  Or try to come up with a really good one that does so.

What *IS* the common denominator for stuff that will be grouped in this test file?  So far it is only network association, what else do you have in mind?

test/test_node.py
```diff
@@ -0,0 +1,113 @@
...
+
+def count_subscribers(network: canopen.Network) -> int:
+    """Count the number of subscribers in the network."""
+    return sum(
+        len(n) for n in network.subscribers.values()
+    )
+
+
+class TestLocalNode(unittest.TestCase):
+    """Unit tests for the LocalNode class."""
```
acolomb, line 17: Not better :wink:

```suggestion
```

test/test_node.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertIs(self.node.nmt.network, uninitalized)
+        self.assertIs(self.node.emcy.network, uninitalized)
+        self.assertEqual(count_subscribers(self.network), n_subscribers)
+
+        # Test that its possible to deassociate the network multiple times
+        self.node.remove_network()
+
+
+class TestRemoteNode(unittest.TestCase):
+    """Unittests for the RemoteNode class."""
```
acolomb, line 68:
```suggestion
```

test/test_node.py
```diff
@@ -0,0 +1,113 @@
...
+        cls.network.connect(interface="virtual")
+
+        cls.node = canopen.RemoteNode(2, canopen.objectdictionary.ObjectDictionary())
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.network.disconnect()
+
+    def test_associate_network(self):
+
```
acolomb, line 83:
```suggestion
```

test/test_node.py
```diff
@@ -0,0 +1,113 @@
...
+        cls.network.connect(interface="virtual")
+
+        cls.node = canopen.LocalNode(2, canopen.objectdictionary.ObjectDictionary())
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.network.disconnect()
+
+    def test_associate_network(self):
+
```
acolomb, line 32:
```suggestion
```

acolomb (changes requested):

acolomb:
I took the liberty to apply my suggested changes.  Should be uncontroversial I hope.  Please indicate whether you are happy with the last commit's logic "simplification", it might just seem simpler to me :-)

acolomb (approved):

canopen/network.py
sveinse: Yes, two lines shorter.

test/test_node.py
sveinse: They are. Some linters, like pyright, is complaining when they are lacking. The reason is that they are not referenced in the `__init__.py` exports, so they need to be explicitly imported.

image

test/test_node.py
sveinse: > What _IS_ the common denominator for stuff that will be grouped in this test file? So far it is only network association, what else do you have in mind?

Eventually we'll have more unit tests of the methods in these two classes, no?

test/test_node.py
acolomb: I understand.  How come the code works without them though?

test/test_node.py
acolomb: Maybe, that's why I'm asking if you have any concrete ideas what will be added?

sveinse:
Comments updated. No objections.

You and I have very different coding styles my friend 😎 While in many projects the commit is rejected for not having docstrings,  here its removed because its not informative enough 🎉 It's fine, I just find it amusing. One certainly needs to learn every project's individual styles, that's for sure. Even the undocumented ones :D

PS! I don't know, but I don't have push authorization no more since the project move. Was that deliberate?

acolomb:
Yes, I know my stance on coding style is not average.  And I'm happy that you're open to adjusting.  It's mostly stuff where my experience tells me that exceptions from any rule might be better than following it blindly.  There should always be room for well-reasoned deviation.

> PS! I don't know, but I don't have push authorization no more since the project move. Was that deliberate?

Not that I know of.  The move was done by Christian, I haven't adjusted any permissions since.  What exactly can you no longer do?

sveinse:
> Yes, I know my stance on coding style is not average. And I'm happy that you're open to adjusting. It's mostly stuff where my experience tells me that exceptions from any rule might be better than following it blindly. There should always be room for well-reasoned deviation.

Well, in practice I don't have much choice, do I? Going into a PR where someone adamantly disagrees with the style is not a good way to create a cooperative environment, right. We have far harder challenges ahead of us with respect to asyncio than to squabble over formatting styles.

> > PS! I don't know, but I don't have push authorization no more since the project move. Was that deliberate?
> 
> Not that I know of. The move was done by Christian, I haven't adjusted any permissions since. What exactly can you no longer do?

I can't merge this branch, even when it's approved.

image

acolomb:
> I can't merge this branch, even when it's approved.

Hope it works now.  My intention was to disallow direct pushes, but merging pull requests should still be allowed for people with write access (which you do have).  I disabled the "Restrict who can push to matching branches" setting again, hope that has the desired effect.

acolomb:
Yes, squash and merge should do it.  Unless we have special circumstances with a perfectly hand-crafted commit history.  Which is pretty rare.

sveinse:
Merged! Thank you for the review @acolomb

acolomb:
Great! Thank you very much for your continued involvement and you patience with reviews :-)

acolomb:
One more thing about the removed imports, forgot to write it down:  The `canopen.objectdictionary.ObjectDictionary` is in fact available as just `canopen.ObjectDictionary`.  Which we should have used.

The same doesn't apply to the `_UNINITIALIZED_NETWORK` thing.  Which should be non-public API anyway.

As long as the tests work, I don't mind much about it though.  It's only unit test code :wink:

PR #575, opened 2025-04-26
Type annotate everywhere Network is used

sveinse:
To finalize #511 this PR implements type annotations for `Network` everywhere in the codebase, making typing of `Nework` feature complete. `Network` is a central object and it makes development much easier when the (typing) references are resolvable everywhere.

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -391,8 +391,11 @@ class NodeScanner:
 
     SERVICES = (0x700, 0x580, 0x180, 0x280, 0x380, 0x480, 0x80)
 
-    def __init__(self, network: Optional[Network] = None):
-        self.network = network
+    def __init__(self, network: Optional[Network] = _UNINITIALIZED_NETWORK):
+        # To preseve the old API where None is uninitialized network
+        if network is None:
+            network = _UNINITIALIZED_NETWORK
+        self.network: Network = network
         #: A :class:`list` of nodes discovered
         self.nodes: List[int] = []
 
@@ -408,8 +411,6 @@ class NodeScanner:
 
     def search(self, limit: int = 127) -> None:
         """Search for nodes by sending SDO requests to all node IDs."""
-        if self.network is None:
-            raise RuntimeError("A Network is required to do active scanning")
         sdo_req = b"\x40\x00\x10\x00\x00\x00\x00\x00"
         for node_id in range(1, limit + 1):
             self.network.send_message(0x600 + node_id, sdo_req)
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -391,8 +391,7 @@ class NodeScanner:
 
     SERVICES = (0x700, 0x580, 0x180, 0x280, 0x380, 0x480, 0x80)
 
-    def __init__(self, network: Optional[Network] = _UNINITIALIZED_NETWORK):
-        # To preseve the old API where None is uninitialized network
+    def __init__(self, network: Optional[Network] = None):
         if network is None:
             network = _UNINITIALIZED_NETWORK
         self.network: Network = network
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
+
 import copy
 import logging
 import re
 from configparser import NoOptionError, NoSectionError, RawConfigParser
+from typing import TYPE_CHECKING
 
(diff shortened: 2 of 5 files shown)


canopen/sync.py
```diff
@@ -1,13 +1,17 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
 from typing import Optional
```
acolomb, line 3:
```suggestion
from __future__ import annotations

from typing import Optional, TYPE_CHECKING
```

canopen/sync.py
```diff
@@ -1,13 +1,17 @@
...
 
+if TYPE_CHECKING:
+    import canopen.network
```
acolomb, line 6:
```suggestion
    import canopen.network

```

canopen/timestamp.py
```diff
@@ -1,7 +1,11 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
 import struct
 import time
 from typing import Optional
```
acolomb, line 5:
```suggestion
from __future__ import annotations

import struct
import time
from typing import Optional, TYPE_CHECKING
```

canopen/timestamp.py
```diff
@@ -1,7 +1,11 @@
...
 
+if TYPE_CHECKING:
+    import canopen.network
```
acolomb, line 8:
```suggestion
    import canopen.network

```

acolomb (commented):
I must have neglected to actually send this review last night...

One thing regarding the API, otherwise just styling hints.  In the hope that we will soon in sync regarding style rules... :wink:  My helpers are mostly just `flake8` and sometimes `black` for guidance.  Just in case you want to let those run in the background of your IDE as well.

acolomb:
We could leave this unmerged and try to get #518 done first.  Then this PR will be a perfect test case to see actual reductions in the mypy errors list.

canopen/network.py
```diff
@@ -391,8 +391,11 @@ class NodeScanner:
 
     SERVICES = (0x700, 0x580, 0x180, 0x280, 0x380, 0x480, 0x80)
 
-    def __init__(self, network: Optional[Network] = None):
-        self.network = network
+    def __init__(self, network: Optional[Network] = _UNINITIALIZED_NETWORK):
```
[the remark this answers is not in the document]
acolomb: Then simply `self.network = network or _UNINITIALIZED_NETWORK`?

sveinse:
Would you be so kind to share your flake8 settings? I normally don't use flake8 (*), and if I use it out of box it only complains about "line too long" from network.py. It would help to have the same settings so I can get these formatting

(*) I use pylance/pyright (as a part of VSCode), mypy and occasionally ruff and isort. Some projects use black rigorously

acolomb:
Nothing special, just a more convenient line length.  This is my `.flake8` at the project root.  Couldn't get it to work with these settings in `pyproject.toml` when I tried.

```
[flake8]
max-line-length = 96
inline-quotes = "
```

canopen/timestamp.py
sveinse: The following code is not a class so it's strictly not needed to have two lines according to black. Is your preference to always have two spaces below imports?

(PS! Can I encourage you to write a coding guideline document? It certainly would lessen your need to review and comment on styling.)

canopen/network.py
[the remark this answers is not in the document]
acolomb: Right, I forgot that it is derived from `MutableMapping` and therefore has a `__len__` dunder evaluating to false if there are no entries (nodes in this case).

The updated code is fine in that regard.  Thanks.

canopen/objectdictionary/eds.py
```diff
@@ -7,6 +9,8 @@
 from canopen.objectdictionary import ObjectDictionary, datatypes
 from canopen.sdo import SdoClient
 
+if TYPE_CHECKING:
+    import canopen.network
```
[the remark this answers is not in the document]
acolomb: Yes, I think it's common practice and desirable to separate the imports block from "actual code" by using a larger gap, just as before starting a class definition.

canopen/timestamp.py
acolomb: > Is your preference to always have two spaces below imports?

Yes.

Sorry I won't have much time to invest for documentation in the coming weeks I'm afraid.  Hope it gets better soon.

acolomb (approved):

PR #576, opened 2025-04-26
Correctly apply offset to custom timestamp value (fixes #563)

### LINKED ISSUE #563: TIME producer does not correct provided timestamp
The documentation on `transmit` suggests the provided timestamp should be a UNIX timestamp, so have Epoch Jan 1 1970 UTC. However, the function does not correctly offset the provided timestamp as it does with the one produced internally if no timestamp is provided.

Either the documentation should be fixed to say the provided timestamp should use the CANopen epoch or the like correcting the timestamp should be fixed to respect Python operator presence:

```
delta = (timestamp or time.time()) - OFFSET
```

sveinse:
Fix issue when providing a custom timestamp. Updated the test case to test provided and not provided timestamp.

Fixes #563

### DIFF
--- a/canopen/timestamp.py
+++ b/canopen/timestamp.py
@@ -26,7 +26,7 @@ class TimeProducer:
         :param float timestamp:
             Optional Unix timestamp to use, otherwise the current time is used.
         """
-        delta = timestamp or time.time() - OFFSET
+        delta = (timestamp or time.time()) - OFFSET
         days, seconds = divmod(delta, ONE_DAY)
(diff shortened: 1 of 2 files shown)


acolomb:
Good arguments :+1:

sveinse:
With all due respect, aren't we putting at lot of scrutiny into this one test case now? We have more than enough to chew in the main project... Please feel free to make the adjustments you see fit.

acolomb:
Sure, sorry for the noise.  I just needed some thinking out loud I guess, to understand what's going on here.  As unit tests are all about correctness, I wanted to understand what you're proposing and the original diff was just too much to see directly.

PR #587, opened 2025-06-11
Abort SDO server upload when the object has zero length

acolomb:
A data size of zero bytes cannot be encoded in the "number of bytes that do not contain data" semantics, as that would require n=4, which is however limited to two bits (maximum value 3).  So the SDO expedited upload protocol simply cannot convey that condition.  The current implementation however still tries and thereby corrupts the reserved, always-zero field value.

Avoid that protocol violation by responding with an SDO abort code of 0800 0024h, No data available.

Add a test case for this condition, verifying that the client raises the appropriate SdoAbortedError exception.

Inspired by the discussion in #566.

### DIFF
--- a/canopen/sdo/server.py
+++ b/canopen/sdo/server.py
@@ -66,7 +66,11 @@ class SdoServer(SdoBase):
 
         data = self._node.get_data(index, subindex, check_readable=True)
         size = len(data)
-        if size <= 4:
+        if size == 0:
+            logger.info("No content to upload for 0x%04X:%02X", index, subindex)
+            self.abort(0x0800_0024)
+            return
+        elif size <= 4:
             logger.info("Expedited upload for 0x%04X:%02X", index, subindex)
             res_command |= EXPEDITED
             res_command |= (4 - size) << 2
--- a/test/test_local.py
+++ b/test/test_local.py
@@ -62,6 +62,13 @@ class TestSDO(unittest.TestCase):
         sampling_rate = self.remote_node.sdo["Sensor Sampling Rate (Hz)"].raw
         self.assertAlmostEqual(sampling_rate, 5.2, places=2)
 
+    def test_upload_zero_length(self):
+        self.local_node.sdo["Manufacturer device name"].raw = b""
+        with self.assertRaises(canopen.SdoAbortedError) as error:
+            self.remote_node.sdo["Manufacturer device name"].data
+        # Should be No data available
+        self.assertEqual(error.exception.code, 0x0800_0024)
+
     def test_segmented_upload(self):
         self.local_node.sdo["Manufacturer device name"].raw = "Some cool device"
         device_name = self.remote_node.sdo["Manufacturer device name"].data


acolomb:
Why is it "not good" to receive an SDO abort message with appropriate abort code?

Either way, the SDO transaction ends with that message.  How is receiving an empty response (beacuse no data is available) any different from receiving an SDO abort message with the reason "no data available"?

Have you seen such behavior on actual devices, as a reference?  I mean this is our SDO server component, we can choose freely how to handle that case, as long as we don't violate the protocol.  Just because it is *possible* for some responses to indicate a length of zero, that doesn't mean it is in any way more correct than the alternative.

We can just as well distinguish by request type.  The server decides whether a segmented or expedited transfer response is sent.  So nothing stops us from sending a response with `e=0, s=1` and then encoding zero bytes in the `d` field.  Not following up with a further segment upload seems to be fine (emphasis added):

> SDOs are uploaded as a sequence of *zero or more* SDO upload segment services preceded by an SDO upload initiate service.

So the client might just not follow up with any SDO upload segment requests.  So how can we be sure the SDO transfer has ended?  The standard allows three cases:

1. Expedited response (e=1). [Cannot encode a zero size.]
2. Upload segment response with c=1. [Needs to be requested first by the client.]
3. Abort (*unsuccessful* completion). [Proposed solution in this PR.]
4. > A new SDO upload initiate request/indication, indicating the *unsuccessful* completion of the upload sequence and the start of a new sequence.

If the client doesn't continue requesting a segment because we sent a size of zero, that's fine and the next SDO request will implicitly terminate the transaction.  It then counts as an *unsuccessful* transfer, same as if we sent the SDO abort code.  Worst case, the client might send an SDO abort indication in response to our size=0 response.  If not, it's absoutely equivalent in terms of bus traffic whether we use the SDO abort or the upload initiate response.

acolomb:
Your responses are well thought out and I agree with your assessment.

1. It is a rather uncommon use case to have objects with zero length (even temporarily, such as your OS prompt).
2. I do see merit in facilitating it, in whatever way the specified protocol allows. Especially for DOMAIN objects, there are certainly valid reasons, like some kind of data streaming applications.
3. I understand your reasoning for avoiding the SDO abort code. Although I think it's more of an implementation problem in the (any) client if they do *not* expect such a response, even though they perfectly anticipated the case of no data being available.
4. This change only concerns our SDO *server* component, which means we can choose any response behavior which the standard covers as valid.
5. Right now (and especially for this PR) we're only concerned with SDO *upload* and the *server* implementation chooses whether a size is even indicated if the available data has zero length.  I confirm your other scenarios though, they look correct to me.  The download cases should be checked against our client code, possibly resulting in another PR.
6. I accept that issue is only about *expedited* transfers.  For upload, the server MUST NOT use an expedited response if the intention is to signal zero data available.  For download, the client simply cannot make such a request while setting the e=1, s=1, as it would convey at least one byte.  e=1, s=0 is theoretically possible, which punts the decision to the server how much data to interpret.
7. I'd extend your listed sequences by a variant which does in fact keep the transfer pending, by sending an upload segment response with c=0, n=7, which can be used for quickly polling repeatedly.  The advantage is that there needs to be no further initiate request, and 7 bytes can be transferred immediately with each request when data does become available.
8. I disagree with your interpretation of Figure 23.  It is just a figure and the text explicitly says that zero is a possible number of segments.  The picture just can't convey that easily.

Now to the really important points: Our server currently gives totally bogus responses for an object with zero data to encode.
* It uses expedited transfer, which is the only(?) mode not capable of signaling zero size.
* It specifies a data size of four bytes.
* Most importantly, it thrashes the "x" flag, clearly violating the specification of "always 0".
* It inflates the response size by inserting byte values because `[4:4+0]` is a zero-length slice.  I suspect that might even throw off some backend interfaces when trying to transmit the frame.
* Not specifying the size isn't even considered in our server upload code (always sets s=1).

Whatever we do, we should fix the clear protocol violations.  Aborting the transfer as suggested is just a simple reaction to the special case.

If you feel like supporting more of the possible response behaviors, I'm open to consider adding them.  The SDO abort is not set in stone, just the most obvious valid choice for me.  But feel free to make an alternative PR with a different choice.  If it ends up requiring a matching response to the following upload segment request, that should be implemented as well though.  The current approach avoids that by ending the transaction (which can easily be retried anyway).

sveinse:
I have to humbly admit that I had completely overlooked that this PR is discussing the SDO server, not the client. _That_ changes things completely.

> Now to the really important points: Our server currently gives totally bogus responses for an object with zero data to encode.
> 
> * It uses expedited transfer, which is the only(?) mode not capable of signaling zero size.
> * It specifies a data size of four bytes.
> * Most importantly, it thrashes the "x" flag, clearly violating the specification of "always 0".
> * It inflates the response size by inserting byte values because `[4:4+0]` is a zero-length slice.  I suspect that might even throw off some backend interfaces when trying to transmit the frame.
> * Not specifying the size isn't even considered in our server upload code (always sets s=1).
> 
> Whatever we do, we should fix the clear protocol violations. Aborting the transfer as suggested is just a simple reaction to the special case.

Agreed. The current implementation is not handling 0 sized data right and must be fixed. On upload its the server's privilege to chose expedited vs segmented, so it makes sense its also the server's privilege (and our choice) to not support zero-length data. This will be compliant with the standard.

> If you feel like supporting more of the possible response behaviors, I'm open to consider adding them. The SDO abort is not set in stone, just the most obvious valid choice for me. But feel free to make an alternative PR with a different choice. If it ends up requiring a matching response to the following upload segment request, that should be implemented as well though. The current approach avoids that by ending the transaction (which can easily be retried anyway).

Again, I'm sorry for this apparent 180 degrees turn - I missed a crucial piece of information. I should have read the PR better before typing off.

sveinse (approved):
I think we can say this PR has been thoroughly discussed. Approved.

acolomb:
> Again, I'm sorry for this apparent 180 degrees turn - I missed a crucial piece of information.

No problem, misunderstandings do happen.  And discussion usually helps get a clearer picture by understanding other people's perspectives.

PR #589, opened 2025-06-14
Fix OD item lookup evaluating object's length (fixes #588)

### LINKED ISSUE #588: Incorrect fetch from the OD objects due to their `__len__` method
As I was looking at #436 I did some tracing and noticed that `ODVariable.__len__` is called many more times than expected and I uncovered a side effect:

In `ObjectDictionary.__getitem__()` we have in L136 the line `item = self.names.get(index) or self.indices.get(index)`. What happens here is that python looks up the index from `self.names` first. When it has found an object, it evaluates if this object is false, and it does so by running its `__len__()` function. That it will call `ODVariable.__len__`, `ODArray.__len__` or `ODRecord.__len__`. If this happens to be 0, the object fails the test it moves on to the next step, potentially returning `None`.

Reference: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

I assume we might be interested in fixing this?

https://github.com/canopen-python/canopen/blob/7ddb19b2ca7b7a85b9cef3090249b00293dd4b4a/canopen/objectdictionary/__init__.py#L132-L142

sveinse:
Here is a proposal for a fix

Fixes #588

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -133,7 +133,9 @@ class ObjectDictionary(MutableMapping):
         self, index: Union[int, str]
     ) -> Union[ODArray, ODRecord, ODVariable]:
         """Get object from object dictionary by name or index."""
-        item = self.names.get(index) or self.indices.get(index)
+        item = self.names.get(index)
+        if item is None:
+            item = self.indices.get(index)
         if item is None:
             if isinstance(index, str) and '.' in index:
                 idx, sub = index.split('.', maxsplit=1)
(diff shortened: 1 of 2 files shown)


acolomb (approved):

sveinse:
Ok, understood. Another thing that should be documented in "how we do things" as this differs between projects :D

acolomb:
> Thanks! I am not permitted to make the merge even you approved it @acolomb, so I don't have all permissions required it seems.

Sorry about that.  I can't see why not, will have to investigate a little.

PR #591, opened 2025-06-15
Avoid truncating unknown types in SDO upload (fixes #436)

### LINKED ISSUE #436: Inconsistent data truncation of unsupported data types in SDO upload

acolomb:
When an OD entry is found in `SdoClient.upload()`, the truncation is currently skipped only for those types explicitly listed as variable length data (domain, strings).  That causes unknown types to be truncated to whatever their length indicates, which is usually one byte.

Invert the condition to check all types with an explicitly known size, a.k.a. those listed in STRUCT_TYPES where the required length can be deduced from the structure format.  Factor out the condition to a new `fixed_size` property, which is easier to override when adding custom data types.

Re-enable the unit tests which were skipped based on the previously buggy behavior.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -419,6 +419,12 @@ class ODVariable:
         """
         self.bit_definitions[name] = bits
 
+    @property
+    def fixed_size(self) -> bool:
+        """Indicate whether the amount of needed data is known in advance."""
+        # Only for types which we parse using a structure.
+        return self.data_type in self.STRUCT_TYPES
+
     def decode_raw(self, data: bytes) -> Union[int, float, str, bytes, bytearray]:
         if self.data_type == VISIBLE_STRING:
             # Strip any trailing NUL characters from C-based systems
--- a/canopen/sdo/client.py
+++ b/canopen/sdo/client.py
@@ -126,9 +126,7 @@ class SdoClient(SdoBase):
         var = self.od.get_variable(index, subindex)
         if var is not None:
             # Found a matching variable in OD
-            # If this is a data type (string, domain etc) the size is
-            # unknown anyway so keep the data as is
-            if var.data_type not in objectdictionary.DATA_TYPES:
+            if var.fixed_size:
                 # Get the size in bytes for this variable
                 var_size = len(var) // 8
                 if response_size is None or var_size < response_size:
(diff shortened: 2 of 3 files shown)


sveinse:
This is an efficient fix.

acolomb:
Does that mean PR approval?

sveinse (approved):
> Does that mean PR approval?

If our goal is to keep the design as close to as-is, this I think is the minimal efforts to fulfil #436. And it has the minimal impact on the existing code base. Approved.👌

PR #593, opened 2025-06-17
Gracefully handle readonly entries in PdoMap.save()  (fixes #541)

### LINKED ISSUE #541: `PdoMap.save()` does not handle cases where 1A00h through 1A03h are read-only

acolomb:
For PDO communication or mapping parameters which are read-only on a
node, the `PdoMap.save()` method currently aborts with an SDO exception.
However, individual parameters may simply not support configuration
changes.  Rely on the Object Dictionary to tell which ones are
expected to fail and skip them.  For communication record entries, log
what is not being saved.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -433,8 +433,8 @@ class PdoMap:
             # mappings for an invalid object 0x0000:00 to overwrite any
             # excess entries with all-zeros.
             self._fill_map(self.map_array[0].raw)
-        for var, entry in zip(self.map, self.map_array):
-            if not entry.writable:
+        for var, entry in zip(self.map, self.map_array.values()):
+            if not entry.od.writable:
                 continue
             logger.info(
                 "Writing %s (0x%04X:%02X, %d bits) to PDO map",
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -15,7 +15,7 @@ from canopen.sdo import SdoAbortedError
 if TYPE_CHECKING:
     from canopen import LocalNode, RemoteNode
     from canopen.pdo import RPDO, TPDO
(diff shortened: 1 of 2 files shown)


acolomb:
The one missing line reported by Codecov is related to the hack for supporting Curtis' mixed up mapping entries.  I don't feel like putting effort into that part.

sveinse:
I'd like to add something to consider.

I know that this PR is not considering the upcoming asyncio PR yet, but I'd like to inform that the `PdoMap.save()` and `PdoMap.read()` will need to be refactored for async. Today the logic and the IO is done together, which doesn't play nice with asyncio. To prevent repeating this logic twice, the asyncio PR refeactors the code following the design principles of [_sansio_](https://sans-io.readthedocs.io/), where the IO is done outside of the logic function.

I would have love if we were able to take this into consideration when refactoring `PdoMap.save()` in thie PR. It would gain us all to think ahead.

sveinse:
No problem. As a consequence I need to update the asyncio PR to reflect the result of this PR.

sveinse (approved):
Are you formatting with black on the new stuff? Because its going a bit back an forth between splitting on the lines and joining them. E.g. base.py  L439 is split up, while L450 is collected together.

It's cosmetics. The code changes looks good.

acolomb:
Yes, I take black as a reference for how to format new / changed code.  In this case, it's simply based on the line length limit of 96 I have configured.  The old style was non-conformant in both cases, and the new loop variable allowed some shortening.

PR #594, opened 2025-06-17
sdo.client: Add missing abort messages

acolomb:
SDO abort messages were sent in some cases when a response times out or has an unexpected server command specifier.  But not consistently, thus the following cases are now added:

* Mismatched scs:
  * ReadableStream, after upload segment request
  * WritableStream, after download initiate request
  * WritableStream, after expedited download request
  * WritableStream, after download segment request

* Toggle bit mismatch (reports as not toggled):
  * ReadableStream, after upload segment request

Some reformatting of abort code literals squeezed in there, as well as a typo fix.

This is based on the changes in #352, isolated from the typing noise.

### DIFF
--- a/canopen/sdo/client.py
+++ b/canopen/sdo/client.py
@@ -64,12 +64,13 @@ class SdoClient(SdoBase):
             else:
                 break
 
-    def read_response(self):
+    def read_response(self, timeout_abort: bool = False):
         try:
             response = self.responses.get(
                 block=True, timeout=self.RESPONSE_TIMEOUT)
         except queue.Empty:
-            self.abort(0x0504_0000)
+            if timeout_abort:
+                self.abort(0x0504_0000)
             raise SdoCommunicationError("No SDO response received")
         res_command, = struct.unpack_from("B", response)
         if res_command == RESPONSE_ABORTED:
@@ -86,7 +87,7 @@ class SdoClient(SdoBase):
             self.send_request(sdo_request)
             # Wait for node to respond
             try:
-                return self.read_response()
+                return self.read_response(timeout_abort=False)
             except SdoCommunicationError as e:
                 retries_left -= 1
                 if not retries_left:
@@ -526,7 +527,7 @@ class BlockUploadStream(io.RawIOBase):
             return self.readall()
 
         try:
-            response = self.sdo_client.read_response()
+            response = self.sdo_client.read_response(timeout_abort=False)
         except SdoCommunicationError:
             response = self._retransmit()
         res_command, = struct.unpack_from("B", response)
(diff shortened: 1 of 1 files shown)


acolomb:
Sending the abort message when `SdoClient.read_response()` times out is not always desired, as some code paths have their own abort / retry handling.  This needs to be investigated more for each added case.

canopen/sdo/client.py
```diff
@@ -64,11 +64,13 @@ def send_request(self, request):
...
                 break
 
-    def read_response(self):
+    def read_response(self, timeout_abort: bool = False):
         try:
             response = self.responses.get(
                 block=True, timeout=self.RESPONSE_TIMEOUT)
         except queue.Empty:
+            if timeout_abort:
+                self.abort(0x0504_0000)
```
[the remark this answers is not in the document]
sveinse: Since we're touching the code with this PR, I think I would suggest doing a quick and dirty fix to add them as literals. It can still be made as a minimal disruption change. E.g. we don't need to make literals of all SDO Errors, just the few we touch here.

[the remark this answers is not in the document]
acolomb: Let's focus on the logic first.  I will follow up with the literals change once this is merged.  The bigger question is whether we want to get rid of the extra parameter again, as mentioned in https://github.com/canopen-python/canopen/pull/594#issuecomment-2983404197?

canopen/sdo/client.py
```diff
@@ -302,8 +304,10 @@ def read(self, size=-1):
         response = self.sdo_client.request_response(request)
         res_command, = struct.unpack_from("B", response)
         if res_command & 0xE0 != RESPONSE_SEGMENT_UPLOAD:
+            self.sdo_client.abort(0x0504_0001)
```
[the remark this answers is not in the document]
acolomb: See #596, as a follow-up PR, yet in the making.

canopen/sdo/client.py
```diff
@@ -576,15 +581,19 @@ def _ack_block(self):
         self._ackseq = 0
 
     def _end_upload(self):
-        response = self.sdo_client.read_response()
+        try:
+            response = self.sdo_client.read_response()
+        except SdoCommunicationError:
+            self.abort(0x0504_0000)
```
sveinse, line 587: This assumes that `read_response()` only generates `SdoCommunicationError` on timeout. It does so now, so it works, but this assumption might be a little fragile.

canopen/sdo/client.py
```diff
@@ -734,14 +743,18 @@ def tell(self):
 
     def _block_ack(self):
         logger.debug("Waiting for acknowledgement of last block...")
-        response = self.sdo_client.read_response()
+        try:
+            response = self.sdo_client.read_response()
+        except SdoCommunicationError:
+            self.sdo_client.abort(0x0504_0000)
```
sveinse, line 749: Ditto

sveinse (commented):
I've reviewed the changes and I've verified all the changed SDO codes against the 301 standard. I've added two comments which I think should be considered. Other that that this looks good.

canopen/sdo/client.py
acolomb: Hm... I don't see that as fragile, it's even in the same source file to check against.

You're right in the opposite direction though.  We don't process other exceptions by sending a different abort message.  But that's a different improvement to be made IMHO.

sveinse: Thank you. Looks good.

sveinse (approved):

acolomb:
Thanks @sveinse!

PR #600, opened 2025-07-15
sync: Prevent from starting periodic task when already running (fixes #582)

### LINKED ISSUE #582: Reference to task is lost when starting a second Sync periodic
Consider this snippet: 

```python
import canopen
network = canopen.Network()
network.connect(channel='vcan0', bustype="socketcan")
network.sync.start(1.0)
network.sync.start(0.5)
# Using `candump -L vcan0` I have the two tasks running in parallel
network.sync.stop() # Only the latter task is stopped
network.sync.stop() # CanOperationError, the latter task is already stopped
```

In `sync.py`, we notice that attribute `self._task` is erased with the new task, which became inaccessible. I don't know if this hack will work, but we could only allow one sync event at the time: 

```python
diff --git a/canopen/sync.py b/canopen/sync.py
index d373451..6965b94 100644
--- a/canopen/sync.py
+++ b/canopen/sync.py
@@ -33,9 +33,14 @@ class SyncProducer:
         if not self.period:
             raise ValueError("A valid transmission period has not been given")
 
+        if self._task is not None:
+            self._task.stop()
+            self._task = None
+
         self._task = self.network.send_periodic(self.cob_id, [], self.period)
 
     def stop(self):
         """Stop periodic transmission of SYNC message."""
         if self._task is not None:
             self._task.stop()
+            self._task = None
```

Does it make sens?

acolomb:
Raise a RuntimeError when a SYNC transmission task has previously been
started.  Make sure to clear the internal _task attribute when
stopping, to make the check work.

### DIFF
--- a/canopen/sync.py
+++ b/canopen/sync.py
@@ -32,6 +32,9 @@ class SyncProducer:
         :param period:
             Period of SYNC message in seconds.
         """
+        if self._task is not None:
+            raise RuntimeError("Periodic SYNC transmission task already running")
+
         if period is not None:
             self.period = period
 
@@ -44,3 +47,4 @@ class SyncProducer:
         """Stop periodic transmission of SYNC message."""
         if self._task is not None:
             self._task.stop()
+        self._task = None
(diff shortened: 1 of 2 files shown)


acolomb:
I squeezed some more fixes in, regarding type hints and possible exceptions (now listed in the docstrings).

PR #602, opened 2025-07-15
PdoBase: Forbid getting item with zero index (fixes #581)

### LINKED ISSUE #581: PDO getitem unexpected returns `PdoVariable` on value 0
Reading PDO maps from a node is a 1-indexed value. However, if the user use index 0 (which is a common expectation for many indexes), `__getitem__` returns a `PdoVariable` instead of the expected `PdoMap`.

```py
node = network.create_node(42, "mynode.eds")
node.rpdo.aread(from_od=False)

# Use correct 1 index
rpdo = node.rpdo[1]
# Type returned is `PdoMap`

# Use incorrect 0 index
rpdo = node.rpdo[0]
# Type returned is a `PdoVariable`
```

The source is due to the else in this code https://github.com/canopen-python/canopen/blob/e840449fbf9ddf810d14899279cb72e61a2447f0/canopen/pdo/base.py#L42-L54

We just spent an hour trying to debug why the PDO didn't work before we realized the obvious mistake. I think using 0 index is such a common expectation, that I think we should give an error or clearly warn the user.

What is the intent of the the else section in L48-L53? When is this used? I could not find any tests showing how this is used. Because of the else `rpdo[0]` as shown above will fetch the first `PdoVariable` from the first PDO, but `rpdo[1]` will not fetch the second `PdoVariable`.

acolomb:
The `PdoBase` mapping uses an enumeration starting from one for the individual PDOs.  Requesting index 0 (zero) accidentally will return the first `PdoVariable` in the first PDO.  This is conincidental, because the PDOs are tried in turn and with a PDO, the numeric index lookup is zero-based.

To avoid hard-to-debug errors, simply forbid the value zero completely in the lookup, which does not make sense for any of the supported lookup methods.


acolomb:
Well this was meant to target v3.0, but the checks do not work seamlessly for the branch other than master.  Will probably reopen to target master, but don't let that keep you from reviewing / discussion.

acolomb:
Superseded by #609.

PR #614, opened 2025-09-11
xdd: Implement option to load object dictionary from XDD file.

zaporozhets:
Hi All,

This is a WIP merge request that adds support for XDD import to CANopen. I'm still in the process of testing and making improvements, but I’d appreciate any feedback you might have in the meantime.

Feel free to share any questions or suggestions.

Best regards,
Taras

### DIFF
--- a/canopen/objectdictionary/xdd.py
+++ b/canopen/objectdictionary/xdd.py
@@ -1,0 +1,413 @@
+import functools
+import logging
+import re
+import xml.etree.ElementTree as etree
+from typing import Any, Callable, Optional, Union, TextIO
+from canopen.objectdictionary import (
+    ODArray,
+    ODRecord,
+    ODVariable,
+    ObjectDictionary,
+    datatypes,
+    objectcodes,
+)
+
+logger = logging.getLogger(__name__)
+autoint = functools.partial(int, base=0)
+hex = functools.partial(int, base=16)
+
+
+def import_xdd(
+    xdd: Union[str, TextIO, None],
+    node_id: Optional[int],
+) -> ObjectDictionary:
+    od = ObjectDictionary()
+
+    if xdd is not None:
+        root = etree.parse(xdd).getroot()
+
+    if node_id is None:
+        device_commissioning_elem = root.find('.//{*}DeviceCommissioning')
+        if device_commissioning_elem is not None:
+            node_id_attr = device_commissioning_elem.get("nodeID")
+            if node_id_attr is None:
+                raise ValueError(
+                    "Missing required nodeID attribute")
+            od.node_id = int(node_id_attr, 0)
+        else:
+            od.node_id = None
+    else:
+        od.node_id = node_id
+
+    _add_device_information(od, root)
+    _add_object_list(od, root)
+    _add_dummy_objects(od, root)
+    return od
+
+
+def _add_device_information(
+    od: ObjectDictionary,
+    root: etree.Element
+):
+    device_identity = root.find('.//{*}DeviceIdentity')
+    if device_identity is None:
+        raise ValueError("Missing 'DeviceIdentity' section in XDD file")
+    if device_identity is not None:
+        identity_fields: list[tuple[str, str, Callable[[str], Any]]] = [
+            ("vendorName", "vendor_name", str),
+            ("vendorID", "vendor_number", hex),
+            ("productName", "product_name", str),
+            ("productID", "product_number", hex),
+        ]
+        for src_prop, dst_prop, f in identity_fields:
+            val = device_identity.find(f'{{*}}{src_prop}')
+            if val is not None and val.text:
+                setattr(od.device_information, dst_prop, f(val.text))
+
+    general_features = root.find('.//{*}CANopenGeneralFeatures')
+    if general_features is None:
+        raise ValueError("Missing 'CANopenGeneralFeatures' element")
+    if general_features is not None:
+        features_fields: list[tuple[str, str, Callable[[str], Any], Any]] = [
+            # properties without default value (default=None) are required
+            ("granularity", "granularity", autoint, None),
+            ("nrOfRxPDO", "nr_of_RXPDO", autoint, "0"),
+            ("nrOfTxPDO", "nr_of_TXPDO", autoint, "0"),
+            ("bootUpSlave", "simple_boot_up_slave", bool, 0),
+        ]
+        for src_prop, dst_prop, f, default in features_fields:
+            val = general_features.get(src_prop, default)
+            if val is None:
+                raise ValueError(
+                    f"Missing required '{src_prop}' property")
+            setattr(od.device_information, dst_prop, f(val))
+
+    baud_rate = root.find('.//{*}PhysicalLayer/{*}baudRate')
+    if baud_rate is None:
+        raise ValueError("Missing 'PhysicalLayer/baudRate' section")
+
+    for baud in baud_rate:
+        try:
+            baud_value = baud.get("value")
+            if baud_value is not None:
+                rate = int(baud_value.replace(' Kbps', ''), 10) * 1000
+                od.device_information.allowed_baudrates.add(rate)
+        except (ValueError, TypeError):
+            pass
+
+    if default_baud := baud_rate.get('defaultValue', '250 Kbps'):
+        try:
+            od.bitrate = int(default_baud.replace(' Kbps', ''), 10) * 1000
+        except (ValueError, TypeError):
+            pass
+
+
+def _add_object_list(
+    od: ObjectDictionary,
+    root: etree.Element
+):
+    # Process all CANopen objects in the file
+    for obj in root.findall('.//{*}CANopenObjectList/{*}CANopenObject'):
+        name = obj.get('name', '')
+        index = int(obj.get('index', '0'), 16)
+        object_type = int(obj.get('objectType', '0'))
+        sub_number = obj.get('subNumber')
+
+        # Simple variable
+        if object_type == objectcodes.VAR:
+            unique_id_ref = obj.get('uniqueIDRef', None)
+            parameters = root.find(
+                f'.//{{*}}parameter[@uniqueID="{unique_id_ref}"]')
+
+            var = _build_variable(parameters, od.node_id, name, index)
+            _set_parameters_from_xdd_canopen_object(od.node_id, var, obj)
+            od.add_object(var)
+
+        # Array
+        elif object_type == objectcodes.ARRAY and sub_number:
+            array = ODArray(name, index)
+            for sub_obj in obj:
+                sub_index_attr = sub_obj.get('subIndex')
+                if sub_index_attr is None:
+                    raise ValueError(
+                        "Missing 'subIndex' attribute for"
+                        " sub-object in array object"
+                        f" 0x{index:04X}")
+                sub_index = int(sub_index_attr, 16)
+                sub_name = sub_obj.get('name', '')
+                sub_unique_id = sub_obj.get('uniqueIDRef', None)
+                sub_parameters = root.find(
+                    f'.//{{*}}parameter[@uniqueID="{sub_unique_id}"]')
+
+                sub_var = _build_variable(
+                    sub_parameters, od.node_id, sub_name, index, sub_index)
+                _set_parameters_from_xdd_canopen_object(
+                    od.node_id, sub_var, sub_obj)
+                array.add_member(sub_var)
+            od.add_object(array)
+
+        # Record/Struct
+        elif object_type == objectcodes.RECORD and sub_number:
+            record = ODRecord(name, index)
+            for sub_obj in obj:
+                sub_index_attr = sub_obj.get('subIndex')
+                if sub_index_attr is None:
+                    raise ValueError(
+                        "Missing 'subIndex' attribute for"
+                        " sub-object in record object"
+                        f" 0x{index:04X}")
+                sub_index = int(sub_index_attr, 16)
+                sub_name = sub_obj.get('name', '')
+                sub_unique_id = sub_obj.get('uniqueIDRef', None)
+                sub_parameters = root.find(
+                    f'.//{{*}}parameter[@uniqueID="{sub_unique_id}"]')
+                sub_var = _build_variable(
+                    sub_parameters, od.node_id, sub_name, index, sub_index)
+                _set_parameters_from_xdd_canopen_object(
+                    od.node_id, sub_var, sub_obj)
+                record.add_member(sub_var)
+            od.add_object(record)
+
+
+def _add_dummy_objects(
+    od: ObjectDictionary,
+    root: etree.Element
+):
+    dummy_section = root.find('.//{*}ApplicationLayers/{*}dummyUsage')
+    if dummy_section is None:
+        return
+
+    for dummy in dummy_section:
+        entry = dummy.get('entry')
+        if entry is None:
+            raise ValueError(
+                "Missing 'entry' attribute for dummy object")
+        p = entry.split('=')
+        key = p[0]
+        value = int(p[1], 10)
+        index = int(key.replace('Dummy', ''), 10)
+        if value == 1:
+            var = ODVariable(key, index, 0)
+            var.data_type = index
+            var.access_type = "const"
+            od.add_object(var)
+
+
+def _set_parameters_from_xdd_canopen_object(
+    node_id: Optional[int],
+    dst: ODVariable,
+    src: etree.Element
+):
+    # PDO mapping of the object, optional, string
+    # Valid values:
+    # * no - not mappable
+    # * default - mapped by default
+    # * optional - optionally mapped
+    # * TPDO - may be mapped into TPDO only
+    # * RPDO - may be mapped into RPDO only
+    pdo_mapping_attr = src.get('PDOmapping')
+    if pdo_mapping_attr is not None:
+        pdo_mapping = pdo_mapping_attr.lower()
+        dst.pdo_mappable = pdo_mapping != 'no'
+
+    # Name of the object, optional, string
+    if var_name := src.get('name', None):
+        dst.name = var_name
+
+    # CANopen data type (two hex digits), optional
+    # data_type matches canopen library, no conversion needed
+    var_data_type_attr = src.get('dataType', None)
+    if var_data_type_attr is not None:
+        dst.data_type = int(var_data_type_attr, 16)
+
+    # Access type of the object; valid values, optional, string
+    # * const - read access only; the value is not changing
+    # * ro - read access only
+    # * wo - write access only
+    # * rw - both read and write access
+    # strings match with access_type in canopen library, no conversion needed
+    access_type_attr = src.get('accessType', None)
+    if access_type_attr is not None:
+        dst.access_type = access_type_attr
+
+    if dst.data_type in datatypes.INTEGER_TYPES:
+        # Low limit of the parameter value, optional, string
+        min_value_attr = src.get('lowLimit', None)
+        if min_value_attr is not None:
+            dst.min = _convert_integer(node_id, dst.data_type, min_value_attr)
+
+        # High limit of the parameter value, optional, string
+        max_value_attr = src.get('highLimit', None)
+        if max_value_attr is not None:
+            dst.max = _convert_integer(node_id, dst.data_type, max_value_attr)
+
+        # Default value of the object, optional, string
+        default_value_attr = src.get('defaultValue')
+        if default_value_attr is not None:
+            if '$NODEID' in default_value_attr:
+                dst.relative = True
+            dst.default = _convert_integer(
+                node_id, dst.data_type, default_value_attr)
+
+
+def _build_variable(
+    par_tree: Optional[etree.Element],
+    node_id: Optional[int],
+    name: str,
+    index: int,
+    subindex: int = 0
+) -> ODVariable:
+    var = ODVariable(name, index, subindex)
+    # Set default parameters
+    var.access_type = 'ro'
+    if par_tree is None:
+        return var
+
+    var.description = par_tree.get('description', '')
+
+    # Extract data type
+    data_types = {
+        'BOOL': datatypes.BOOLEAN,
+        'SINT': datatypes.INTEGER8,
+        'INT': datatypes.INTEGER16,
+        'DINT': datatypes.INTEGER32,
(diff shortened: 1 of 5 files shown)


acolomb:
Thank you very much for this contribution.  Happy to see some progress on the XDD front.  I started taking a look and overall I think it's going in a good direction.

Need to take some more time to put down my thoughts in detail, but some high-level pointers first:

* For new code, we expect to have full type annotations.  The mypy linter run in the PR checks can give you hints about what needs improvement, or best run it locally.  Note that any functions without argument or return type annotations are not even checked by default.
* I don't like the idea of inflating the `canopen.utils` module, even though avoiding code duplication is a worthy goal.  I'll try to come up with a simplification soon.
* Where I do want less duplication is the object code constants declared in both eds.py and xdd.py.  Have a look at commits 7c6dee5a1b040a3de8ea18263f7d8dad9120e076 and db9110e9cd0545b2aa7426243f169d0f863a7640 in master, those can be cherry-picked easily.
* A little more attention to code formatting would be nice.  I use flake8 locally and try to keep it from complaining.  Will report any outstanding issues during more thorough review.
* I assume there is no copyright involved for the `sample.xdd` file?  I don't know where the CANopenNode project got its references from, but just trying to be careful about legal issues with files we include here.

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
+import logging
+
+import re
+import xml.etree.ElementTree as etree
+from configparser import NoOptionError
+from typing import TYPE_CHECKING
```
acolomb:
```suggestion
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+    _add_dummy_objects_to_od(od, root)
+
+
+    return od
+
+def _add_device_information_to_od(od, root):
+    device_identity = root.find('.//{*}DeviceIdentity')
+    if device_identity is not None:
+        for src_prop, dst_prop, f in [
+            ("vendorName", "vendor_name", lambda val: str(val)),
```
acolomb:
```suggestion
            ("vendorName", "vendor_name", str),
```
This can be simplified in many places, no need for a lambda.

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+            ("vendorID", "vendor_number", lambda val: int(val, 0)),
```
acolomb: You could create a `autoint = functools.partial(int, base=0)` constructor once in this module, then use it to simplify in these cases.

acolomb (commented):
Just a partial review again, but I hope you get the general idea... More to come when I get another look with a fresher brain.

tomasbarton-stemcell:
hi @zaporozhets, thanks for the work on this PR! Are you still planning to continue with it? I have a project where I could really use xdd/xdc parsing right now, so I'd be glad to pick up where you left off and contribute to this effort if you don't have the bandwidth.

zaporozhets:
@acolomb, thank you for the review, and @tomasbarton-stemcell for your interest in this feature.

I was extremely busy last year and didn’t have the chance to address all the comments.

I’m currently on sick leave until Monday, so I finally have some time to catch up. Please give me a few days to refresh my memory and go through all the suggestions. I’ll keep you posted. I’ll definitely appreciate a thorough review here, as Python isn’t really my main area of expertise.

BR,
Taras

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
+import logging
+
+import re
+import xml.etree.ElementTree as etree
+from configparser import NoOptionError
```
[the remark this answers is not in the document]
zaporozhets: yup. thanks, fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
+import logging
+
```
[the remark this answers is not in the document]
zaporozhets: fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+import xml.etree.ElementTree as etree
+from configparser import NoOptionError
+from typing import TYPE_CHECKING
+
+from canopen import objectdictionary
+from canopen.objectdictionary import ObjectDictionary
+from canopen.utils import signed_int_from_hex, calc_bit_length
+
+if TYPE_CHECKING:
+    import canopen.network
```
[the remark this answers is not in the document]
zaporozhets: fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Object type. Don't confuse with Data type
+VAR = 7
+ARR = 8
+RECORD = 9
+
+
+def import_xdd(xdd, node_id):
```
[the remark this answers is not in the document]
zaporozhets: fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+                    pass
+    else:
+        od.node_id = node_id
+
+    _add_device_information_to_od(od, root)
+    _add_object_list_to_od(od, root)
+    _add_dummy_objects_to_od(od, root)
+
+
+    return od
```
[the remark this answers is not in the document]
zaporozhets: fixed

canopen/objectdictionary/xdd.py
zaporozhets: fixed

canopen/objectdictionary/xdd.py
zaporozhets: fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+
+def _add_device_information_to_od(od, root):
```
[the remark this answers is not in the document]
zaporozhets: fixed

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,306 @@
...
+        root = etree.parse(xdd).getroot()
+
+    if node_id is None:
+        device_commissioning = root.find('.//{*}DeviceCommissioning')
+        if device_commissioning is not None:
+            if node_id := device_commissioning.get('nodeID', None):
+                try:
+                    od.node_id = int(node_id, 0)
+                except (ValueError, TypeError):
+                    pass
```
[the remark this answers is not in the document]
zaporozhets: refactored

canopen/objectdictionary/xdd.py
zaporozhets: fixed

zaporozhets:
Thank you for your patience. I’ve addressed (hopefully all) the comments and refactored several sections to simplify the parsing logic.

1.Added full type annotations and verified the code with `mypy --config-file pyproject.toml ....`
2. Removed changes from `canopen.utils`.
3. Applied updates from `esd.py` to reduce duplicated code.
4. Fixed formatting and validated it with `flake8`.
5. Created a new `sample.xdd` based on `sample.eds`, without relying on third-party EDS/XDD files.
6. Squashed changes in one commit.

@acolomb, @tomasbarton-stemcell could you please review?

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+                    "Missing required nodeID attribute")
+            od.node_id = int(node_id_attr, 0)
+        else:
+            od.node_id = None
+    else:
+        od.node_id = node_id
+
+    _add_device_information(od, root)
+    _add_object_list(od, root)
+    _add_dummy_objects(od, root)
```
tomasbarton-stemcell, line 44: if `xdd` is `None` (which it can be according to the type signature) then this will throw here because root won't be defined.

thisone `root.find('.//{*}DeviceCommissioning')` as well

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    data_types = {
+        'BOOL': datatypes.BOOLEAN,
+        'SINT': datatypes.INTEGER8,
+        'INT': datatypes.INTEGER16,
+        'DINT': datatypes.INTEGER32,
+        'LINT': datatypes.INTEGER64,
+        'USINT': datatypes.UNSIGNED8,
+        'UINT': datatypes.UNSIGNED16,
+        'UDINT': datatypes.UNSIGNED32,
+        'ULINT': datatypes.UNSIGNED32,
```
tomasbarton-stemcell, line 278: is this right here? `LINT` above is defined to be 64b, but `ULINT` here is 32b

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+        else:
+            return int(re.sub(r'\+?\$NODEID\+?', '', value), 0) + node_id
+    else:
+        if var_type in datatypes.SIGNED_TYPES:
+            return _signed_int_from_hex(value, _calc_bit_length(var_type))
+        else:
+            return int(value, 0)
+
+
+def _convert_variable(
```
tomasbarton-stemcell, line 396: unused. Didn't check other files (shouldn't be as this is _), but there is no reference to it in this file.

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def _add_device_information(
+    od: ObjectDictionary,
+    root: etree.Element
+):
+    device_identity = root.find('.//{*}DeviceIdentity')
+    if device_identity is None:
+        raise ValueError("Missing 'DeviceIdentity' section in XDD file")
+    if device_identity is not None:
```
tomasbarton-stemcell, line 55: redundant None check (statement above guarantees this is not None).

There is the same problem below: `if general_features is not None:`

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    general_features = root.find('.//{*}CANopenGeneralFeatures')
+    if general_features is None:
+        raise ValueError("Missing 'CANopenGeneralFeatures' element")
+    if general_features is not None:
+        features_fields: list[tuple[str, str, Callable[[str], Any], Any]] = [
+            # properties without default value (default=None) are required
+            ("granularity", "granularity", autoint, None),
+            ("nrOfRxPDO", "nr_of_RXPDO", autoint, "0"),
+            ("nrOfTxPDO", "nr_of_TXPDO", autoint, "0"),
+            ("bootUpSlave", "simple_boot_up_slave", bool, 0),
```
tomasbarton-stemcell, line 76: is `bool` going to work here? I am not sure what the xml parser gives you, but if it is a string "true" or "false" they will both get converted to True using the `bool` function.

I was just going by the example xdd you have in test: `<CANopenGeneralFeatures granularity="8" nrOfRxPDO="4" nrOfTxPDO="4" bootUpSlave="true" />`

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    value: Optional[str]
+) -> Optional[int]:
+    if value is None:
+        return None
+    # COB-ID can contain '$NODEID+' so replace this with node_id
+    # before converting
+    value = value.replace(" ", "").upper()
+    if '$NODEID' in value:
+        if node_id is None:
+            logger.warn(
```
tomasbarton-stemcell, line 384: `logger.warn` is deprecated: https://docs.python.org/3/library/logging.html#logging.Logger.warning

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    var = ODVariable(name, index, subindex)
+    # Set default parameters
+    var.access_type = 'ro'
+    if par_tree is None:
+        return var
+
+    var.description = par_tree.get('description', '')
+
+    # Extract data type
+    data_types = {
```
tomasbarton-stemcell, line 269: Could probably be a file level constant

tomasbarton-stemcell (commented):
Hi, thanks for the update! I tested this on a few xdc and xdd files I have in my project (all generated by the editor that comes with the canopennode C library)  and all that I tried worked fine. I added some comments, but I am not really a Python programmer so 🤷🏻‍♂️ 

As a side not, could the `import_od` be relaxed to load xdc files as well? I am not exactly sure what the difference is, but with the files we have I can load them just fine and talk to the devices.

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+) -> ObjectDictionary:
+    od = ObjectDictionary()
+
+    if xdd is not None:
+        root = etree.parse(xdd).getroot()
+
+    if node_id is None:
+        device_commissioning_elem = root.find('.//{*}DeviceCommissioning')
+        if device_commissioning_elem is not None:
+            node_id_attr = device_commissioning_elem.get("nodeID")
```
tomasbarton-stemcell, line 32: One more thing I noticed with my xdc file from CANopenNode editor:

```
<deviceCommissioning NodeID="2" nodeName="scale" actualBaudRate="0" networkNumber="0" networkName="" CANopenManager="false" />
```

The casing is different, it is lower case `deviceCommissioning` and then upper `case NodeID`. I am not sure which is is correct according to the standard.

bizfsc (commented):
Automated review based on static analysis of the PR diff. Several correctness bugs need fixing before merge; design issues are suggestions.

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    ODRecord,
+    ODVariable,
+    ObjectDictionary,
+    datatypes,
+    objectcodes,
+)
+
+logger = logging.getLogger(__name__)
+autoint = functools.partial(int, base=0)
+hex = functools.partial(int, base=16)
```
bizfsc, line 17: **Bug: `hex` shadows the `builtins.hex` function.**

`hex` is a Python builtin. Redefining it here makes it unavailable for any downstream code in this module and is confusing.

Suggestion: rename to `parse_hex` or `_from_hex`.

```python
parse_hex = functools.partial(int, base=16)
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def import_xdd(
+    xdd: Union[str, TextIO, None],
+    node_id: Optional[int],
+) -> ObjectDictionary:
+    od = ObjectDictionary()
+
+    if xdd is not None:
```
bizfsc, line 26: **Bug: `NameError` when `xdd=None`.**

If `xdd is None`, `root` is never assigned, but it is used immediately at line 30 (`root.find('.//{*}DeviceCommissioning')`) and again at lines 42-44.

Either raise early or guard all usages:

```python
if xdd is None:
    raise ValueError("source must not be None")
root = etree.parse(xdd).getroot()
```

canopen/objectdictionary/xdd.py
bizfsc, line 55: **Dead branch: `if device_identity is not None:` is always `True` here.**

The `raise ValueError` on the line above guarantees `device_identity is not None` at this point. The `if` guard is misleading — it implies the block could be skipped. Just dedent the body:

```python
if device_identity is None:
    raise ValueError("Missing 'DeviceIdentity' section in XDD file")
identity_fields = [...]
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+        ]
+        for src_prop, dst_prop, f in identity_fields:
+            val = device_identity.find(f'{{*}}{src_prop}')
+            if val is not None and val.text:
+                setattr(od.device_information, dst_prop, f(val.text))
+
+    general_features = root.find('.//{*}CANopenGeneralFeatures')
+    if general_features is None:
+        raise ValueError("Missing 'CANopenGeneralFeatures' element")
+    if general_features is not None:
```
bizfsc, line 70: **Same dead-branch issue as with `device_identity`.**

After `raise ValueError("Missing 'CANopenGeneralFeatures' element")`, the `if general_features is not None:` guard is always `True`. Dedent the body.

canopen/objectdictionary/xdd.py
bizfsc, line 76: **Bug: `bool("false") == True` in Python.**

The XDD attribute value is the string `"true"` or `"false"`. Passing it to `bool(...)` will always return `True` for any non-empty string, including `"false"`.

Fix:

```python
("bootUpSlave", "simple_boot_up_slave", lambda s: str(s).lower() == "true", "false"),
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+        name = obj.get('name', '')
+        index = int(obj.get('index', '0'), 16)
+        object_type = int(obj.get('objectType', '0'))
+        sub_number = obj.get('subNumber')
+
+        # Simple variable
+        if object_type == objectcodes.VAR:
+            unique_id_ref = obj.get('uniqueIDRef', None)
+            parameters = root.find(
+                f'.//{{*}}parameter[@uniqueID="{unique_id_ref}"]')
```
bizfsc, line 120: **Potential XPath injection / parse error via unsanitised `unique_id_ref`.**

`unique_id_ref` comes directly from an XML attribute. If it contains `"` or `]`, the constructed XPath `'[@uniqueID="..."]'` will either raise a parse error or silently match the wrong element.

The same pattern is repeated for `sub_unique_id` in the ARRAY and RECORD branches.

Mitigation: validate/escape the value, or use a two-step approach (findall + attribute check in Python):

```python
parameters = next(
    (p for p in root.findall('.//{*}parameter')
     if p.get('uniqueID') == unique_id_ref),
    None,
)
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+
+            var = _build_variable(parameters, od.node_id, name, index)
+            _set_parameters_from_xdd_canopen_object(od.node_id, var, obj)
+            od.add_object(var)
+
+        # Array
+        elif object_type == objectcodes.ARRAY and sub_number:
```
bizfsc, line 127: **Silently skips ARRAY/RECORD objects that have no `subNumber` attribute.**

If `sub_number` is `None` (missing attribute), the whole `elif` is skipped without any warning. The object is silently dropped from the OD. The same applies to the RECORD branch.

Add a `logger.warning(...)` for the unexpected case:

```python
elif object_type == objectcodes.ARRAY:
    if not sub_number:
        logger.warning("ARRAY 0x%04X has no subNumber, skipping", index)
        continue
    ...
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+    dummy_section = root.find('.//{*}ApplicationLayers/{*}dummyUsage')
+    if dummy_section is None:
+        return
+
+    for dummy in dummy_section:
+        entry = dummy.get('entry')
+        if entry is None:
+            raise ValueError(
+                "Missing 'entry' attribute for dummy object")
+        p = entry.split('=')
```
bizfsc, line 185: **`IndexError` if `entry` contains no `=` character.**

`entry.split('=')` returns a list with one element when `=` is absent; `p[1]` then raises `IndexError`.

Safer:

```python
key, _, raw_value = entry.partition('=')
if not _:
    logger.warning("Unexpected dummyUsage entry format: %r", entry)
    continue
value = int(raw_value, 10)
```

canopen/objectdictionary/xdd.py
```diff
@@ -0,0 +1,413 @@
...
+        'ULINT': datatypes.UNSIGNED32,
+        'REAL': datatypes.REAL32,
+        'LREAL': datatypes.REAL64,
+        'STRING': datatypes.VISIBLE_STRING,
+        'BITSTRING': datatypes.DOMAIN,
+        'WSTRING': datatypes.UNICODE_STRING
+    }
+
+    for k, v in data_types.items():
+        if par_tree.find(f'{{*}}{k}') is not None:
```
bizfsc, line 287: **Last-wins data-type lookup — missing `break`.**

The loop iterates over all keys even after a match is found. If a `<parameter>` element happened to contain two type tags (malformed XDD), the last one in the dict wins silently.

Add `break` after setting `var.data_type`:

```python
for k, v in data_types.items():
    if par_tree.find(f'{{*}}{k}') is not None:
        var.data_type = v
        break
```

canopen/objectdictionary/xdd.py
bizfsc, line 384: **`logger.warn()` is deprecated since Python 3.2.**

Use `logger.warning(...)` instead.

canopen/objectdictionary/xdd.py
bizfsc, line 396: **`_convert_variable` is defined but never called (dead code).**

It handles `OCTET_STRING`, `DOMAIN`, `VISIBLE_STRING`, `FLOAT_TYPES` etc., but none of the callers use it — `_build_variable` and `_set_parameters_from_xdd_canopen_object` only call `_convert_integer`. This means default values for non-integer types (e.g. `REAL32`, `VISIBLE_STRING`) are silently ignored.

Either wire up `_convert_variable` for non-integer defaults, or remove it.

test/test_xdd.py
```diff
@@ -0,0 +1,171 @@
...
+        "int24": [
+            {"hex_str": "7FFFFF", "bit_length": 24, "expected": 8388607},
+            {"hex_str": "800000", "bit_length": 24, "expected": -8388608},
+            {"hex_str": "FFFFFF", "bit_length": 24, "expected": -1},
+            {"hex_str": "000000", "bit_length": 24, "expected": 0},
+            {"hex_str": "000001", "bit_length": 24, "expected": 1}
+        ],
+        "int32": [
+            {"hex_str": "7FFFFFFF", "bit_length": 32, "expected": 2147483647},
+            {"hex_str": "80000000", "bit_length": 32, "expected": -2147483648},
```
bizfsc, line 36: **Test missing: `bool("false")` bug.**

No test checks that `bootUpSlave="false"` (string) is parsed as `False`. This is the easiest way to catch the `bool(str)` bug in CI.

```python
def test_boot_up_slave_false(self):
    # sample.xdd has bootUpSlave="true"
    self.assertTrue(self.od.device_information.simple_boot_up_slave)
```
And a complementary file with `bootUpSlave="false"` should assert `False`.

PR #615, opened 2025-09-21
Custom Options for EDS file

FedericoSpada:
Vector CANeds tool groups all unknown options contained in an EDS File in a dedicated section called "Unknown Entries".
We use that section to save additional information that is not considered by the CANopen specifications (for example: unit, factor, offset, category, etc.), and so I had to modify this library to read and write those options.

If you open a file that contains unknown option names, they will be grouped in the "custom_options" dictionary.
If you desire to save a value with a specific option name, you can update the "custom_options" dictionary accordingly.

P.S.: In the meantime, I've improved some of the bad code.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -209,6 +209,8 @@ class ODRecord(MutableMapping):
         self.storage_location = None
         self.subindices = {}
         self.names = {}
+        #: Key-Value pairs not defined by the standard
+        self.custom_options = {}
 
     def __repr__(self) -> str:
         return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
@@ -268,6 +270,8 @@ class ODArray(Mapping):
         self.storage_location = None
         self.subindices = {}
         self.names = {}
+        #: Key-Value pairs not defined by the standard
+        self.custom_options = {}
 
     def __repr__(self) -> str:
         return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
@@ -374,6 +378,8 @@ class ODVariable:
         self.storage_location = None
         #: Can this variable be mapped to a PDO
         self.pdo_mappable = False
+        #: Key-Value pairs not defined by the standard
+        self.custom_options = {}
 
     def __repr__(self) -> str:
         subindex = self.subindex if isinstance(self.parent, (ODRecord, ODArray)) else None
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -131,6 +131,7 @@ def import_eds(source, node_id):
 
             if object_type in (VAR, DOMAIN):
                 var = build_variable(eds, section, node_id, index)
+                var.custom_options = _get_custom_options(eds, section)
                 od.add_object(var)
             elif object_type == ARR and eds.has_option(section, "CompactSubObj"):
                 arr = objectdictionary.ODArray(name, index)
@@ -140,14 +141,17 @@ def import_eds(source, node_id):
                 arr.add_member(last_subindex)
                 arr.add_member(build_variable(eds, section, node_id, index, 1))
                 arr.storage_location = storage_location
+                arr.custom_options = _get_custom_options(eds, section)
                 od.add_object(arr)
             elif object_type == ARR:
                 arr = objectdictionary.ODArray(name, index)
                 arr.storage_location = storage_location
+                arr.custom_options = _get_custom_options(eds, section)
                 od.add_object(arr)
             elif object_type == RECORD:
                 record = objectdictionary.ODRecord(name, index)
                 record.storage_location = storage_location
+                record.custom_options = _get_custom_options(eds, section)
                 od.add_object(record)
 
             continue
@@ -253,6 +257,18 @@ def _revert_variable(var_type, value):
     else:
         return f"0x{value:02X}"
 
+_STANDARD_OPTIONS = ["ObjectType"      , "ParameterName" , "DataType"    , "AccessType"   ,
+                     "PDOMapping"      , "LowLimit"      , "HighLimit"   , "DefaultValue" ,
+                     "ParameterValue"  , "Factor"        , "Description" , "Unit"         ,
+                     "StorageLocation" , "CompactSubObj"                                  ]
+
+def _get_custom_options(eds, section):
+    custom_options = {}
+    for option, value in eds.items(section):
+        if option not in _STANDARD_OPTIONS:
+            custom_options[option] = value
+    return custom_options
+
 
 def build_variable(eds, section, node_id, index, subindex=0):
     """Creates a object dictionary entry.
@@ -332,6 +348,8 @@ def build_variable(eds, section, node_id, index, subindex=0):
             var.unit = eds.get(section, "Unit")
         except ValueError:
             pass
+
+    var.custom_options = _get_custom_options(eds, section)
     return var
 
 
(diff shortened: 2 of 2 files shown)


bizfsc:
Thanks for adding custom options support! I reviewed the changes and have a few observations:

**1. Duplicate `_get_custom_options()` call for VAR types**

For `VAR` / `DOMAIN` objects, `_get_custom_options()` is called twice:
- Once in `import_eds()` (line 134): `var.custom_options = _get_custom_options(eds, section)`
- Once at the end of `build_variable()` (line 352): `var.custom_options = _get_custom_options(eds, section)`

The second call overwrites the first. One of them should be removed. Since `build_variable()` is also called for subindexes, keeping it only there seems cleanest – then the calls in `import_eds()` for ARR/RECORD containers still make sense (they parse the parent section's custom options).

**2. `_STANDARD_OPTIONS` is incomplete**

The list is missing at least:
- `"NrOfEntries"` (used in `[index]Name` sections and `SubNumber` calculation)
- `"SubNumber"` (used in `export_record`)
- `"ObjFlags"` and `"Denotation"` (valid per CiA 306)
- Numeric keys like `"0"`, `"1"`, etc. (used as subindex names in CompactSubObj sections)

Any standard option not in this list will incorrectly end up in `custom_options`. Consider also making the comparison case-insensitive (e.g. using a lowercase set) as a safety measure, even though `optionxform = str` is set.

**3. No tests**

It would be great to add a test that:
- Imports an EDS with a non-standard option (e.g. `Category=MyCategory`) on a variable
- Verifies it appears in `var.custom_options` and standard options do not
- Exports back to EDS and verifies the custom option is preserved

**4. Typo fix should be a separate commit**

The `manufacturer_idices` → `manufacturer_indices` rename is a good fix but is unrelated to the custom options feature. Splitting it into its own commit would keep the history clean.

FedericoSpada:
Hi,
Regarding the absence of "NrOfEntries" and "SubNumber", I've found the problem myself, but I didn't report it here since no one has been interested in this PR for 6 months.
For "ObjFlags" and "Denotation", since they are not managed in this Library, maybe we can make them appear in "custom_options", so the user can still access their value. What do you think about it?
For the tests, could you please write them yourself after my next commit? I've never used the "unittest" library, and testing in general is not my cup of tea. Thanks in advance.

bizfsc:
@FedericoSpada no problem at all, I created a new PR (#653) because I can not edit your fork.
I split the whole thing into three separate PRs (#642, already merged; #653 open, #654 open).

acolomb:
Sorry for not reacting earlier, there is definitely interest in your work @FedericoSpada.  Just not enough spare time on my side to work on the backlog regarding this library.  I'm planning on it though in the next few days.  Good to see @bizfsc picking it up and pushing it forward.

FedericoSpada:
I've implemented the requested changes.

canopen/objectdictionary/__init__.py
```diff
@@ -209,6 +209,8 @@ def __init__(self, name: str, index: int):
         self.storage_location = None
         self.subindices = {}
         self.names = {}
+        #: Key-Value pairs not defined by the standard
+        self.custom_options = {}
```
acolomb, line 213: For new code, we require proper typing hints now, to avoid an ever-growing error list from the mypy checker.  (Same thing below.)


```suggestion
        self.custom_options: dict[str, str] = {}
```

canopen/objectdictionary/eds.py
```diff
@@ -253,6 +256,19 @@ def _revert_variable(var_type, value):
     else:
         return f"0x{value:02X}"
 
+_STANDARD_OPTIONS = ["objecttype"      , "parametername" , "datatype"    , "accesstype"   ,
+                     "pdomapping"      , "lowlimit"      , "highlimit"   , "defaultvalue" ,
+                     "parametervalue"  , "factor"        , "description" , "unit"         ,
+                     "storagelocation" , "compactsubobj" , "nrofentries" , "subnumber"    ,
+                     "objflags"        , "denotation"                                     ]
+
```
acolomb, line 264: This is definitely funky formatting.  We mostly follow "black" style, though not strictly enforded.  A simple list, one item per row, would be fine here, however.  Makes it easier to insert grouping comments.

As for the name, I think `KNOWN_OBJECT_OPTIONS` would be a better fit, as some of these are not strictly standard.  And we don't strictly need to mark them "internal", thus moving toward the top of the file and removing the underscore seems appropriate.

This would be my approach, sorted as in CiA 306:

```suggestion
KNOWN_OBJECT_OPTIONS = [
    # EDS generic:
    "parametername",
    "objecttype",
    "datatype",
    "accesstype",
    "defaultvalue",
    "pdomapping",
    "subnumber",
    "lowlimit",
    "highlimit",
    "objflags",
    "compactsubobj",
    "nrofentries",
    # DCF only:
    "parametervalue",
    "uploadfile",
    "downloadfile",
    "denotation",
    # nonstandard extensions:
    "factor",
    "unit",
    "description",
    "storagelocation",
]
```

canopen/objectdictionary/eds.py
```diff
@@ -253,6 +256,19 @@ def _revert_variable(var_type, value):
...
+def _get_custom_options(eds, section):
```
acolomb, line 265: Type hints, please, for new code.

canopen/objectdictionary/eds.py
```diff
@@ -253,6 +256,19 @@ def _revert_variable(var_type, value):
...
+    custom_options = {}
+    for option, value in eds.items(section):
+        if not (option.lower() in _STANDARD_OPTIONS or option.isdigit()) :
```
acolomb, line 268: Please pay attention to code formatting.  The space before `:` violates PEP-8 for example.  Having some linters running in the background really helps to catch these problems early, so I recommend it (`flake8` and `black --diff` for example).

Same thing with the number of blank lines between top-level definitions, btw.

canopen/objectdictionary/__init__.py
```diff
@@ -209,6 +209,8 @@ def __init__(self, name: str, index: int):
         self.storage_location = None
         self.subindices = {}
         self.names = {}
+        #: Key-Value pairs not defined by the standard
```
acolomb, line 212: The library already uses some nonstandard options, so the description is a bit misleading.  We should call them "not handled explicitly" instead.

canopen/objectdictionary/eds.py
```diff
@@ -253,6 +256,19 @@ def _revert_variable(var_type, value):
...
+            custom_options[option] = value
```
acolomb, line 269: This whole function would be a good candidate for a dict comprehension instead of an explicit loop.

acolomb (changes requested):
Thank you very much for this contribution.  Structurally, I think it's alright and a useful addition.  There are some issues regarding code style though, especially the missing type hints.  While that could be easily fixed, I think #653 and #654 are already more extensive and a bit cleaner, so let's continue on those instead.

Still added some comments here as guidance, mostly also relevant for the other PRs.

canopen/objectdictionary/__init__.py
FedericoSpada: I always copy the format of existing code.
Maybe it is worth the effort to type-hint the whole library.

canopen/objectdictionary/eds.py
FedericoSpada: I think it would be too complicated/long:
```
{option: value for option, value in eds.items(section) if not (option.lower() in _STANDARD_OPTIONS or option.isdigit())}
```

acolomb:
Superseded by #653.  Thanks to @FedericoSpada for the initiative and to @bizfsc for pushing it over the finish line.

PR #621, opened 2025-12-09
network: Avoid notifier re-creating in connect function

xndcn:


xndcn:
ping, thanks

xndcn:
If we do not release notifier in disconnect, then when we connect again with the same Network object, it do not re-create the notifier with the new bus.

xndcn:
If network connect again without disconnect, current implementation will not re-create bus, but notifier will always be re-created. I think it is better to check the notifier in connect() and release it in disconnect, as well as the bus object.

bizfsc:
The fix is correct – without this patch, calling `connect()` again without a prior `disconnect()` always re-creates the notifier, leaking the old one. The guard on `self.bus` already existed, so this makes the notifier handling symmetrical. Nice catch!

Two suggestions:

**1. Simplify `try/except/finally` in `disconnect()`**

The `except Exception as e: raise e` is a no-op and can be removed. This simplifies to:

```python
try:
    self.check()
finally:
    self.notifier = None
```

**2. Add regression tests**

Consider adding tests to `test/test_network.py` to verify the new behavior:

```python
def test_network_connect_does_not_recreate_notifier(self):
    self.network.connect(interface="virtual")
    self.addCleanup(self.network.disconnect)
    notifier1 = self.network.notifier
    self.assertIsNotNone(notifier1)
    # Calling connect() again should reuse the existing notifier
    self.network.connect(interface="virtual")
    self.assertIs(self.network.notifier, notifier1)

def test_network_disconnect_releases_notifier(self):
    self.network.connect(interface="virtual")
    self.assertIsNotNone(self.network.notifier)
    self.network.disconnect()
    self.assertIsNone(self.network.notifier)

def test_network_disconnect_releases_notifier_on_exception(self):
    self.network.connect(interface="virtual")

    class Custom(Exception):
        pass

    self.network.notifier.exception = Custom("fake")
    with self.assertRaises(Custom):
        with self.assertLogs(level=logging.ERROR):
            self.network.disconnect()
    # Notifier must be released even when check() raises
    self.assertIsNone(self.network.notifier)
```

I verified locally that all three tests pass with the patched code.

xndcn:
@bizfsc Thank you for the thorough review!
Suggestion 1 – I amended the commit to simplify the try/except/finally in disconnect(). The redundant except Exception as e: raise e is now removed.
Suggestion 2 – I added the three regression tests to test/test_network.py and committed them with you as co-author.

bizfsc (approved):
Reviewed

friederschueler:
Hi @xndcn, thanks for the PR! I ran the tests locally and found one regression: the existing `test_network_check` fails with the new code.

**Root cause:** The cleanup function in `test_network_check` accesses `self.network.notifier.exception` after `disconnect()` has already set `self.notifier = None` (via the new `finally` block). This causes:

```
AttributeError: 'NoneType' object has no attribute 'exception'
```

**Fix:** Add a `None`-guard in the cleanup function in `test/test_network.py`:

```python
def cleanup():
    # We must clear the fake exception installed below, since
    # .disconnect() implicitly calls .check() during test tear down.
    if self.network.notifier is not None:
        self.network.notifier.exception = None
    self.network.disconnect()
```

With that one-line change all 19 tests pass. Could you add this to your branch?

xndcn:
@friederschueler Thanks, fixed.

friederschueler (approved):

PR #633, opened 2026-03-19
fix: handle KeyError in PdoMap __getitem__

alinenecchi:


acolomb:
This is so fundamentally wrong, on all possible levels. Sorry.

PR #640, opened 2026-04-16
pdo: Handle None cob_id correctly in __repr__

stoicfeats:
fxes a crash when calling repr() on a pdomap that doesn't have a cob_id assigned yet.

before this, working in an interactive session would throw a typeerror if we evaluated uninitialized maps like node.rpdo[1]. 
the string formatter was trying to apply : X hex-formatting to a nonetype value.

this adds a quick inline fallback so it prints "unassigned" instead of throwing error on the  console.

### DIFF
--- a/canopen/pdo/base.py
+++ b/canopen/pdo/base.py
@@ -229,7 +229,8 @@ class PdoMap:
         self._task = None
 
     def __repr__(self) -> str:
-        return f"<{type(self).__qualname__} {self.name!r} at COB-ID 0x{self.cob_id:X}>"
+        cob = f"0x{self.cob_id:X}" if self.cob_id is not None else "Unassigned"
+        return f"<{type(self).__qualname__} {self.name!r} at COB-ID {cob}>"
 
     def __getitem_by_index(self, value):
         valid_values = []
@@ -492,7 +493,7 @@ class PdoMap:
         associated with read() or save(), if the local PDO setup is
         known to match what's stored on the node.
         """
-        if self.enabled:
+        if self.enabled and self.cob_id is not None:
             logger.info("Subscribing to enabled PDO 0x%X on the network", self.cob_id)
             self.pdo_node.network.subscribe(self.cob_id, self.on_message)
 
@@ -540,6 +541,8 @@ class PdoMap:
 
     def transmit(self) -> None:
         """Transmit the message once."""
+        if self.cob_id is None:
+            raise ValueError("A valid COB-ID has not been configured")
         self.pdo_node.network.send_message(self.cob_id, self.data)
 
     def start(self, period: Optional[float] = None) -> None:
@@ -559,6 +562,8 @@ class PdoMap:
 
         if not self.period:
             raise ValueError("A valid transmission period has not been given")
+        if self.cob_id is None:
+            raise ValueError("A valid COB-ID has not been configured")
         logger.info("Starting %s with a period of %s seconds", self.name, self.period)
 
         self._task = self.pdo_node.network.send_periodic(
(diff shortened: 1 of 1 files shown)


stoicfeats:
@acolomb please have a look into this, thanks.

acolomb:
Hi and thanks for your contribution.  I've experienced the issue you mention several times.  It's usually a sign of weird ordering in the API usage, because you need to assign a valid COB ID by some means first.  But sure, the `__repr__` case is very easy to hit without even trying any actual bus configuration.

We have several more places that can fail with the same error, mainly logging calls but also `send_message()` and friends.  Ideally, we would forbid the `cob_id` member to ever be `None` (rather zero, which is also not valid).  But that's a rather large change.

Would you mind checking the other places though, to fix other occurrences where the format strings can easily explode with `cob_id = None`?

stoicfeats:
these were the places where `cob_id = None` could have cause the issues.

added explicit none-checks to subscribe, start, transmit, and remote_request to raise an immediate value error.

@acolomb

stoicfeats:
@acolomb can you please have a look into this once ?

canopen/pdo/base.py
```diff
@@ -229,7 +229,8 @@ def __init__(self, pdo_node, com_record, map_array):
         self._task = None
 
     def __repr__(self) -> str:
-        return f"<{type(self).__qualname__} {self.name!r} at COB-ID 0x{self.cob_id:X}>"
+        cob = f"0x{self.cob_id:X}" if self.cob_id is not None else "Unassigned"
```
acolomb:
```suggestion
        cob = f"0x{self.cob_id:X}" if self.cob_id else "Unassigned"
```

canopen/pdo/base.py
```diff
@@ -492,7 +493,7 @@ def subscribe(self) -> None:
         associated with read() or save(), if the local PDO setup is
         known to match what's stored on the node.
         """
-        if self.enabled:
+        if self.enabled and self.cob_id is not None:
```
acolomb:
```suggestion
        if self.enabled and self.cob_id:
```

canopen/pdo/base.py
```diff
@@ -540,6 +541,8 @@ def add_variable(
 
     def transmit(self) -> None:
         """Transmit the message once."""
+        if self.cob_id is None:
```
acolomb:
```suggestion
        if not self.cob_id:
```

canopen/pdo/base.py
```diff
@@ -559,6 +562,8 @@ def start(self, period: Optional[float] = None) -> None:
 
         if not self.period:
             raise ValueError("A valid transmission period has not been given")
+        if self.cob_id is None:
```
acolomb:
```suggestion
        if not self.cob_id:
```

canopen/pdo/base.py
```diff
@@ -579,7 +584,7 @@ def remote_request(self) -> None:
         """Send a remote request for the transmit PDO.
         Silently ignore if not allowed.
         """
-        if self.enabled and self.rtr_allowed:
+        if self.enabled and self.rtr_allowed and self.cob_id is not None:
```
acolomb:
```suggestion
        if self.enabled and self.rtr_allowed and self.cob_id:
```

acolomb (changes requested):
Looks good in general, thank you!

I just broadened the checks a bit to cover more than just `None` values. Zero is not a valid COB ID for PDOs either, so checking for the thruth value of `self.cob_id` catches both cases.

acolomb (approved):

stoicfeats:
Thank you for the review comments.
Will take care of these edge cases in upcoming PRs.
@acolomb

PR #643, opened 2026-04-28
Add unit tests for LSS master (coverage 42% → 93%)

bizfsc:
## Summary

Add 28 unit tests for `canopen/lss.py` to improve test coverage from 42% to 93%.

## Tests Added

- **Switch state commands**: global configuration/waiting mode, selective with match/no-match
- **Node ID configuration**: set node ID, error response handling
- **Bit timing configuration**: set bit timing, activate with delay
- **Store configuration**: success and error responses
- **LSS address inquiry**: vendor ID, product code, revision number, serial number
- **Timeout handling**: verify `LssError` raised on missing responses
- **Fast scan**: full protocol flow with binary search
- **Obsolete aliases**: `send_switch_state_global` backward compat
- **LssError**: exception instantiation and inheritance

## Approach

Uses `unittest.mock.MagicMock` to mock the network layer. A helper method injects CAN responses during `send_message()` calls to simulate real LSS slave behavior, including only responding to commands that expect a reply per the protocol.

## Coverage

```
canopen/lss.py    42% → 93%
Overall           77% → 80%
```


test/test_lss.py
```diff
@@ -0,0 +1,335 @@
...
+        self.network.send_message.side_effect = self._send_no_response
+        self.lss.send_switch_state_global(LssMaster.WAITING_STATE)
+        _, data = self.sent_messages[0]
+        self.assertEqual(data[0], CS_SWITCH_STATE_GLOBAL)
+        self.assertEqual(data[1], LssMaster.WAITING_STATE)
+
+    def test_send_switch_state_global_no_response_expected(self):
+        self.network.send_message.side_effect = self._send_no_response
+        self.lss.send_switch_state_global(LssMaster.CONFIGURATION_STATE)
+
```
acolomb: What's the use of this?  Already covered by `test_send_switch_state_global_configuration()`.

test/test_lss.py
```diff
@@ -0,0 +1,266 @@
...
+        self.lss.send_switch_state_selective(0xDEADBEEF, 0xCAFEBABE, 0x12345678, 0x9ABCDEF0)
+
+        data = self.sent_messages[0][1]
+        packed = struct.unpack_from('<I', data, 1)[0]
+        self.assertEqual(packed, 0xDEADBEEF)
+
+        data = self.sent_messages[1][1]
+        packed = struct.unpack_from('<I', data, 1)[0]
+        self.assertEqual(packed, 0xCAFEBABE)
+
```
acolomb: This is equivalent to what happens in `test_send_switch_state_selective_success()`, should be removed.

test/test_lss.py
```diff
@@ -0,0 +1,266 @@
...
+from canopen.lss import (
+    LssMaster,
+    LssError,
+    CS_INQUIRE_VENDOR_ID,
+    CS_INQUIRE_PRODUCT_CODE,
+    CS_INQUIRE_REVISION_NUMBER,
+    CS_INQUIRE_SERIAL_NUMBER,
+    ListMessageNeedResponse,
+)
```
acolomb: Imports should be sorted and in this case, I 'd prefer `from canopen import lss` with `lss.WHATEVER` instead for the constants.  The class names can stay here IMHO, but on one line.

acolomb (changes requested):
Good stuff, thank you!

As with other modules' tests, the raw byte sequence should be compared to catch wrong constants definitions, instead of using those from the module to be tested.  I reworked that accordingly and fixed some style nits.  Few comments remain, please have a look.

bizfsc (commented):
Thanks for reviewing, you are 100% correct and I also like the suggested import-style more.
Disclaimer: I used Claude Opus 4.6 to generate those tests after analysis and reviewed them manually.

acolomb:
> I used Claude Opus

So I guessed. :-)  But that's fine with me as long as the result is sane and reviewed by yourself.

acolomb (approved):

PR #644, opened 2026-04-28
Add unit tests for CiA 402 profile (coverage 36% → 45%)

bizfsc:
## Summary

Add 24 unit tests for `canopen/profiles/p402.py` to improve test coverage from 36% to 45%.

## Tests Added

- **State402 enum**: all states present, statusword decoding via bitmask
- **State decoding**: `State402.decode_statusword()` for all defined states
- **Command words**: verify controlword values for every state transition
- **Next-state calculation**: walk through the CiA 402 state machine graph
- **Homing status**: `IS_HOMING` / `HOMING_COMPLETED` / `HOMING_ERROR` evaluation
- **Operation mode**: switching and reading via OD entries 0x6060/0x6061
- **TPDO callback**: statusword update propagation through PDO mapping

## Approach

Uses a minimal `ObjectDictionary` with the required DS402 objects (0x6040 controlword, 0x6041 statusword, 0x6060/0x6061 operation mode, 0x6502 supported drives). PDO maps are mocked with `unittest.mock.MagicMock` to test the TPDO callback path without requiring a real CAN bus.

## Coverage

```
canopen/profiles/p402.py    36% → 45%
```

The remaining uncovered code is primarily in `BaseNode402` methods that require a live CAN bus interaction (SDO reads/writes, NMT state changes, PDO communication).

### DIFF
--- a/test/test_p402.py
+++ b/test/test_p402.py
@@ -108,12 +108,8 @@ class TestState402(unittest.TestCase):
         self.assertEqual(
             State402.next_state_indirect("SWITCH ON DISABLED"), "READY TO SWITCH ON"
         )
-        self.assertEqual(
-            State402.next_state_indirect("READY TO SWITCH ON"), "SWITCHED ON"
-        )
-        self.assertEqual(
-            State402.next_state_indirect("SWITCHED ON"), "OPERATION ENABLED"
-        )
+        self.assertEqual(State402.next_state_indirect("READY TO SWITCH ON"), "SWITCHED ON")
+        self.assertEqual(State402.next_state_indirect("SWITCHED ON"), "OPERATION ENABLED")
         self.assertEqual(State402.next_state_indirect("FAULT"), "SWITCH ON DISABLED")
         self.assertEqual(State402.next_state_indirect("FAULT REACTION ACTIVE"), "FAULT")
         self.assertEqual(
--- a/test/test_p402.py
+++ b/test/test_p402.py
@@ -1,9 +1,9 @@
 import unittest
 from unittest.mock import MagicMock
 
-from canopen.objectdictionary import ObjectDictionary, ODVariable
+from canopen.objectdictionary import ODVariable, ObjectDictionary
 from canopen.objectdictionary.datatypes import INTEGER8, UNSIGNED16, UNSIGNED32
-from canopen.profiles.p402 import BaseNode402, Homing, OperationMode, State402
+from canopen.profiles.p402 import BaseNode402, OperationMode, State402
 
 
 def _make_od():
--- a/test/test_p402.py
+++ b/test/test_p402.py
@@ -1,9 +1,9 @@
 import unittest
 from unittest.mock import MagicMock
 
-from canopen.objectdictionary import ObjectDictionary, ODVariable
-from canopen.objectdictionary.datatypes import UNSIGNED16, UNSIGNED32, INTEGER8
-from canopen.profiles.p402 import BaseNode402, State402, OperationMode, Homing
+from canopen.objectdictionary import ODVariable, ObjectDictionary
+from canopen.objectdictionary.datatypes import INTEGER8, UNSIGNED16, UNSIGNED32
+from canopen.profiles.p402 import BaseNode402, Homing, OperationMode, State402
 
 
 def _make_od():
--- a/test/test_p402.py
+++ b/test/test_p402.py
@@ -203,7 +203,7 @@ class TestBaseNode402State(unittest.TestCase):
 
     def test_check_statusword_no_tpdo_pointers(self):
         """check_statusword returns cached statusword when no TPDO pointers."""
-        self._inject_statusword(0x0027)
+        self._inject_statusword(0x0027)  # OPERATION ENABLED
         result = self.node.check_statusword()
         self.assertEqual(result, 0x0027)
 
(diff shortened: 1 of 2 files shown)


test/test_p402.py
```diff
@@ -0,0 +1,281 @@
...
+    def test_is_faulted_false(self):
+        self.node.tpdo_values[0x6041] = 0x0040  # SWITCH ON DISABLED
+        self.assertFalse(self.node.is_faulted())
+
+    def test_controlword_read_raises(self):
+        with self.assertRaises(RuntimeError):
+            _ = self.node.controlword
+
+
+class TestBaseNode402NextState(unittest.TestCase):
```
acolomb: This is an internal method, not an API.  While it's nice to see some consistency checks on it, the real test subject should be the `.state` property.  I know that's a bit more involved, but injecting TPDO value changes should provide a nice way to simulate a real CAN node answering.

test/test_p402.py
```diff
@@ -0,0 +1,281 @@
...
+            with self.subTest(statusword=hex(sw)):
+                self.node.tpdo_values[0x6041] = sw
+                # Test the bitmask logic directly without calling _homing_status,
+                # which would require a fully configured TPDO setup
+                status = None
+                for key, value in Homing.STATES.items():
+                    bitmask, bits = value
+                    if sw & bitmask == bits:
+                        status = key
+                self.assertEqual(status, expected)
```
acolomb: This is pretty useless, as the node itself is not even involved.  Again, trying to simulate TPDO reception would be a more valuable approach.  As is, this test class could be nuked.

test/test_p402.py
```diff
@@ -0,0 +1,281 @@
...
+    """Test OperationMode lookup tables."""
+
+    def test_code2name_name2code_round_trip(self):
+        for code, name in OperationMode.CODE2NAME.items():
+            self.assertEqual(OperationMode.NAME2CODE[name], code)
+
+    def test_supported_bitmask_unique(self):
+        values = list(OperationMode.SUPPORTED.values())
+        # All bitmasks should be unique (each is a single bit)
+        self.assertEqual(len(values), len(set(values)))
```
acolomb: This is a funky AI idea.  The standard defines the bit values, I see no need for testing them against some artificial logic condition.

acolomb (changes requested):
Overall this looks nice, but much more valuable would be a simulation of actual TPDO reception.

test/test_p402.py
```diff
@@ -0,0 +1,407 @@
...
+        self._inject_statusword(0x0008)  # FAULT
+        self.assertTrue(self.node.is_faulted())
+
+    def test_is_faulted_false(self):
+        self._inject_statusword(0x0040)  # SWITCH ON DISABLED
+        self.assertFalse(self.node.is_faulted())
+
+    def test_controlword_read_raises(self):
+        with self.assertRaises(RuntimeError):
+            _ = self.node.controlword
```
acolomb, line 170: Not sure we want to guarantee this API property moving forward, but it does increase coverage :-)

test/test_p402.py
```diff
@@ -0,0 +1,407 @@
...
+        node = BaseNode402(1, _make_od())
+        node.sdo = MagicMock()
+        node.sdo.__getitem__ = MagicMock(return_value=MagicMock(raw=0x0040))
+        self.assertEqual(node.statusword, 0x0040)
+
+    def test_check_statusword_no_tpdo_pointers(self):
+        """check_statusword returns cached statusword when no TPDO pointers."""
+        self._inject_statusword(0x0027)
+        result = self.node.check_statusword()
+        self.assertEqual(result, 0x0027)
```
acolomb: This is a bit weird, as there should be no TPDO value when it is not mapped.  The implementation does behave this way, but it is not an intended access pattern.  Thus I'm fine with removing this test at the cost of slightly lower coverage.

test/test_p402.py
```diff
@@ -0,0 +1,407 @@
...
+            (0x1000, "ATTAINED"),
+            (0x1400, "TARGET REACHED"),
+            (0x2000, "ERROR VELOCITY IS NOT ZERO"),
+            (0x2400, "ERROR VELOCITY IS ZERO"),
+        ]
+        for sw, expected in test_cases:
+            with self.subTest(statusword=hex(sw)):
+                _inject_tpdo(self.node, 0x6041, sw)
+                result = self.node._homing_status()
+                self.assertEqual(result, expected)
```
acolomb: While this is a nice overall exercise of the statusword mapping, it is an internal method and thus not a good candidate for future-proof testing.  I'll let you decide whether you want to keep it.

acolomb (commented):
I've read through the code, but didn't verify each test case against the spec.  While it does increase coverage significantly, some test cases are of questionable value - typical for AI fantasies I guess.

But still better to have a basis to improve on later, we can always go back and remove / improve tests when the library code evolves.

Please have a look at the few outstanding comments.

test/test_p402.py
bizfsc: True

acolomb:
One thing about the import ordering: I added a bit of ruff configuration in pyproject.toml to make it behave in line with isort.  The latter is what we previously applied for sorting, but the ruff default settings disagree because of type-grouping in a case-insensitive manner.  Let's quiet down that noise so you can keep using ruff.  I recently started using it in parallel to flake8, but unfortunately they sometimes clash, while flake8 does give way more specific hints.  So the current approach is to tame ruff to accept the established style.

bizfsc:
Addressed the outstanding review comments from @acolomb:

- Removed `test_check_statusword_no_tpdo_pointers`: tested an unintended access pattern (cached TPDO value without a mapped TPDO pointer)
- Removed `TestBaseNode402Homing` (incl. `test_homing_status`): tested the internal method `_homing_status()`, which is not a stable public API

Also merged the latest `master`.

acolomb (approved):

PR #645, opened 2026-04-29
Add venv/ and .venv/ to .gitignore

bizfsc:
Add `venv/` and `.venv/` directories to `.gitignore` so that local virtual environments are not accidentally tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
+venv/
+.venv/
 build/
 develop-eggs/
 dist/


acolomb:
Please use local ignores for such things.  There is no code in the repo to create these artifacts, thus I do not want to litter the curated ignore list with random stuff used by some developers.  Let's keep it for generated stuff that is expected because of what the committed code does.

See here:
https://docs.github.com/en/get-started/git-basics/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

Maybe even:
https://docs.github.com/en/get-started/git-basics/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

bizfsc:
I am fine with this. I just though I add this because there were already some added lines.

PR #647, opened 2026-04-29
Fix trivial mypy errors

bizfsc:
Fix trivial mypy errors to prepare for type checking adoption.

### `pyproject.toml`
- Add `[[tool.mypy.overrides]]` to ignore missing imports for `can` and `canmatrix` (third-party libraries without type stubs).

### `canopen/objectdictionary/__init__.py`
- Replace `List[int]` with `list[int]` (PEP 585, supported since Python 3.9).
- Add type annotations for `subindices` and `names` dicts in `ODRecord` and `ODArray`.

### `canopen/lss.py`
- Annotate `self.responses` as `queue.Queue[bytes]`.

### `canopen/node/local.py`
- Annotate `_read_callbacks` and `_write_callbacks` as `list[Callable]`.

### `canopen/node/remote.py`
- Annotate `sdo_channels` as `list[SdoClient]`.

## Impact
Reduces mypy errors from 95 to 87 (eliminates all `import-not-found`, `name-defined`, and `var-annotated` errors).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,3 @@ exclude = [
     "^test*",
     "^setup.py*",
 ]
-
-[[tool.mypy.overrides]]
-module = ["canmatrix.*"]
-ignore_missing_imports = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,3 +56,7 @@ exclude = [
     "^test*",
     "^setup.py*",
 ]
+
+[[tool.mypy.overrides]]
+module = ["canmatrix.*"]
+ignore_missing_imports = true
--- a/canopen/lss.py
+++ b/canopen/lss.py
@@ -85,7 +85,7 @@ class LssMaster:
         self.network: canopen.network.Network = canopen.network._UNINITIALIZED_NETWORK
         self._node_id = 0
         self._data = None
-        self.responses = queue.Queue()
+        self.responses: queue.Queue[bytes] = queue.Queue()
 
     def send_switch_state_global(self, mode):
         """switch mode to CONFIGURATION_STATE or WAITING_STATE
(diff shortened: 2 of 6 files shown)


acolomb:
Do the two libraries really not ship any typing info?  At least for python-can, I'm certain they do have them, but not as stubs.

Oh, and please, could we try to separate a bit more, touching only related modules at a time?

bizfsc:
> Oh, and please, could we try to separate a bit more, touching only related modules at a time?

I though about this, but I decided against it, because I really tried only to fix the most trivial cases in this PR. There will be future PRs (currently working on it) which will be separated by mypy topic.
Per module PRs are not applicable here, then we would have Count of PR = topic * module. Putting together different topics but for one module in a single PR also seems bad.
The structure of the changes in this PR are coherent and straight forward to review, don't you think?

acolomb:
Alright, it is in fact fairly straightforward.  One more thing I'd stuff in there though is annotating the argument of `add_read_callback()` and `add_write_callback()`, since their only purpose is writing to the newly types lists.

What about the dependency typing info?  canmatrix has it in the development branch, but not yet released.  can is properly typed and marked as such.

pyproject.toml
```diff
@@ -56,3 +56,10 @@ exclude = [
     "^test*",
     "^setup.py*",
 ]
+
+[[tool.mypy.overrides]]
+module = [
+    "can.*",
+    "canmatrix.*",
+]
+ignore_missing_imports = true
```
acolomb:
```suggestion
[[tool.mypy.overrides]]
module = ["canmatrix"]
follow_untyped_imports = true
```

Less harmful when type hints make it upstream?  Not needed for can, AFAICT.

bizfsc: You are right, somehow my python-can install was broken.

acolomb (approved):

PR #648, opened 2026-04-29
Fix encode_phys precision loss for int64 with factor=1

bizfsc:
Skip the float division in encode_phys when factor is 1 (the default). Previously, `value /= self.factor` always performed float division, causing up to 10 bits of precision loss for large int64 values (e.g. 0x55554444AAAABBBB became 0x55554444AAAABC00).

Use `!= 1` for the comparison since we are comparing values, not identity. Also clean up the redundant int(round(...)) to just round(), since round() already returns int when ndigits is omitted.

Add unit tests covering:
- int64 roundtrip with factor=1 (the original bug)
- int type preservation with factor=1
- rounding with integer factor != 1 (existing behavior preserved)
- float factor (existing behavior preserved)

Based on the discussion in #611.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -484,8 +484,8 @@ class ODVariable:
 
     def encode_phys(self, value: Union[int, bool, float, str, bytes]) -> int:
         if self.data_type in INTEGER_TYPES:
-            value /= self.factor
-            value = int(round(value))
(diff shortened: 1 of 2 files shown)


acolomb (approved):
Slight changes to the tests applied.  Including one additional test to ensure a float factor always promotes the result to float (when decoding).

friederschueler (approved):

PR #650, opened 2026-04-29
Fix Optional handling for node IDs and PDO variables

bizfsc:
Fix `Optional` handling to prevent operations on potentially `None` values.

## Problem

mypy reports `[operator]` and `[union-attr]` errors where `Optional` values are used without None-checks — e.g. `0x600 + self.id` when `self.id` could be `None`, or `divmod(self.offset, 8)` when `self.offset` is `None`.

### `canopen/node/base.py`
- Annotate `self.id` as `Optional[int]` to match its runtime value (`node_id or self.object_dictionary.node_id`).

### `canopen/node/remote.py` / `canopen/node/local.py`
- Add `assert self.id is not None` in `__init__` and re-annotate `self.id: int` to narrow the type for all methods.

### `canopen/pdo/base.py`
- Add `assert` guards for `self.offset` and `self.pdo_parent` at the start of `get_data()` and `set_data()`.
- Fix `wait_for_reception()` return type: `Optional[float]` (was `float`, but returns `None` on timeout).

## Impact
Eliminates 21 `[operator]` and `[union-attr]` errors in node and PDO modules.", "draft">false

### DIFF
--- a/canopen/node/base.py
+++ b/canopen/node/base.py
@@ -1,4 +1,4 @@
-from typing import TextIO, Union
+from typing import Optional, TextIO, Union
 
 import canopen.network
 from canopen.objectdictionary import ObjectDictionary, import_od
@@ -25,7 +25,7 @@ class BaseNode:
             object_dictionary = import_od(object_dictionary, node_id)
         self.object_dictionary = object_dictionary
 
-        self.id = node_id or self.object_dictionary.node_id
+        self.id: Optional[int] = node_id or self.object_dictionary.node_id
 
     def has_network(self) -> bool:
         """Check whether the node has been associated to a network."""
--- a/canopen/node/local.py
+++ b/canopen/node/local.py
@@ -24,6 +24,8 @@ class LocalNode(BaseNode):
         object_dictionary: Union[ObjectDictionary, str],
     ):
         super(LocalNode, self).__init__(node_id, object_dictionary)
+        assert self.id is not None, "node_id must be specified"
+        self.id: int
 
         self.data_store: dict[int, dict[int, bytes]] = {}
         self._read_callbacks = []
--- a/canopen/node/remote.py
+++ b/canopen/node/remote.py
@@ -35,6 +35,8 @@ class RemoteNode(BaseNode):
         load_od: bool = False,
     ):
         super(RemoteNode, self).__init__(node_id, object_dictionary)
+        assert self.id is not None, "node_id must be specified"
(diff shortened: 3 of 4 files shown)


acolomb:
I think this is going in the wrong direction.  The possibility of `None` for node IDs causes many problems later on, so instead we should make sure that the ID is definitely set and maybe even in the valid range.

The intention here (according to the docstring) is that a dummy value can be passed for `node_id` in order to respect the one from the OD.  Note that the mentioned passing of `None` even contradicts the type hint and is bad advice.  Passing zero is just as good for an "unspecified" sentinel value.  If neither source leads to a usable node ID, I don't think we should even finish constructing the node object.

What do you think?  Should we try the opposite way and restrict `self.id` to purely (valid) integers in general?

bizfsc:
Agreed — this is the right direction. Making [self.id](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) [Optional](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) just pushes the problem downstream to every caller. It's better to enforce a valid ID at construction time.

The current pattern [node_id or self.object_dictionary.node_id](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) already treats 0 and None identically as "unspecified", which aligns with your suggestion. The fix would be: if the result is still None (or 0), raise a [ValueError](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [BaseNode.__init__](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instead of silently leaving the object in a broken state.

Happy to rework the PR in that direction.

acolomb:
I'm working on a counter suggestion PR right now.  See #657.

friederschueler:
Closed in favor of #657

acolomb:
I think there are some other improvements in here that #657 doesn't cover.  Feel free to reopen with reduced scope, or make a new PR.

PR #651, opened 2026-04-29
Fix arg-type, return-value, and assignment mypy errors

bizfsc:
Fix `arg-type`, `return-value`, `assignment`, `return`, `index`, and `operator` mypy errors across the codebase.

### `canopen/nmt.py`
- Use `bytes()` instead of `list[int]` for `send_message()` and `send_periodic()` calls.
- Use `b''` instead of `None` for node guarding remote frame data.

### `canopen/emcy.py`
- Fix `wait()` return type: `Optional[EmcyError]` (was `EmcyError`, but returns `None` on timeout).

### `canopen/network.py`
- Fix `add_node()` return type to `Union[RemoteNode, LocalNode]` (accepts both).
- Fix `create_node()` parameter type to `Union[int, LocalNode]`.
- Add `assert node.id is not None` for index safety.
- Add `type: ignore[arg-type]` for node construction where OD may be None.

### `canopen/variable.py`
- Add `raise ValueError` at end of `read()` to satisfy missing return.
- Annotate `Bits.raw` with proper type.
- Add `isinstance` check for `value_descriptions` dict access.
- Add `type: ignore` for dynamically typed operations (`decode_phys`, `decode_desc`, `encode_bits`).

### `canopen/sdo/base.py`
- Add explicit `return None` in `get_variable()`.
- Add `type: ignore[return-value]` for `SdoArray.__len__()`.

### `canopen/objectdictionary/__init__.py`
- Fix `__getitem__` dot-notation access with `isinstance` check for indexability.
- Add explicit `return None` in `get_variable()`.
- Annotate `ODVariable.parent` and `ODArray.parent` with proper Union types.
- Fix `decode_phys()` return type to `Union[int, float]`.
- Rewrite `encode_phys()` without intermediate division.
- Add `type: ignore` for `encode_raw` string operations and min/max comparisons.
- Add `type: ignore` for `dest.close()` and `source.name` in file handling.

## Impact
Reduces mypy error count from 95 to 51 (standalone), and combined with PRs #647, #649, #650 will bring the count down significantly further.",
<parameter name="draft">false

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -207,7 +207,7 @@ class ODRecord(MutableMapping):
         #: Name of record
         self.name = name
         #: Storage location of index
-        self.storage_location = None
+        self.storage_location: Optional[str] = None
         self.subindices: dict[int, ODVariable] = {}
         self.names: dict[str, ODVariable] = {}
 
@@ -268,7 +268,7 @@ class ODArray(Mapping):
         #: Name of array
         self.name = name
         #: Storage location of index
-        self.storage_location = None
+        self.storage_location: Optional[str] = None
         self.subindices: dict[int, ODVariable] = {}
         self.names: dict[str, ODVariable] = {}
 
@@ -369,6 +369,8 @@ class ODVariable:
         self.data_type: Optional[int] = None
         #: Access type, should be "rw", "ro", "wo", or "const"
         self.access_type: str = "rw"
+        #: The variable represents a DOMAIN ObjectType
+        self.is_domain: bool = False
         #: Description of variable
         self.description: str = ""
         #: Dictionary of value descriptions
@@ -376,7 +378,7 @@ class ODVariable:
         #: Dictionary of bitfield definitions
         self.bit_definitions: dict[str, list[int]] = {}
         #: Storage location of index
-        self.storage_location = None
+        self.storage_location: Optional[str] = None
         #: Can this variable be mapped to a PDO
         self.pdo_mappable = False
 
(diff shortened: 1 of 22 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -139,7 +139,10 @@ def __getitem__(
         if item is None:
             if isinstance(index, str) and '.' in index:
                 idx, sub = index.split('.', maxsplit=1)
-                return self[idx][sub]
+                parent = self[idx]
+                if not isinstance(parent, (ODRecord, ODArray)):
+                    raise KeyError(f"{pretty_index(index)} was not found in Object Dictionary")
```
acolomb: Is this really better than a TypeError?  I think it's pretty clear to see "not indexable" in a message where a sub-element of some object was requested via dot notation.

acolomb (commented):
Some good stuff in here, but also questionable decisions.  Especially the `# type: ignore` stuff makes me want to fix the underlying problem instead of hiding the error.  Some of the assertions may already be remedied by #657, and we need to find an overall strategy for handling the general type flexibility inherent in the API (where #652 does give a good overview).

I think we'll pick some of the uncontroversial changes from this, but not all changes here align with my vision of a "good solution".  Need a bit more time to pick them apart, though.

acolomb:
> * Fix `decode_phys()` return type to `Union[int, float]`.

This method must be questioned at a deeper level.  The `.phys` property docstring mentions "Non integers will be passed as is."

So it is in fact desirable to allow passing back unmodified non-numeric values.  On the other hand, why should the factor only be applied to integer types, but not if the underlying variable is already a float?

As for typing, would generics help us here?  If an integer or float is passed, the result is definitely numeric (depending on the factor), otherwise it's the same as the argument type.  This could probably be expressed via `@overload` annotations, which I just learned more about.

> * Rewrite `encode_phys()` without intermediate division.

This overlaps with #648.  Regarding type checks, an assertion after the `.data_type` check would be a good compromise.

bizfsc:
Updated this PR

PR #653, opened 2026-05-04
eds: Support custom_options for unknown EDS fields

bizfsc:
Supersedes #615.

When importing an EDS/DCF file, any key that is not part of the CiA 306 standard field set is now collected into a `custom_options` dict on the object. This allows applications to round-trip vendor-specific metadata without losing it.

## Changes

- `ODVariable`, `ODRecord` and `ODArray` gain a `custom_options: dict[str, str]` attribute (default `{}`).
- New `_STANDARD_OPTIONS` set in `eds.py` lists all keys that are parsed explicitly. `ObjFlags` and `Denotation` are intentionally **not** in this set — they flow through `custom_options` and survive round-trips until first-class support is added in #654.
- New `_get_custom_options(eds, section)` helper collects every key not in `_STANDARD_OPTIONS`.
- `custom_options` is populated for top-level VAR/DOMAIN, ARRAY and RECORD objects as well as for sub-variables via `build_variable()`.
- `copy_variable()` now gives each copied variable its own `custom_options` dict to avoid shared-state mutations between shallow-copy clones.
- `export_variable()` and `export_record()` write `custom_options` back to the EDS file, guarded by a `_STANDARD_OPTIONS` check to prevent user-supplied keys from clobbering standard fields.

## Tests

- `test_reading_custom_options` — VAR with `Category=Motor`, `Offset=100`
- `test_custom_options_standard_keys_excluded` — standard keys absent from `custom_options`
- `test_custom_options_empty_for_standard_object` — heartbeat time has `{}`
- `test_custom_options_record` — RECORD-level `custom_options`
- `test_roundtrip_custom_options` — full EDS export/import cycle preserves values
- `test_roundtrip_custom_options_not_duplicated_as_standard` — standard keys stay out after round-trip

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -298,7 +298,7 @@ class ODArray(Mapping):
                          "bit_definitions", "storage_location"):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
-            var.custom_options = getattr(template, "custom_options", {})
+            var.custom_options = template.custom_options
         else:
             raise KeyError(f"Could not find subindex {pretty_index(None, subindex)}")
         return var
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -295,8 +295,7 @@ class ODArray(Mapping):
                          "bit_definitions", "storage_location"):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
-            if "custom_options" in template.__dict__:
-                var.__dict__["custom_options"] = template.__dict__["custom_options"].copy()
+            var.custom_options = getattr(template, "custom_options", {})
         else:
             raise KeyError(f"Could not find subindex {pretty_index(None, subindex)}")
         return var
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -295,10 +295,9 @@ class ODArray(Mapping):
             var.parent = self
             for attr in ("data_type", "unit", "factor", "min", "max", "default",
                          "access_type", "description", "value_descriptions",
-                         "bit_definitions", "storage_location"):
-                if attr in template.__dict__:
-                    var.__dict__[attr] = template.__dict__[attr]
-            var.custom_options = template.custom_options
+                         "bit_definitions", "storage_location", "custom_options"):
+                if (template_value := getattr(template, attr)) is not None:
+                    setattr(var, attr, template_value)
         else:
             raise KeyError(f"Could not find subindex {pretty_index(None, subindex)}")
         return var
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -280,7 +280,10 @@ class ODArray(Mapping):
         return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
 
     def __getitem__(self, subindex: Union[int, str]) -> ODVariable:
-        var = self.names.get(subindex) or self.subindices.get(subindex)
+        var = (
+            self.names.get(subindex)  # type: ignore[arg-type]
+            or self.subindices.get(subindex)  # type: ignore[arg-type]
+        )
         if var is not None:
             # This subindex is defined
             pass
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -210,6 +210,8 @@ class ODRecord(MutableMapping):
         self.storage_location: Optional[str] = None
         self.subindices: dict[int, ODVariable] = {}
         self.names: dict[str, ODVariable] = {}
+        #: Key-Value pairs not defined by the standard
+        self.custom_options: dict[str, str] = {}
 
     def __repr__(self) -> str:
         return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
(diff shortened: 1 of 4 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -287,7 +291,7 @@ def __getitem__(self, subindex: Union[int, str]) -> ODVariable:
             var.parent = self
             for attr in ("data_type", "unit", "factor", "min", "max", "default",
                          "access_type", "description", "value_descriptions",
-                         "bit_definitions", "storage_location"):
+                         "bit_definitions", "storage_location", "custom_options"):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
```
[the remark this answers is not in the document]
bizfsc: Shared custom options for all ODArray items is correct, otherwise use ODRecord

acolomb:
With all the AI output in between, it's hard to keep track of what changed.  I'd really prefer not force-pushing additional fixes (or even merging the master branch).  Much easier to see what's going on with simple commits on top, ideally with a one-line mention what was touched.  Will be squashed away when merging.  Thanks!

canopen/objectdictionary/__init__.py
```diff
@@ -291,6 +295,8 @@ def __getitem__(self, subindex: Union[int, str]) -> ODVariable:
                          "bit_definitions", "storage_location"):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
+            if "custom_options" in template.__dict__:
```
acolomb: This dunder access looks hacky.  Better to use `getattr()` in new code, which also collapses the existence check and value retrieval.

canopen/objectdictionary/eds.py
```diff
@@ -257,6 +260,25 @@ def _revert_variable(var_type, value):
     else:
         return f"0x{value:02X}"
 
+_STANDARD_OPTIONS = {
```
acolomb, line 275:
```suggestion

_STANDARD_OPTIONS = {
```

canopen/objectdictionary/eds.py
```diff
@@ -257,6 +260,25 @@ def _revert_variable(var_type, value):
...
+    "ParameterValue", "Factor", "Description", "Unit",
+    "StorageLocation", "CompactSubObj",
+    # CiA 306 fields parsed explicitly:
+    "SubNumber",
+    # ObjFlags and Denotation are intentionally absent: they are not yet
+    # parsed by this codebase, so they flow through custom_options and
+    # survive round-trips. Proper first-class support is tracked in #654.
+}
+
+def _get_custom_options(eds, section):
```
acolomb:
```suggestion

def _get_custom_options(eds, section):
```

canopen/objectdictionary/eds.py
```diff
@@ -359,6 +383,8 @@ def copy_variable(eds, section, subindex, src_var):
     # It is only the name and subindex that varies
     var.name = name
     var.subindex = subindex
+    # Give the copy its own custom_options dict to avoid shared-state mutations
+    var.custom_options = src_var.custom_options.copy()
```
acolomb: This is only used once, and IIRC related to arrays.  Doesn't the "shared custom_options for array members" approach apply here as well?

test/test_eds.py
```diff
@@ -251,6 +251,54 @@ def test_roundtrip_domain_objects(self):
         self.assertTrue(od2[0x3063].is_domain)
         self.assertTrue(od2[0x3064][1].is_domain)
 
+    def test_reading_custom_options(self):
+        # custom options (unknown EDS keys) are collected in custom_options dict
```
acolomb: Please change the function comments into docstrings where appropriate.

acolomb (changes requested):
Looks great overall, but one question about the shared mutable dicts, see below.

canopen/objectdictionary/__init__.py
```diff
@@ -291,6 +295,7 @@ def __getitem__(self, subindex: Union[int, str]) -> ODVariable:
                          "bit_definitions", "storage_location"):
                 if attr in template.__dict__:
                     var.__dict__[attr] = template.__dict__[attr]
+            var.custom_options = getattr(template, "custom_options", {})
```
acolomb: On second thought, this attribute always exists.  Let's avoid the potential creation of non-shared dicts here as well.

```suggestion
            var.custom_options = template.custom_options
```

acolomb (changes requested):

canopen/objectdictionary/__init__.py
acolomb: Or handle it as part of the enumerated attributes?  The existence check is not really needed in this specific case, but still simpler to use a common code path.  If you decide to do that, please switch to getattr / setattr in the process.

acolomb:
The two remaining "partial" lines from the Codecov report can be ignored IMHO, that's only the safety check to not rewrite standard options.

canopen/objectdictionary/eds.py
bizfsc: yes

canopen/objectdictionary/__init__.py
bizfsc: yes

canopen/objectdictionary/__init__.py
acolomb: This is what I had in mind: (GutHub won't render it correctly as a suggestion)

```python3
            for attr in ("data_type", "unit", "factor", "min", "max", "default",
                         "access_type", "description", "value_descriptions",
                         "bit_definitions", "storage_location", "custom_options"):
                if (template_value := getattr(template, attr)) is not None:
                    setattr(var, attr, template_value)
```

bizfsc:
We use getattr/setattr instead of __dict__ access.
Merge ready?

acolomb (approved):

PR #654, opened 2026-05-04
eds: Add first-class support for CiA 306 ObjFlags and Denotation

bizfsc:
`ObjFlags` (UNSIGNED32 bitfield) and `Denotation` (DCF-only string) are standard CiA 306 fields that were previously silently dropped. This PR adds first-class support for both.

## Changes

- `ODVariable`, `ODRecord` and `ODArray` gain an `obj_flags: int = 0` attribute.
- `ODVariable` gains a `denotation: str = ""` attribute.
- New `_get_obj_flags(eds, section)` helper reads and validates the `ObjFlags` integer value.
- `ObjFlags` is parsed for VAR/DOMAIN (via `build_variable()`), ARRAY and RECORD top-level objects.
- `ObjFlags` is exported whenever non-zero.
- `Denotation` is read in `build_variable()` and written only in DCF mode (`device_commisioning=True`), matching the CiA 306 specification.

## Tests

- `test_reading_obj_flags` — VAR with `ObjFlags=0x1` reads back correctly
- `test_reading_obj_flags_default` — standard objects default to `0`
- `test_roundtrip_obj_flags` — full EDS export/import cycle preserves `obj_flags`
- `test_denotation_roundtrip_dcf` — `denotation` survives a DCF round-trip
- `test_denotation_not_exported_in_eds_mode` — `denotation` is absent in plain EDS export

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -208,11 +208,16 @@ class ODRecord(MutableMapping):
         self.name = name
         #: Storage location of index
         self.storage_location: Optional[str] = None
+        #: CiA 306 ObjFlags bitfield
+        self.obj_flags: int = 0
+        #: CiA 306 Denotation string (DCF only)
+        self.denotation: str = ""
         self.subindices: dict[int, ODVariable] = {}
         self.names: dict[str, ODVariable] = {}
 
     def __repr__(self) -> str:
-        return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}>"
+        flags = f" flags=0x{self.obj_flags:X}" if self.obj_flags else ""
+        return f"<{type(self).__qualname__} {self.name!r} at {pretty_index(self.index)}{flags}>"
 
     def __getitem__(self, subindex: Union[int, str]) -> ODVariable:
         item = self.names.get(subindex) or self.subindices.get(subindex)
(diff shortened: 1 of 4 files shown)


canopen/objectdictionary/__init__.py
```diff
@@ -372,6 +376,10 @@ def __init__(self, name: str, index: int, subindex: int = 0):
         self.bit_definitions: dict[str, list[int]] = {}
         #: Storage location of index
         self.storage_location = None
+        #: CiA 306 ObjFlags bitfield
+        self.obj_flags: int = 0
+        #: CiA 306 Denotation string (DCF only)
+        self.denotation: str = ""
```
acolomb, line 395: Why is this only applied to ODVariable, not arrays and records?

Do you think this makes sense as part of the `__repr__` or at least string representation, if set?

bizfsc:
Is this YAGNI?

acolomb:
Frankly, I never heard of it. But now that I have, I had an immediate idea where it could be used in my code, replacing a previous workaround.

Some controllers have extra objects for "user variables" which can be utilized in a user-supplied small PLC program on the device. The function of this custom program defines what the variable is really used for (denotation), yet the device manufacturer can only assign a generic name of "user variable 1" in the generic device EDS. That sounds like a perfect application of this less widespread option. 

The ObjFlags option also seems to solve a problem I've previously worked around by hard-coding objects to skip.

I'll have to take a deeper look to really judge if it works like I imagine. And get better acquainted with their definition in CiA 306.

acolomb (commented):
Again just a partial review, forgot to send this a few days ago...

acolomb:
@bizfsc It would help review a bit if you could push additional fix commits instead of force-pushing.  Easier to see the difference after some time has passed and the PR got swapped out of brain.  Thanks!

canopen/objectdictionary/__init__.py
bizfsc: Yes, if set (!=0) we include this now.

bizfsc:
Well with the rebase from master and the other PRs already merged, this does not work :(  I could use merge instead of rebase, but then there will merge changes in the PR commit history which is often not wanted.

acolomb:
The PR commit history is not important after (squash-)merging.  But during review, it helps to see what changed, even if that is just "merged from master branch".  If there are no conflicts, the rebase / merge isn't strictly needed either.  I have deliberately disables the requirement to have an up-to-date upstream in the PR for merging.

bizfsc:
Ok, I will no longer rebase and force-push in my PRs from now on. Sorry!

acolomb:
Sorry to keep nagging about your PR structure... But this last commit is really mixing in a load of unrelated changes.  There may be good reasons for these changes, but we need to keep PRs focused on the actually intended change.  Any roadside cleanup touching other code belongs in its own PR.

Actually I have quite a bit of cleanup and simplifications on the table for this module already.  Just wanted to wait for #653 and #658 to land before touching that file again and ending up with lots of conflicts.  I will now concentrate on that effort first.

Some of the added "fixes" here I am actually not happy with, because they are hiding type checker errors instead of fixing them.  Others have already been fixed in other PRs, so a merge from master would be the better approach.

PR #655, opened 2026-05-04
Memorized DOMAIN ObjectType for a proper export

FedericoSpada:
While working with an EDS that had DOMAIN objects, I've noted that when imported and exported using this library, all DOMAIN Objects were converted to normal Variables.

While importing an EDS, I've changed to save whether a Variable was created from an object with `ObjectType==DOMAIN`. In this way, it is possible to export it unchanged with the proper ObjectType value.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -364,6 +364,8 @@ class ODVariable:
         self.data_type: Optional[int] = None
         #: Access type, should be "rw", "ro", "wo", or "const"
         self.access_type: str = "rw"
+        #: The variable represents a DOMAIN ObjectType
+        self.is_domain: bool = False
         #: Description of variable
         self.description: str = ""
         #: Dictionary of value descriptions
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -129,7 +129,7 @@ def import_eds(source, node_id):
                 storage_location = None
 
             if object_type in (objectcodes.VAR, objectcodes.DOMAIN):
-                var = build_variable(eds, section, node_id, index)
+                var = build_variable(eds, section, node_id, index, is_domain=object_type==objectcodes.DOMAIN)
                 od.add_object(var)
             elif object_type == objectcodes.ARRAY and eds.has_option(section, "CompactSubObj"):
                 arr = ODArray(name, index)
@@ -158,7 +158,11 @@ def import_eds(source, node_id):
             subindex = int(match.group(2), 16)
             entry = od[index]
             if isinstance(entry, (ODRecord, ODArray)):
-                var = build_variable(eds, section, node_id, index, subindex)
+                try:
+                    object_type = int(eds.get(section, "ObjectType"), 0)
+                except NoOptionError:
+                    object_type = objectcodes.VAR
+                var = build_variable(eds, section, node_id, index, subindex, is_domain=object_type==objectcodes.DOMAIN)
                 entry.add_member(var)
 
         # Match [index]Name
@@ -252,13 +256,14 @@ def _revert_variable(var_type, value):
         return f"0x{value:02X}"
 
 
-def build_variable(eds, section, node_id, index, subindex=0):
+def build_variable(eds, section, node_id, index, subindex=0, is_domain=False):
     """Creates a object dictionary entry.
     :param eds: String stream of the eds file
     :param section:
     :param node_id: Node ID
     :param index: Index of the CANOpen object
-    :param subindex: Subindex of the CANOpen object (if presente, else 0)
+    :param subindex: Subindex of the CANOpen object (if present, else 0)
+    :param is_domain: variable represents a DOMAIN ObjectType (if present, else False)
     """
     name = eds.get(section, "ParameterName")
     var = ODVariable(name, index, subindex)
@@ -268,6 +273,7 @@ def build_variable(eds, section, node_id, index, subindex=0):
         var.storage_location = None
     var.data_type = int(eds.get(section, "DataType"), 0)
     var.access_type = eds.get(section, "AccessType").lower()
+    var.is_domain = is_domain
     if var.data_type > 0x1B:
         # The object dictionary editor from CANFestival creates an optional object if min max values are used
         # This optional object is then placed in the eds under the section [A0] (start point, iterates for more)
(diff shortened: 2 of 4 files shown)


bizfsc (approved):
Looking good. As far as I can check it, no new problems introduced for mypy.

canopen/objectdictionary/eds.py
```diff
@@ -129,7 +129,7 @@ def import_eds(source, node_id):
                 storage_location = None
 
             if object_type in (objectcodes.VAR, objectcodes.DOMAIN):
-                var = build_variable(eds, section, node_id, index)
+                var = build_variable(eds, section, node_id, index, is_domain=object_type==objectcodes.DOMAIN)
```
friederschueler: PEP 8: missing spaces around the `==` comparison inside the keyword argument.

```suggestion
                var = build_variable(eds, section, node_id, index, is_domain=object_type == objectcodes.DOMAIN)
```

canopen/objectdictionary/eds.py
```diff
@@ -158,7 +158,11 @@ def import_eds(source, node_id):
             subindex = int(match.group(2), 16)
             entry = od[index]
             if isinstance(entry, (ODRecord, ODArray)):
-                var = build_variable(eds, section, node_id, index, subindex)
+                try:
+                    object_type = int(eds.get(section, "ObjectType"), 0)
+                except NoOptionError:
+                    object_type = objectcodes.VAR
+                var = build_variable(eds, section, node_id, index, subindex, is_domain=object_type==objectcodes.DOMAIN)
```
friederschueler: Same PEP 8 spacing issue as above.

```suggestion
                var = build_variable(eds, section, node_id, index, subindex, is_domain=object_type == objectcodes.DOMAIN)
```

test/test_eds.py
```diff
@@ -111,6 +111,7 @@ def test_variable(self):
         self.assertEqual(var.name, 'Producer heartbeat time')
         self.assertEqual(var.data_type, canopen.objectdictionary.UNSIGNED16)
         self.assertEqual(var.access_type, 'rw')
+        self.assertEqual(var.is_domain, False)
```
friederschueler: Nit: for boolean checks the `unittest` idiom is `assertFalse`/`assertTrue` rather than `assertEqual(..., False/True)`. Same pattern applies to the other four analogous assertions added in this PR.

```suggestion
        self.assertFalse(var.is_domain)
```

canopen/objectdictionary/__init__.py
```diff
@@ -364,6 +364,8 @@ def __init__(self, name: str, index: int, subindex: int = 0):
         self.data_type: Optional[int] = None
         #: Access type, should be "rw", "ro", "wo", or "const"
         self.access_type: str = "rw"
+        #: The variable represents a DOMAIN ObjectType
+        self.is_domain: bool = False
```
friederschueler, line 368: Design consideration: a boolean `is_domain` only captures one special ObjectType. Storing the full `object_type: int` attribute (defaulting to `objectcodes.VAR`) would be more flexible (e.g. `NULL`, `DEFTYPE`) and avoids losing that information for manually constructed `ODVariable` instances. Just something to think about — a boolean is perfectly fine for the stated use-case.

friederschueler (commented):
Good fix for a real round-trip data-loss bug. The overall approach — adding `is_domain` on import and using it on export — is clean and well-tested.

A few small points:
- Two places in `eds.py` have missing spaces around `==` inside a keyword argument (see inline comments).
- The new boolean assertions in the tests would be more idiomatic as `assertFalse`/`assertTrue`.
- Worth considering storing the full `object_type` integer instead of just a boolean, but that is a larger API change and not required for this fix.

canopen/objectdictionary/eds.py
FedericoSpada: I've changed it, but for me, it is incredibly misleading: you are giving "priority" to the assignment instead of the operation that is performed before.
A possible solution would be `is_domain=(object_type == objectcodes.DOMAIN)`, but Pylint complains about the unnecessary parentheses.
For me, the proper way to write it would be `is_domain = object_type==objectcodes.DOMAIN`.
I've never understood why PEP 8 requires no spaces for "=" in function invocations...

canopen/objectdictionary/__init__.py
FedericoSpada: At the moment, `build_variable()` is called only when `object_type` is `VAR` or `DOMAIN`, so a boolean is enough.
Probably `NULL` and `DEFTYPE` would require dedicated classes, so even in the future, the chosen solution will be enough.

friederschueler (approved):

acolomb:
@friederschueler please use the "Squash and merge" GitHub function for PRs, with a sensible, hand-checked commit message.

In this case, I hadn't even got around to looking at the PR contents, so I'd like a bit more time before you merge future contributions.  Thank you! :slightly_smiling_face:

canopen/objectdictionary/eds.py
```diff
@@ -158,7 +158,11 @@ def import_eds(source, node_id):
             subindex = int(match.group(2), 16)
             entry = od[index]
             if isinstance(entry, (ODRecord, ODArray)):
-                var = build_variable(eds, section, node_id, index, subindex)
+                try:
+                    object_type = int(eds.get(section, "ObjectType"), 0)
+                except NoOptionError:
+                    object_type = objectcodes.VAR
+                var = build_variable(eds, section, node_id, index, subindex, is_domain=object_type == objectcodes.DOMAIN)
```
acolomb, line 165: This is certainly beyond a sensible line length limit. ;-)

I've codified 96 chars as a guideline in #660, please try to stick to that in future contributions.

canopen/objectdictionary/eds.py
```diff
@@ -129,7 +129,7 @@ def import_eds(source, node_id):
                 storage_location = None
 
             if object_type in (objectcodes.VAR, objectcodes.DOMAIN):
-                var = build_variable(eds, section, node_id, index)
+                var = build_variable(eds, section, node_id, index, is_domain=object_type == objectcodes.DOMAIN)
```
acolomb, line 132: Line length, see below.

friederschueler:
@acolomb I am sorry, this was muscle memory from work. I forgot we use hand-crafted messages. I personally think that the automated PR merge messages  are just fine, because the PR should be named in a good way. Also isn't squash commit the default? I will also give you more time to review.

canopen/objectdictionary/eds.py
```diff
@@ -252,13 +256,14 @@ def _revert_variable(var_type, value):
         return f"0x{value:02X}"
 
 
-def build_variable(eds, section, node_id, index, subindex=0):
+def build_variable(eds, section, node_id, index, subindex=0, is_domain=False):
```
acolomb, line 259: Frankly I don't particularly like this function signature.  The `is_domain` attribute could be set directly on the resulting object instead of adding a new call parameter.

Or even better, pass the whole `object_code` value and have the logic for the boolean attribute here in this function.  That's why we have this function with centralized logic.

PR #657, opened 2026-05-06
Make sure any BaseNode always has a valid Node ID

acolomb:
The BaseNode.id attribute can be initialized either from the given node_id parameter, or from the Object Dictionary if zero was passed. The None value as mentioned in the docstring actually violates the type hint, so drop that suggestion.  Make sure the inferred type of the self.id attribute does not allow None, but is always an integer.

Furthermore, check the resulting Node ID against the allowed value range and raise a ValueError instead of constructing a node object with an invalid Node ID, which will cause problems down the road. This is technically an API change, but it just causes things to fail early on instead of raising exceptions later on.

Copy the docstring for LocalNode as well, to document the expected node_id "invalid" value.  And because there was none at all before, include a bit of functional description to set the expectations straight.

### DIFF
--- a/canopen/node/base.py
+++ b/canopen/node/base.py
@@ -8,7 +8,7 @@ class BaseNode:
     """A CANopen node.
 
     :param node_id:
-        Node ID (set to None or 0 if specified by object dictionary)
+        Node ID (set to 0 if specified by object dictionary)
     :param object_dictionary:
         Object dictionary as either a path to a file, an ``ObjectDictionary``
         or a file like object.
@@ -25,7 +25,9 @@ class BaseNode:
             object_dictionary = import_od(object_dictionary, node_id)
         self.object_dictionary = object_dictionary
 
-        self.id = node_id or self.object_dictionary.node_id
+        self.id = node_id or object_dictionary.node_id or 0
+        if not self.id or not 1 <= self.id <= 127:
+            raise ValueError("No valid Node ID provided, %r not in range 1..127")
 
     def has_network(self) -> bool:
         """Check whether the node has been associated to a network."""
--- a/canopen/node/base.py
+++ b/canopen/node/base.py
@@ -26,8 +26,8 @@ class BaseNode:
         self.object_dictionary = object_dictionary
 
         self.id = node_id or object_dictionary.node_id or 0
-        if not self.id or not 1 <= self.id <= 127:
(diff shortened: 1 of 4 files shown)


acolomb:
Replaces #650, as far as Node ID assumptions are concerned.

canopen/node/base.py
```diff
@@ -25,7 +25,9 @@ def __init__(
             object_dictionary = import_od(object_dictionary, node_id)
         self.object_dictionary = object_dictionary
 
-        self.id = node_id or self.object_dictionary.node_id
+        self.id = node_id or object_dictionary.node_id or 0
+        if not self.id or not 1 <= self.id <= 127:
+            raise ValueError("No valid Node ID provided, %r not in range 1..127")
```
friederschueler: The format specifier `%r` is used in the string but no value is passed, so `self.id` will never appear in the error message. This should be:

```python
raise ValueError("No valid Node ID provided, %r not in range 1..127" % self.id)
```

or using an f-string:

```python
raise ValueError(f"No valid Node ID provided, {self.id!r} not in range 1..127")
```

canopen/node/base.py
```diff
@@ -25,7 +25,9 @@ def __init__(
             object_dictionary = import_od(object_dictionary, node_id)
         self.object_dictionary = object_dictionary
 
-        self.id = node_id or self.object_dictionary.node_id
+        self.id = node_id or object_dictionary.node_id or 0
+        if not self.id or not 1 <= self.id <= 127:
```
friederschueler: The `not self.id` check (catching `0`) is already a subset of `not (1 <= self.id <= 127)`, since `0` is outside that range. The condition can be simplified to:

```python
if not (1 <= self.id <= 127):
```

friederschueler (changes requested):
Two issues found in `canopen/node/base.py`:

1. **Bug**: The `ValueError` message uses `%r` as a format specifier but never substitutes `self.id`, so the actual invalid value is lost in the error message.
2. **Minor**: The condition `not self.id or not 1 <= self.id <= 127` has redundancy — `not self.id` is already covered by the range check. Can be simplified to `not (1 <= self.id <= 127)`.

canopen/node/base.py
acolomb: Of course you're right, on both accounts.  That's what you get for putting together a "quick" fix...

friederschueler (approved):

PR #658, opened 2026-05-06
eds: Fix signed integer limit parsing for all bit widths, log errors

friederschueler:
Fixes incomplete and incorrect signed integer handling for `LowLimit`/`HighLimit` in EDS parsing and export, as noted in #352.

### `_calc_bit_length` — extended to all SIGNED_TYPES

The existing function only handled INTEGER8/16/32/64. It now covers all 8 `SIGNED_TYPES` (INTEGER8/16/24/32/40/48/56/64) by deriving the bit width dynamically from `ODVariable.STRUCT_TYPES` instead of a hardcoded if/elif chain.

### `_signed_int_from_hex` — decimal and hex support, range validation

The parser now accepts both formats for `LowLimit`/`HighLimit`:
- Unsigned two's-complement hex (e.g. `0xFFFF` for INTEGER16 → `-1`)
- Signed decimal (e.g. `-32768` for INTEGER16 → `-32768`)

An explicit `ValueError` is raised if the value does not fit in the type's bit width. Previously, a value like `0xFFFF` for an 8-bit field would silently produce a wrong result.

### `_int_to_hex` — new helper for EDS integer serialization

Formats any integer as a valid EDS hex literal. Negative signed values are written as two's-complement unsigned hex (e.g. INTEGER32 `-1` → `0xFFFFFFFF`), which is the canonical EDS format. Used by `_revert_variable` and the export of `LowLimit`/`HighLimit`.

### `_revert_variable` — fix for negative integers

Previously used `f"0x{value:02X}"` unconditionally, which raises `ValueError` for negative values and broke export of any negative `DefaultValue`. Now delegates to `_int_to_hex`.

### Export of `LowLimit`/`HighLimit`

Previously written as raw Python integers (e.g. `-2147483648`). Now uses `_int_to_hex` to produce canonical unsigned two's-complement hex.

### Silent parse errors → `logger.warning`

All `except ValueError: pass` blocks in `build_variable` now log the field name, value, object name and index:
- `LowLimit`, `HighLimit`, `DefaultValue`, `ParameterValue`, `Factor`

Dead `try/except ValueError` around `Description` and `Unit` removed — `ConfigParser.get()` never raises `ValueError`.

## Tests

- Add `sample.eds` entries covering all four combinations of signed/unsigned × hex/decimal limits
- Refactor `test_record_with_limits` as a table-driven test with `subTest`
- Add `test_signed_int_from_hex_accepts_decimal` and `test_signed_int_from_hex_rejects_out_of_range`

Closes part of #352.

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -259,7 +259,7 @@ def _decode_from_eds(node_id: int, var_type: int, value: Any) -> Any:
 def _encode_to_eds(var_type: int, value: Any) -> Any:
     if value is None:
         return None
-    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN) and isinstance(value, bytes):
+    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN):
         return bytes.hex(value)
     elif var_type in (datatypes.VISIBLE_STRING, datatypes.UNICODE_STRING):
         return value
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -222,22 +222,22 @@ def _signed_int_from_hex(hex_str, bit_length):
     max_signed = (1 << (bit_length - 1)) - 1
     max_unsigned = (1 << bit_length) - 1
 
+    if number < min_signed:
+        raise ValueError(
+            f"Value {hex_str!r} is out of range for a {bit_length}-bit signed integer"
+        )
     if number < 0:
-        # Negative decimal literal (e.g. LowLimit=-32768)
-        if number < min_signed:
-            raise ValueError(
-                f"Value {hex_str!r} is out of range for a {bit_length}-bit signed integer"
-            )
+        # Negative literal (e.g. LowLimit=-32768 or -0x8000)
         return number
-    else:
+
+    if number > max_unsigned:
+        raise ValueError(
+            f"Value {hex_str!r} is out of range for a {bit_length}-bit signed integer"
+        )
+    if number > max_signed:
         # Unsigned hex literal, two's-complement (e.g. LowLimit=0xFFFF → -1 for INTEGER16)
-        if number > max_unsigned:
-            raise ValueError(
-                f"Value {hex_str!r} is out of range for a {bit_length}-bit signed integer"
-            )
-        if number > max_signed:
-            return number - (1 << bit_length)
-        return number
+        return number - (1 << bit_length)
+    return number
 
 
 def _convert_variable(node_id, var_type, value):
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -204,26 +204,28 @@ def import_from_node(node_id: int, network: canopen.network.Network):
     return od
 
 
-_SIGNED_BIT_LENGTHS = {
-    datatypes.INTEGER8: 8,
-    datatypes.INTEGER16: 16,
-    datatypes.INTEGER24: 24,
-    datatypes.INTEGER32: 32,
-    datatypes.INTEGER40: 40,
-    datatypes.INTEGER48: 48,
-    datatypes.INTEGER56: 56,
-    datatypes.INTEGER64: 64,
-}
+def _calc_bit_length(data_type: int) -> int:
+    if data_type in datatypes.SIGNED_TYPES:
+        st = ODVariable.STRUCT_TYPES[data_type]
+        if isinstance(st, datatypes.IntegerN):
+            return st.width
+        return st.size * 8
+    else:
+        raise ValueError(
+            f"Invalid data_type '{data_type}', expecting a signed integer data_type."
+        )
 
 
 def _signed_int_from_hex(hex_str, bit_length):
     number = int(hex_str, 0)
+    if number < 0 or number >= (1 << bit_length):
+        raise ValueError(
+            f"Value {hex_str!r} is out of range for a {bit_length}-bit signed integer"
+        )
     max_value = (1 << (bit_length - 1)) - 1
-
     if number > max_value:
         return number - (1 << bit_length)
-    else:
-        return number
+    return number
 
 
 def _convert_variable(node_id, var_type, value):
@@ -242,6 +244,18 @@ def _convert_variable(node_id, var_type, value):
             return int(value, 0)
 
 
+def _int_to_hex(data_type: int, value: int) -> str:
+    """Format an integer as EDS hex string.
+
+    Signed types with a negative value are written as two's-complement hex
+    (e.g. INTEGER8 -1 → 0xFF) so the output is a valid EDS literal.
+    """
+    if data_type in datatypes.SIGNED_TYPES and value < 0:
+        bit_length = _calc_bit_length(data_type)
+        return f"0x{value + (1 << bit_length):0{bit_length // 4}X}"
+    return f"0x{value:02X}"
+
+
 def _revert_variable(var_type, value):
     if value is None:
         return None
@@ -252,7 +266,7 @@ def _revert_variable(var_type, value):
     elif var_type in datatypes.FLOAT_TYPES:
         return value
     else:
-        return f"0x{value:02X}"
+        return _int_to_hex(var_type, value)
 
 
 def build_variable(
@@ -302,20 +316,26 @@ def build_variable(
         try:
             min_string = eds.get(section, "LowLimit")
             if var.data_type in datatypes.SIGNED_TYPES:
-                var.min = _signed_int_from_hex(min_string, _SIGNED_BIT_LENGTHS[var.data_type])
+                var.min = _signed_int_from_hex(min_string, _calc_bit_length(var.data_type))
             else:
                 var.min = int(min_string, 0)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid LowLimit %r for %s (0x%X), ignoring",
+                eds.get(section, "LowLimit"), var.name, var.index,
+            )
     if eds.has_option(section, "HighLimit"):
         try:
             max_string = eds.get(section, "HighLimit")
             if var.data_type in datatypes.SIGNED_TYPES:
-                var.max = _signed_int_from_hex(max_string, _SIGNED_BIT_LENGTHS[var.data_type])
+                var.max = _signed_int_from_hex(max_string, _calc_bit_length(var.data_type))
             else:
                 var.max = int(max_string, 0)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid HighLimit %r for %s (0x%X), ignoring",
+                eds.get(section, "HighLimit"), var.name, var.index,
+            )
     if eds.has_option(section, "DefaultValue"):
         try:
             var.default_raw = eds.get(section, "DefaultValue")
@@ -323,30 +343,33 @@ def build_variable(
                 var.relative = True
             var.default = _convert_variable(node_id, var.data_type, var.default_raw)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid DefaultValue %r for %s (0x%X), ignoring",
+                var.default_raw, var.name, var.index,
+            )
     if eds.has_option(section, "ParameterValue"):
         try:
             var.value_raw = eds.get(section, "ParameterValue")
             var.value = _convert_variable(node_id, var.data_type, var.value_raw)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid ParameterValue %r for %s (0x%X), ignoring",
+                var.value_raw, var.name, var.index,
+            )
     # Factor, Description and Unit are not standard according to the CANopen specifications, but
     # they are implemented in the python canopen package, so we can at least try to use them
     if eds.has_option(section, "Factor"):
         try:
             var.factor = float(eds.get(section, "Factor"))
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid Factor %r for %s (0x%X), ignoring",
+                eds.get(section, "Factor"), var.name, var.index,
+            )
     if eds.has_option(section, "Description"):
-        try:
-            var.description = eds.get(section, "Description")
-        except ValueError:
-            pass
+        var.description = eds.get(section, "Description")
     if eds.has_option(section, "Unit"):
-        try:
-            var.unit = eds.get(section, "Unit")
-        except ValueError:
-            pass
+        var.unit = eds.get(section, "Unit")
     return var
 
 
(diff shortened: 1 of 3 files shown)


acolomb:
Please always separate style changes from functional changes.  Except of course fixing in the immediate vicinity of the functional changes.

canopen/objectdictionary/eds.py
```diff
@@ -204,17 +204,16 @@ def import_from_node(node_id: int, network: canopen.network.Network):
...
+_SIGNED_BIT_LENGTHS = {
+    datatypes.INTEGER8: 8,
+    datatypes.INTEGER16: 16,
+    datatypes.INTEGER24: 24,
+    datatypes.INTEGER32: 32,
+    datatypes.INTEGER40: 40,
+    datatypes.INTEGER48: 48,
+    datatypes.INTEGER56: 56,
+    datatypes.INTEGER64: 64,
+}
```
acolomb: We don't need to hard-code these at all:

```suggestion
def _calc_bit_length(data_type: int) -> int:
    if data_type in datatypes.SIGNED_TYPES:
        st = ODVariable.STRUCT_TYPES[data_type]
        if isinstance(st, datatypes.IntegerN):
            return st.width
        return st.size * 8
    else:
        raise ValueError(
            f"Invalid data_type '{data_type}', expecting a signed integer data_type."
        )
```

canopen/objectdictionary/eds.py
```diff
@@ -292,20 +291,20 @@ def build_variable(eds, section, node_id, index, subindex=0, is_domain=False):
         try:
             min_string = eds.get(section, "LowLimit")
             if var.data_type in datatypes.SIGNED_TYPES:
-                var.min = _signed_int_from_hex(min_string, _calc_bit_length(var.data_type))
+                var.min = _signed_int_from_hex(min_string, _SIGNED_BIT_LENGTHS[var.data_type])
             else:
                 var.min = int(min_string, 0)
         except ValueError:
-            pass
+            logger.warning("Failed to parse LowLimit for %s: %r", name, min_string)
```
acolomb: Logging is better than not logging, of course.  But this whole treatment looks suspicious:

1. How does the outside code even notice that the EDS partly failed to parse?  Informing the user for troubleshooting is not enough IMHO, but raising would be a massive API break, causing potential errors with previously "working" EDS files.
2. AFAICT the _signed_int_from_hex() helper does not even fully check whether the value is representable in the given bit width.  That's another class of EDS errors we could easily catch and report here, but don't.
3. Ignoring `ValueError` seems like a pattern *blindly* applied to every option.  Even for `Description` and `Unit` below, which are strings and there is not even a chance of a `ValueError` from `ConfigParser.get()`.

When we touch / fix this code, let's put some more thought into what really makes sense.  Especially point 1 is not solved by logging alone.  Being stricter in parsing EDS files might be beneficial.

test/test_eds.py
```diff
@@ -148,6 +148,28 @@ def test_record_with_limits(self):
...
         self.assertEqual(int64.max, +10)
+        # Verify all remaining SIGNED_TYPES are handled (INTEGER24/40/48/56)
+        for index in (0x3031, 0x3032, 0x3033, 0x3034):
+            var = self.od[index]
+            self.assertEqual(var.min, -1, f"min mismatch at 0x{index:04X}")
+            self.assertEqual(var.max, 0, f"max mismatch at 0x{index:04X}")
+
+    def test_invalid_limit_logs_warning(self):
+        import io
+        import logging
```
acolomb: This really points toward bad API design, needing to examine logs for functional verification.

acolomb (changes requested):

canopen/objectdictionary/eds.py
friederschueler: Not a big fan of the function, following the zen of python:

simple is better than complex
explicit is better than implicit

a lookup dict is the way to go. one could argue to  put it into datatypes.py

test/test_eds.py
friederschueler: Well this is just to fulfill the requirements to test new lines. If you remove the logging, we can kill this test.

canopen/objectdictionary/eds.py
friederschueler: Ok, I get the point and tend to agree to just don't touch it at this point.

canopen/objectdictionary/eds.py
acolomb: In this case I'm more convinced by the "Don't Repeat Yourself" principle.  Listing combinations of numbers and a symbol including the same number, sounds pretty useless.  As the data is already there, I prefer a helper function in the `object_dictionary` module, instead of duplicating it here.  A dict can be generated as well from the existing data, but it is really a waste of memory 99.9 % of the time.

What makes this function a bit complicated is the fact that the dict in class `ODVariable` uses two different approaches for different bit widths.  But that is really material for another round of cleanup, maybe changing the way we generate that dict instead of re-using it elsewhere.

Please go with the function to not add technical debt here.

canopen/objectdictionary/eds.py
acolomb: What I really meant was to fix it now.  But of course it can be another PR, to keep this focused.  You up to it, or should I?

friederschueler (commented):
As the old importer couldn't use decimal format for signed limits, but we exported decimal there was definitively a bug. Now we also export in hex format by default, which is more common in eds I think. We could change that, or we could remember on import the exact format and then export it in that format, but this seems like an overkill.

canopen/objectdictionary/eds.py
friederschueler: Ok, I fixed all the relevant places in eds.py

canopen/objectdictionary/eds.py
friederschueler: ok

acolomb:
Thanks for drilling deeper, the incomplete number parsing really was a nuisance.  I don't know which is more common, hex or decimal for the limit values in EDS files.  My random sample shows both in use, but never hexadecimal coding for negative limits.

But definitely, any format with a leading minus sign is easier to grasp than writing two's complement hex numbers.  Since that is very uncommon for hexadecimal numbers though, I'd favor defaulting to plain decimal notation on export.  CiA 306 never mentions negative numbers, let alone writing any number in two's complement.  But we can keep this special parsing feature of course.

Need another slot of free time to finish reviewing, sorry for this incomplete feedback...

canopen/objectdictionary/eds.py
```diff
@@ -245,6 +256,18 @@ def _convert_variable(node_id, var_type, value):
             return int(value, 0)
 
 
+def _int_to_hex(data_type: int, value: int) -> str:
+    """Format an integer as EDS hex string.
+
+    Signed types with a negative value are written as two's-complement hex
+    (e.g. INTEGER8 -1 → 0xFF) so the output is a valid EDS literal.
```
acolomb: According to CiA 306:

> Integer numbers shall be written as decimal numbers, hexadecimal numbers or octal numbers.

I don't see any need to restrict ourselves to writing hex or two's complement at all.  We may still accept it, though, as the data type rules are very clear on the possible interpretation.  But the spec doesn't mention this at all.

It's also completely unclear how to handle floating point numbers in EDS files.  I guess if we can parse them in Python, we can accept them.  But I wouldn't assume any particular conversion to apply for a literal in hex notation:  It's fine to write `-0xFFFF.FF` as a theoretical floating-point number noted in hex, just that Python only parses decimal floats.

```python
>>> float(int("-0xFFFF", 16))  # this works fine, but not directly
-65535.0
```

Just because CiA 301 defines a wire format for REAL types, doesn't mean we should try parsing hexadecimal byte sequences as IEEE floats.  When exporting, the best approach I can see is to simply use Python's float to string conversion.  This function gets in the way though, prohibiting us from writing anything but integer limits at all.  Why not simply use `_revert_variable()` for the limits as well?  The fall-back (integer) case in there can still do some limit checking, but no hex conversion.

acolomb (changes requested):

canopen/objectdictionary/eds.py
```diff
@@ -204,15 +204,12 @@ def import_from_node(node_id: int, network: canopen.network.Network):
...
-        return 64
+def _calc_bit_length(data_type: int) -> int:
+    if data_type in datatypes.SIGNED_TYPES:
+        st = ODVariable.STRUCT_TYPES[data_type]
+        if isinstance(st, datatypes.IntegerN):
+            return st.width
+        return st.size * 8
     else:
         raise ValueError(
             f"Invalid data_type '{data_type}', expecting a signed integer data_type."
```
acolomb: Works equally well for unsigned integers (and could be used for floats if needed):

```suggestion
    if data_type in datatypes.INTEGER_TYPES:
        st = ODVariable.STRUCT_TYPES[data_type]
        if isinstance(st, datatypes.IntegerN):
            return st.width
        return st.size * 8
    else:
        raise ValueError(
            f"Invalid data_type 0x{data_type:04X}, expecting an integer data_type."
```

acolomb (commented):
I took the liberty of rearranging the `_signed_int_from_hex()` logic a bit, as I was confused by it during review.

Note that a negative literal may also be hexadecimal: `int('-0xFF', 0) == -255`.  Thus I adjusted the comment accordingly.

friederschueler:
Hopefully all points correctly addressed.

acolomb (approved):

acolomb (changes requested):
Actually, I have some more points to look at in detail...

friederschueler:
I will check them after my vacation.

canopen/objectdictionary/eds.py
```diff
@@ -414,9 +434,9 @@ def export_variable(var, eds):
         eds.set(section, "PDOMapping", hex(var.pdo_mappable))
 
         if getattr(var, 'min', None) is not None:
-            eds.set(section, "LowLimit", var.min)
+            eds.set(section, "LowLimit", _revert_variable(var.data_type, var.min))
```
acolomb:
```suggestion
            eds.set(section, "LowLimit", var.min)
```

canopen/objectdictionary/eds.py
```diff
@@ -414,9 +434,9 @@ def export_variable(var, eds):
...
         if getattr(var, 'max', None) is not None:
-            eds.set(section, "HighLimit", var.max)
+            eds.set(section, "HighLimit", _revert_variable(var.data_type, var.max))
```
acolomb:
```suggestion
            eds.set(section, "HighLimit", var.max)
```

canopen/objectdictionary/eds.py
```diff
@@ -248,14 +259,14 @@ def _convert_variable(node_id: int, var_type: int, value: Any) -> Any:
...
-    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN):
+    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN) and isinstance(value, bytes):
         return bytes.hex(value)
     elif var_type in (datatypes.VISIBLE_STRING, datatypes.UNICODE_STRING):
         return value
     elif var_type in datatypes.FLOAT_TYPES:
         return value
     else:
-        return f"0x{value:02X}"
+        return str(value)
```
acolomb:
```suggestion
        return f"0x{value:02X}"
```

canopen/objectdictionary/eds.py
```diff
@@ -248,14 +259,14 @@ def _convert_variable(node_id: int, var_type: int, value: Any) -> Any:
 def _revert_variable(var_type: int, value: Any) -> Any:
     if value is None:
         return None
-    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN):
+    if var_type in (datatypes.OCTET_STRING, datatypes.DOMAIN) and isinstance(value, bytes):
```
acolomb: I was really confused why this is necessary to avoid a test failure now.  Here's my analysis:

1. The sample.eds has object 2020 with `DataType=0x40` and `LowLimit=0x3`.
2. `build_variable()` sees it's an unknown data type without a corresponding `DefaultValue` in a section `40sub1`.
3. The `data_type` is forced to DOMAIN.
4. `var.min` is correctly parsed from the `LowLimit` to an `int(3)`.
5. During export, `export_variable` now tries to pass the `int` to `_revert_variable()`.
6. For a DOMAIN variable, the value is expected to be of type `bytes`, so the string conversion via `bytes.hex()` fails with the integer argument.

**Why exactly do we need to call `_revert_variable()` at all on the limit values?  We didn't do that before.**

This function is terribly named by the way, I cannot even imagine whether it should be applied to the limits (which are always numeric).  But as it mirrors `_convert_variable()`, one should assume that it requires `bytes` for a DOMAIN variable unconditionally.  Simply skipping to a `str()` conversion seems wrong.

canopen/objectdictionary/eds.py
```diff
@@ -309,7 +320,10 @@ def build_variable(
             else:
                 var.min = int(min_string, 0)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid LowLimit %r for %s (0x%X), ignoring",
+                eds.get(section, "LowLimit"), var.name, var.index,
```
acolomb:
```suggestion
                "Invalid LowLimit %r for %s (0x%X), ignoring", min_string, var.name, var.index
```

canopen/objectdictionary/eds.py
```diff
@@ -318,38 +332,44 @@ def build_variable(
             else:
                 var.max = int(max_string, 0)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid HighLimit %r for %s (0x%X), ignoring",
+                eds.get(section, "HighLimit"), var.name, var.index,
```
acolomb:
```suggestion
                "Invalid HighLimit %r for %s (0x%X), ignoring", max_string, var.name, var.index
```

canopen/objectdictionary/eds.py
```diff
@@ -318,38 +332,44 @@ def build_variable(
...
         try:
             var.default_raw = eds.get(section, "DefaultValue")
             if '$NODEID' in var.default_raw:
                 var.relative = True
             var.default = _convert_variable(node_id, var.data_type, var.default_raw)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid DefaultValue %r for %s (0x%X), ignoring",
+                eds.get(section, "DefaultValue"), var.name, var.index,
```
acolomb:
```suggestion
                var.default_raw, var.name, var.index,
```

canopen/objectdictionary/eds.py
```diff
@@ -318,38 +332,44 @@ def build_variable(
...
+            )
     if eds.has_option(section, "ParameterValue"):
         try:
             var.value_raw = eds.get(section, "ParameterValue")
             var.value = _convert_variable(node_id, var.data_type, var.value_raw)
         except ValueError:
-            pass
+            logger.warning(
+                "Invalid ParameterValue %r for %s (0x%X), ignoring",
+                eds.get(section, "ParameterValue"), var.name, var.index,
```
acolomb:
```suggestion
                var.value_raw, var.name, var.index,
```

acolomb (changes requested):
Sorry for the late change of heart.  While trying to verify what the current scope of the PR is (to adjust the title), I re-read the code and noticed the questionable change in `_revert_variable()`, see below.

The other suggestions are related to avoiding repeated `eds.get()` calls, re-using the retrieved value.  I have some more cleanups in mind in that area, but let's keep this focused.

I have also prepared some more tests to cover the added warning logs, will push soon.  It's kind of strange that we only validate the limits for signed integers, but that's for another PR.

canopen/objectdictionary/eds.py
acolomb: This additional restriction is not needed when not passing limit values through the function (as it was before the PR).  Pushed a commit reverting this hunk.

acolomb (approved):

PR #604, opened 2025-07-21
Add unit tests for emcy.py (coverage improved from 73% → 99%)

yuvraj-kolkar17:
## What type of PR is this?

/kind test  
/area emcy.py

---

## What this PR does / why we need it:

Adds a complete unit test suite for the emcy.py component under `canopen/emcy.py`.
This test ensures core EMCY functionalities work correctly and helps future development by validating key scenarios.

---

## Which issue(s) this PR fixes:

Related to #514

---

image

## Does this PR introduce a user-facing change?

```text
NONE

PR #659, opened 2026-05-06
emcy: Catch callback exceptions and expand test coverage

friederschueler:
## What type of PR is this?

/kind bugfix test
/area emcy.py

---

## What this PR does / why we need it

This PR incorporates the review feedback from #604 (by @yuvraj-kolkar17) as suggested by @acolomb, since the original contributor has been unresponsive for several months.

**Bug fix in `canopen/emcy.py`:**
An exception raised in one EMCY callback would silently abort all subsequent callbacks in the list. The `on_emcy()` loop now wraps each callback in `try/except` and logs the exception, so remaining callbacks are always called.

**Test improvements in `test/test_emcy.py`:**
- Convert leading comment to docstring in `test_emcy_consumer_on_emcy`
- Remove superfluous inline `# Dispatch …` comments
- `test_emcy_consumer_initialization`: verify initial state (without testing `threading.Condition` type as an implementation detail)
- `test_emcy_consumer_multiple_callbacks`: simplified using lambdas
- `test_emcy_consumer_callback_exception_handling`: rewritten with lambdas; adds assertion `assertEqual(['success1', 'success2'])` that verifies the fix
- `test_emcy_consumer_error_reset_variants`: various reset code patterns
- `test_emcy_consumer_wait_timeout_edge_cases`: zero/near-zero timeouts
- `test_emcy_consumer_wait_concurrent_errors`: multiple concurrent errors
- `TestEmcyIntegration`: producer-consumer roundtrip tests using `network.subscribe()` instead of manual `rxbus.recv()` calls
- Removed tests that only exercised internal implementation details or Python's own exception mechanism

---

## Which issue(s) this PR fixes

Closes #604

---

## Does this PR introduce a user-facing change?

```text
NONE
```

### DIFF
--- a/test/test_emcy.py
+++ b/test/test_emcy.py
@@ -1,6 +1,7 @@
 import logging
 import threading
 import unittest
+from contextlib import contextmanager
 
 import can
 
@@ -9,6 +10,17 @@ import canopen.emcy
 
 
 TIMEOUT = 0.1
+
+
+@contextmanager
+def mock_rx_thread(consumer: canopen.emcy.EmcyConsumer, func):
+    t = threading.Thread(target=func)
+    try:
+        with consumer.emcy_received:
+            t.start()
+            yield t
+    finally:
+        t.join(TIMEOUT)
 
 
 class TestEmcy(unittest.TestCase):
@@ -86,40 +98,28 @@ class TestEmcy(unittest.TestCase):
         self.assertIsNone(emcy.wait(timeout=TIMEOUT))
 
         # Check unfiltered wait, on success.
-        t = threading.Thread(target=push_err)
-        with self.assertLogs(level=logging.INFO):
-            with emcy.emcy_received:
-                t.start()
-                err = emcy.wait(timeout=TIMEOUT)
-        t.join(TIMEOUT)
-        check_err(err)
+        with (
+            self.assertLogs(level=logging.INFO),
+            mock_rx_thread(emcy, push_err),
+        ):
+            check_err(emcy.wait(timeout=TIMEOUT))
 
         # Check filtered wait, on success.
-        t = threading.Thread(target=push_err)
-        with self.assertLogs(level=logging.INFO):
-            with emcy.emcy_received:
-                t.start()
-                err = emcy.wait(0x2001, TIMEOUT)
-        t.join(TIMEOUT)
-        check_err(err)
+        with (
+            self.assertLogs(level=logging.INFO),
+            mock_rx_thread(emcy, push_err),
+        ):
+            check_err(emcy.wait(0x2001, TIMEOUT))
 
         # Check filtered wait, on timeout.
-        t = threading.Thread(target=push_err)
-        with emcy.emcy_received:
-            t.start()
-            result = emcy.wait(0x9000, TIMEOUT)
-        t.join(TIMEOUT)
-        self.assertIsNone(result)
+        with mock_rx_thread(emcy, push_err):
+            self.assertIsNone(emcy.wait(0x9000, TIMEOUT))
 
         def push_reset():
             emcy.on_emcy(0x81, b'\x00\x00\x00\x00\x00\x00\x00\x00', 100)
 
-        t = threading.Thread(target=push_reset)
-        with emcy.emcy_received:
-            t.start()
-            result = emcy.wait(0x9000, TIMEOUT)
-        t.join(TIMEOUT)
-        self.assertIsNone(result)
+        with mock_rx_thread(emcy, push_reset):
+            self.assertIsNone(emcy.wait(0x9000, TIMEOUT))
 
     def test_emcy_consumer_multiple_callbacks(self):
         """Test adding multiple callbacks and their execution order."""
@@ -172,12 +172,12 @@ class TestEmcy(unittest.TestCase):
             emcy.on_emcy(0x81, b'\x01\x20\x01\x01\x02\x03\x04\x05', 100)
             emcy.on_emcy(0x81, b'\x02\x20\x01\x01\x02\x03\x04\x05', 101)
             emcy.on_emcy(0x81, b'\x03\x20\x01\x01\x02\x03\x04\x05', 102)
-        t = threading.Thread(target=push_multiple_errors)
-        with self.assertLogs(level=logging.INFO):
-            with emcy.emcy_received:
-                t.start()
-                err = emcy.wait(0x2003, timeout=TIMEOUT)
-        t.join(TIMEOUT)
+
+        with (
+            self.assertLogs(level=logging.INFO),
+            mock_rx_thread(emcy, push_multiple_errors),
+        ):
+            err = emcy.wait(0x2003, timeout=TIMEOUT)
         self.assertIsNotNone(err)
         self.assertEqual(err.code, 0x2003)
 
@@ -327,13 +327,14 @@ class TestEmcyIntegration(unittest.TestCase):
         """Test that producer and consumer work together."""
         received_errors = []
         self.consumer.add_callback(lambda err: received_errors.append(err))
-        t = threading.Thread(
-            target=lambda: self.producer.send(0x2001, 0x02, b'\x01\x02\x03\x04\x05'),
-        )
-        with self.assertLogs(level=logging.INFO):
-            t.start()
+        with (
+            self.assertLogs(level=logging.INFO),
+            mock_rx_thread(
+                self.consumer,
+                lambda: self.producer.send(0x2001, 0x02, b'\x01\x02\x03\x04\x05'),
+            ),
+        ):
             err = self.consumer.wait(0x2001, timeout=TIMEOUT)
-        t.join(TIMEOUT)
         self.assertIsNotNone(err)
         self.assertEqual(err.code, 0x2001)
         self.assertEqual(err.register, 0x02)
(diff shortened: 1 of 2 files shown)


acolomb:
It would really help the review process if you based your version on top of the commits in #604.  Is that feasible?  It will be squashed anyway when merging.

friederschueler:
I really tried to rebase this commit to #604 and then apply the changes, but even with AI this didn't work.

Instead here is a summary of how #659 addresses your specific review comments on #604:

| Your review comment on #604 | Addressed in #659 |
|---|---|
| Convert comment to docstring | ✅ Done |
| Superfluous `# Dispatch …` comments | ✅ Removed |
| `emcy_received` type is an implementation detail | ✅ Removed |
| Simplify callbacks with lambdas | ✅ Done |
| `len(callbacks)` is not public API | ✅ Removed |
| `successful_callbacks` unused — check actual behavior | ✅ `assertEqual(['success1', 'success2'])` |
| "Avoid artificial delays" (`time_expiry_during_execution`) | ✅ Removed |
| Human-readable string checks are fragile | ✅ Only `assertIn(f"0x{code:04X}", s)` |
| Why not use `network.subscribe()`? | ✅ Integration tests use `rx_net.subscribe()` |
| Redundant tests (`struct_packing`, `network_assignment`, `emcy_error_inheritance`) | ✅ Removed |

Additionally, #659 includes the actual **bug fix in `emcy.py`** (try/except around each callback) which #604 does not.

test/test_emcy.py
```diff
@@ -123,6 +120,65 @@ def push_reset():
             t.start()
             self.assertIsNone(self.emcy.wait(0x9000, TIMEOUT))
 
+    def test_emcy_consumer_initialization(self):
+        consumer = canopen.emcy.EmcyConsumer()
+        self.assertEqual(consumer.log, [])
+        self.assertEqual(consumer.active, [])
+        self.assertEqual(consumer.callbacks, [])
+
```
acolomb: What's this trying to prove?  See also https://github.com/canopen-python/canopen/pull/604#discussion_r2237762353

test/test_emcy.py
```diff
@@ -123,6 +120,65 @@ def push_reset():
...
+        self.emcy.add_callback(lambda err: call_order.append('callback1'))
+        self.emcy.add_callback(lambda err: call_order.append('callback2'))
+        self.emcy.add_callback(lambda err: call_order.append('callback3'))
+        self.emcy.on_emcy(0x81, b'\x01\x20\x02\x00\x01\x02\x03\x04', 1000)
+        self.assertEqual(call_order, ['callback1', 'callback2', 'callback3'])
+
+    def test_emcy_consumer_callback_exception_handling(self):
+        """Test that callback exceptions don't break other callbacks or the system."""
+        successful_callbacks = []
+        self.emcy.add_callback(lambda err: successful_callbacks.append('success1'))
```
acolomb: Please review for places where the common `self.emcy` object is piling up callbacks over several tests.  While it doesn't affect the test result, mutating the common class state in each test seems error-prone.

test/test_emcy.py
```diff
@@ -123,6 +120,65 @@ def push_reset():
...
+
+    def test_emcy_consumer_wait_concurrent_errors(self):
+        """Test wait method when multiple errors arrive concurrently."""
+        def push_multiple_errors():
+            self.emcy.on_emcy(0x81, b'\x01\x20\x01\x01\x02\x03\x04\x05', 100)
+            self.emcy.on_emcy(0x81, b'\x02\x20\x01\x01\x02\x03\x04\x05', 101)
+            self.emcy.on_emcy(0x81, b'\x03\x20\x01\x01\x02\x03\x04\x05', 102)
+        t = threading.Timer(TIMEOUT / 2, push_multiple_errors)
+        with self.assertLogs(level=logging.INFO):
+            t.start()
```
acolomb: This was not addressed: https://github.com/canopen-python/canopen/pull/604#discussion_r2240931783

Isn't there a way to parallelize waiting and sending, without artificially waiting before send?  What's important is only the execution order (start wait, then push).

Same thing below in the integration tests.

test/test_emcy.py
```diff
@@ -180,6 +236,26 @@ def check(code, expected):
...
+        self.assertEqual(error.code, 0x1000)
+        self.assertEqual(error.register, 0)
+        self.assertEqual(error.data, b'')
+        self.assertEqual(error.timestamp, 123.456)
+        error = EmcyError(0xFFFF, 0xFF, b'\xFF' * 5, float('inf'))
+        self.assertEqual(error.code, 0xFFFF)
+        self.assertEqual(error.register, 0xFF)
+        self.assertEqual(error.data, b'\xFF' * 5)
+        self.assertEqual(error.timestamp, float('inf'))
+
```
acolomb: What's proven here?  Given the simplicity of the constructor, I don't see any value in this test.

test/test_emcy.py
```diff
@@ -180,6 +236,26 @@ def check(code, expected):
...
+    def test_emcy_error_str_edge_cases(self):
+        for code in (0x0000, 0x0001, 0x0100, 0xFFFF):
+            error = EmcyError(code, 0, b'', 1000)
+            s = str(error)
+            self.assertIsInstance(s, str)
```
acolomb: Wow.

test/test_emcy.py
```diff
@@ -220,6 +296,91 @@ def check(*args, res):
         check(3, res=b'\x00\x00\x03\x00\x00\x00\x00\x00')
         check(3, b"\xaa\xbb", res=b'\x00\x00\x03\xaa\xbb\x00\x00\x00')
 
+    def test_emcy_producer_initialization(self):
+        producer = canopen.emcy.EmcyProducer(0x123)
+        self.assertEqual(producer.cob_id, 0x123)
+        self.assertIsNotNone(producer.network)
+
```
acolomb: This does not even increase coverage, as there is a prior instantiation.  The assertions can be dropped, no value.  Hence, this whole test.

acolomb (changes requested):
This is certainly better than the last state in #604.  However, there is still quite a bit of AI slop in there, code just for the sake of having produced some code.  Please review whether the tests are of actual value and sensible.  Nothing beats human common sense here.

test/test_emcy.py
friederschueler: :D

friederschueler:
Thank you very much for your manual review and the time and effort you contributed. I will use this material as example for our internal AI discussion. I will look for my original query, but I explicitly stated that the AI should check for every test if it is really useful, as in testing the internal logic. if it provides coverage and if there are no unnecessary assertions :D
also used state of the art claude opus 4.6 agent and still you get garbage tests like this. very interesting

test/test_emcy.py
```diff
@@ -1,12 +1,11 @@
 import logging
 import threading
 import unittest
-from contextlib import contextmanager
```
acolomb, line 4: Actually I liked the contextmanager solution, not requiring an explicit join() in several places.  Let me suggest an alternative solution, integrating the contextmanager and Timer --> Thread replacement consistently.

test/test_emcy.py
```diff
@@ -1,12 +1,11 @@
...
 
 import can
 
 import canopen
-from canopen.emcy import EmcyError
+import canopen.emcy
```
acolomb: I don't see this as an improvement.

test/test_emcy.py
```diff
@@ -84,67 +81,65 @@ def check_err(err):
...
+        t.join(TIMEOUT)
         check_err(err)
 
         # Check filtered wait, on timeout.
-        with timer(push_err) as t:
+        t = threading.Thread(target=push_err)
+        with self.emcy.emcy_received:
             t.start()
-            self.assertIsNone(self.emcy.wait(0x9000, TIMEOUT))
+            result = self.emcy.wait(0x9000, TIMEOUT)
```
acolomb: Unnecessary if we stick with a context manager.

test/test_emcy.py
```diff
@@ -277,7 +254,7 @@ def tearDown(self):
 
     def check_response(self, expected):
         msg = self.rxbus.recv(TIMEOUT)
-        self.assertIsNotNone(msg)
+        assert msg is not None
```
acolomb, line 260: This can stay IMHO.  It's part of the test to see if something was received.  And it does abort further execution either way.

test/test_emcy.py
acolomb: The `.emcy_received` trick is missing here, so this is flaky.

test/test_emcy.py
```diff
@@ -366,16 +337,15 @@ def test_producer_consumer_integration(self):
 
     def test_producer_reset_consumer_integration(self):
         """Test producer reset clears consumer active errors."""
-        t = threading.Timer(
-            TIMEOUT / 2,
-            lambda: self.producer.send(0x2001, 0x02, b'\x01\x02\x03\x04\x05'),
+        t = threading.Thread(
+            target=lambda: self.producer.send(0x2001, 0x02, b'\x01\x02\x03\x04\x05'),
```
acolomb: Same here, requires the condition lock to work reliably.

test/test_emcy.py
```diff
@@ -366,16 +337,15 @@ def test_producer_consumer_integration(self):
...
         )
         with self.assertLogs(level=logging.INFO):
             t.start()
             self.consumer.wait(0x2001, timeout=TIMEOUT)
         t.join(TIMEOUT)
         self.assertEqual(len(self.consumer.active), 1)
-        t = threading.Timer(TIMEOUT / 2, self.producer.reset)
+        t = threading.Thread(target=self.producer.reset)
```
acolomb: Once more, need lock.

test/test_emcy.py
```diff
@@ -72,8 +71,6 @@ def test_emcy_consumer_reset(self):
         self.assertEqual(len(self.emcy.active), 0)
 
     def test_emcy_consumer_wait(self):
-        PAUSE = TIMEOUT / 2
-
         def push_err():
             self.emcy.on_emcy(0x81, b'\x01\x20\x01\x01\x02\x03\x04\x05', 100)
```
acolomb: These can all be switched to a local `EmcyConsumer` object. we don't actually need the instance attribute as closure.  I'd prefer that to make the tests more robust against execution order and parallelization.

acolomb (changes requested):
The solution with a Thread instead of Timer is really nice, thanks!  However, it is not used consistently, so some tests are now flaky.  I'll come up with an integrated version...

acolomb:
Note that there are actually two errors in the `EmcyConsumer.wait()` implementation.  That's for another PR though.  Noticed while checking whether two nested `with` blocks on the same thread and condition are safe.

bizfsc:
Ok, I fixed the member/local variables.
@acolomb This should be merge ready now?

acolomb:
> Ok, I fixed the member/local variables. @acolomb This should be merge ready now?

Actually, it has been ready before your last commit.  I was just waiting for your confirmation on the changes I had introduced.  The comment about local `EmcyConsumer` was about the tests for that specific class, which I had already fixed.  Of course the integration tests can use a class-wide producer and consumer object.  So I just reverted the latest change.

acolomb (approved):

acolomb:
One last question about the added `import canopen.emcy`.  I don't get why that should be necessary.  We already `import canopen` and all places using the `emcy` submodule already work.  Why do you / does pylance insist on the extra import?

PR #660, opened 2026-05-07
Add some basic rules for ruff linter and black formatting

acolomb:
Line-length 96 is a compomise because 80 is problematic with existing
code, but we want a bit of headroom for the next common limit of 100.

String "normalization" is complete noise in most cases, that could be
enabled again after some general cleanup.

Make sure that ruff agrees with the previously used import ordering
style from isort.

The last change is extracted from #644.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,10 @@ exclude = [
 
 [tool.ruff]
 line-length = 96
+[tool.ruff.lint.isort]
+case-sensitive = true
+order-by-type = false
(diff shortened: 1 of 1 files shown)


acolomb:
I guess this should help a bit with the disturbances encountered in #644 and #658.  I still want a more gradual approach to reformatting and style adjustments.  Feature work and fixes are definitely top priority, and we can get back to code style issues when the backlog has been reduced.

PR #662, opened 2026-05-12
od: Fix shared-state mutations in ODArray.__getitem__ generated subindices

bizfsc:
When `ODArray.__getitem__` auto-generates a variable from the array template for an undefined subindex, mutable attributes were assigned by reference. This means mutations to `value_descriptions` or `bit_definitions` on one generated variable would silently affect all others sharing the same template.

## Root cause

```python
for attr in (..., "value_descriptions", "bit_definitions", ...):
    if attr in template.__dict__:
        var.__dict__[attr] = template.__dict__[attr]  # reference, not a copy
```

## Fix

Copy the two affected attributes independently:
- `value_descriptions: dict[int, str]` — shallow copy (`.copy()`) is sufficient since values are immutable `str`
- `bit_definitions: dict[str, list[int]]` — `copy.deepcopy()` required since values are `list[int]`

## Tests

- `test_generated_subindex_value_descriptions_independent` — mutating `value_descriptions` on one generated subindex does not affect another or the template
- `test_generated_subindex_bit_definitions_independent` — mutating a list inside `bit_definitions` on one generated subindex does not affect another or the template

## Related

Spotted while reviewing #653 (`custom_options` has the same shape and was already fixed there separately).

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -4,6 +4,7 @@ Object Dictionary module
 
 from __future__ import annotations
 
+import copy
 import logging
 import struct
 from collections.abc import Iterator, Mapping, MutableMapping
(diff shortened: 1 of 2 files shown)


bizfsc:
I need this bugfix to test https://github.com/canopen-python/canopen/pull/454. All my bit_definitions are shared otherwise.

acolomb:
Without going deep into details (limited time right now), why exactly would each array entry need its own independent version of these attributes?  Isn't the purpose of an array to hold entries with a common schema?  Maybe your "container" object should be an `ODRecord` instead?

Side-note: Mutating `__dict__` on instances here is really ugly, we should use `getattr()` / `setattr()` instead.

bizfsc:
You are right, with different attributes I should use an ODRecord instead. I was thinking about mapping different BOOLs for a config-like object from a bit-sequence. There is no use-case for different attributes in an array.

BUT if someone (like me) creates the ODArray manually and then changes a subindex like:

`arr[5].bit_definitions["extra"] = [3]`

It will set it for all items, not just arr[5] which is unexpected I think.

acolomb:
Maybe unexpected, but still reasonable.  Not quite sure where we could put a documentation hint to that effect.

PR #663, opened 2026-05-12
Correct type handling for bitwise variable access

acolomb:
Contrary to the existing type hints, the `decode_bits()` / `encode_bits()` methods can handle strings for lookup as well as an arbitrary collection of integer bit offsets. Correct the hints to reflect that and check at runtime to only lookup from strings.

Correct return type for encode_bits(), as it works only for integers.

Document both functions to explain the intention and caveats to consider, especially the lookup KeyError exception.

Extend test to cover an iterable and a failed lookup.

### DIFF
--- a/canopen/variable.py
+++ b/canopen/variable.py
@@ -179,7 +179,10 @@ class Bits(Mapping):
     @staticmethod
     def _get_bits(key: Union[slice, int, str, Collection[int]]) -> Union[str, Collection[int]]:
         if isinstance(key, slice):
-            return range(key.start, key.stop, key.step)
+            if key.stop is None:
+                raise IndexError("Bits cannot be enumerated from open-ended slice")
+            else:
+                return range(key.start, key.stop, key.step or 1)
         if isinstance(key, int):
             return [key]
         return key
--- a/canopen/variable.py
+++ b/canopen/variable.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 import logging
-from collections.abc import Mapping
+from collections.abc import Collection, Mapping
 from typing import Union
 
 from canopen import objectdictionary
@@ -118,7 +120,7 @@ class Variable:
         self.raw = self.od.encode_desc(desc)
 
     @property
-    def bits(self) -> "Bits":
+    def bits(self) -> Bits:
         """Access bits using integers, slices, or bit descriptions."""
         return Bits(self)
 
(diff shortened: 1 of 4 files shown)


acolomb:
This should properly fix some issues that #651 only silences.

acolomb:
Note that there is a logic error in the existing encoder function, caught by the test in commit a0ffffb and fixed in commit dcf7bf4.

canopen/variable.py
```diff
@@ -169,23 +171,26 @@ def write(
 class Bits(Mapping):
 
     def __init__(self, variable: Variable):
+        assert variable.od.data_type in objectdictionary.datatypes.INTEGER_TYPES
```
bizfsc, line 174: As I learned like five minutes ago, it seems that -O or -OO flags (mean running python in optimized mode) will disable assertions and thus it is always better to use if / raise. Personally I like the inline assertions a lot and use them quite often. Also libs like pydantic use them in business logic. AI complains you should only use assertions in tests and otherwise raise an appropriate exception (probably TypeError).

canopen/variable.py
```diff
@@ -169,23 +171,26 @@ def write(
...
-            bits = [key]
-        else:
-            bits = key
-        return bits
-
-    def __getitem__(self, key) -> int:
+            if key.stop is None:
+                raise IndexError("Bits cannot be enumerated from open-ended slice")
+            else:
+                return range(key.start, key.stop, key.step or 1)
```
bizfsc: slice(3) will lead to slice(None, 3, None) which will lead to TypeError in the range.

`range(key.start or 0, key.stop, key.step or 1)`

will fix it.

bizfsc (approved):

canopen/variable.py
acolomb: Thanks for catching this important ommission!  Fixed.

canopen/variable.py
acolomb: Thanks for the pointer, I didn't know either how easy it is to skip them.

The real question here is what API guarantees we want to give.  Bitwise access for a non-integer object is a programming error.  Catching that early and reporting in a meaningful way is important during development (of the application I mean, not the library).  But once the application straightens out its use of objects and the corresponding types, I see nothing wrong with skipping this check for performance, if optimization is requested by the user.  There will be an exception at some point, but disabling assertions on unverified code is basically asking for less useful error diagnostics.

PR #667, opened 2026-05-21
variable: Fix type-related issues with desc property

acolomb:
Bundle some fixes related to #651 and the variable data access through the `desc` property or `write(fmt="desc", ...)`.  This takes a different approach, adding some runtime checks instead of squelching mypy errors.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -500,11 +500,10 @@ class ODVariable:
     def decode_desc(self, value: int) -> str:
         if not self.value_descriptions:
             raise ObjectDictionaryError("No value descriptions exist")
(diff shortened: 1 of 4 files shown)


bizfsc (approved):
Walrus operator ":=" use seems correct here.

acolomb:
@bizfsc Thanks for reviewing. You apparently haven't accepted the invitation for your biz user to this organisation. Thus the review is not counted. Will you join the org with both users, or should I just explicitly assign premissions to the biz user (if that's possible)?

bizfsc:
@acolomb I have accepted the invitation, but it seems there is still some permission problem. If you can, try to set the rights manually.

acolomb:
Thanks @bizfsc, worked now!

PR #668, opened 2026-05-26
emcy: Fix maximum timeout logic errors in EmcyConsumer.wait()

acolomb:
If an EMCY package was received, but filtered out by the emcy_code matching, the condition waiting is started again with the same timeout.  Thus the actual maximum waiting time is not correctly limited to the given argument, as the docstring promises.  Track the remaining time until the initial deadline instead, as basis for the condition wait.

Further, spurious wake-ups from the condition wait on the OS level are not handled correctly.  The threading.Condition docs explicitly recommend checking the shared state (number of logged entries in this case) in the while loop, because the wait() call may abort early.  The current code assumes that this necessarily indicates a timeout, without checking the actually passed time again.  Reorder the check against end_time in the loop to avoid this.

### DIFF
--- a/canopen/emcy.py
+++ b/canopen/emcy.py
@@ -69,16 +69,19 @@ class EmcyConsumer:
         while True:
             with self.emcy_received:
                 prev_log_size = len(self.log)
-                self.emcy_received.wait(timeout)
+                remaining = end_time - time.time()
+                if remaining <= 0:
+                    return None  # Actual timeout reached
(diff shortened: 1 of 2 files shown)


acolomb:
This was discovered while reviewing the tests in #659.  Let's merge that one first to have an accurate picture of the test coverage.  Will add some more specific tests regarding this failure mode subsequently.

PR #669, opened 2026-05-26
od: Robust path handling without type errors

acolomb:
Fix the union-attr errors squelched in #651 properly.  In the process, extend `import_od()` to gracefully handle some edge cases around missing file names and non-`str` paths in the `source` argument.

Extend unit tests around import and export functions.  Note that these are still a bit convoluted because the EDS import is tested through the generic, format-agnostic OD API.  This is still much easier than splitting tests between generic and format-specific though.  If ever someone adds unit tests covering the EPF import, that might need some refactoring.

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 import os
 import struct
+import sys
 from collections.abc import Collection, Iterator, Mapping, MutableMapping
 from typing import Optional, TextIO, Union
 
@@ -50,7 +51,9 @@ def export_od(
 
     opened_here: Optional[TextIO] = None
     try:
-        if isinstance(dest, (str, os.PathLike)):
+        if dest is None:
+            dest = sys.stdout
+        elif isinstance(dest, (str, os.PathLike)):
             if doc_type is None:
                 _, suffix = os.path.splitext(os.fspath(dest).lower())
                 for t in supported_doctypes:
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 def export_od(
     od: ObjectDictionary,
     dest: Union[str, TextIO, None] = None,
-    doc_type: Optional[str] = None
+    doc_type: Optional[str] = None,
 ) -> None:
     """Export an object dictionary.
 
@@ -47,7 +47,7 @@ def export_od(
             f"supported formats: {supported}"
         )
 
-    opened_here = False
+    opened_here: Optional[TextIO] = None
     try:
         if isinstance(dest, str):
             if doc_type is None:
@@ -58,7 +58,7 @@ def export_od(
                 else:
                     doc_type = "eds"
             dest = open(dest, 'w')
-            opened_here = True
+            opened_here = dest
 
         if doc_type == "eds":
             from canopen.objectdictionary import eds
@@ -68,8 +68,8 @@ def export_od(
             return eds.export_dcf(od, dest)
     finally:
         # If dest is opened in this fn, it should be closed
-        if opened_here:
-            dest.close()
+        if opened_here is not None:
+            opened_here.close()
 
 
 def import_od(
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -69,7 +69,7 @@ def export_od(
             from canopen.objectdictionary import eds
             return eds.export_dcf(od, dest)
     finally:
-        # If dest is opened in this fn, it should be closed
+        # If dest is opened in this function, it should be closed
         if opened_here is not None:
             opened_here.close()
 
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -5,6 +5,7 @@ Object Dictionary module
 from __future__ import annotations
 
 import logging
+import os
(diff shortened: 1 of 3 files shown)


bizfsc (changes requested):
A few suggestions below.

**canopen/objectdictionary/__init__.py**

the standard library already has the right abstraction for this. contextlib.nullcontext wraps a caller-owned object in a no-op context manager, so the try/finally and the flag variable can be replaced with a plain with statement:
```python
    import contextlib

    if isinstance(dest, (str, os.PathLike)):
        if doc_type is None:
            _, suffix = os.path.splitext(os.fspath(dest).lower())
            for t in supported_doctypes:
                if suffix == f".{t}":
                    doc_type = t
                    break
            else:
                doc_type = "eds"
        # A path was given: open the file here. The with statement
        # guarantees it is closed on exit, including on exceptions.
        ctx = open(dest, 'w')
    else:
        # A file-like object or None (stdout) was passed in. The caller owns
        # its lifecycle; use a no-op context manager so the with block below
        # works uniformly without closing the caller's object.
        ctx = contextlib.nullcontext(dest)

    with ctx as dest:
        if doc_type == "eds":
            from canopen.objectdictionary import eds
            return eds.export_eds(od, dest)
        elif doc_type == "dcf":
            from canopen.objectdictionary import eds
            return eds.export_dcf(od, dest)
```

**test/test_eds.py**

test_export_eds_auto_close has three issues:

First, canopen.export_od(self.od, m_open) passes a callable Mock as dest. This works by accident because Mock satisfies any hasattr check, but it does not actually test the intended behaviour.

Second, fd = m_open() calls the mock again, adding an unwanted call to its call_count. Use m_open.return_value instead.

Third, with the contextlib approach above the with statement calls __exit__, not close() directly. The assertion should match:
```python
    def test_export_eds_auto_close(self):
        # File-like object passed in directly: must NOT be closed by export_od
        with io.StringIO() as dest:
            canopen.export_od(self.od, dest, "eds")
            self.assertFalse(dest.closed)
        for path in ("mock.eds", pathlib.Path("mock.eds")):
            with self.subTest(path=path):
                m_open = mock_open()
                with patch("canopen.objectdictionary.open", m_open):
                    canopen.export_od(self.od, path)
                # File object opened at path must be closed before return
                m_open.return_value.__exit__.assert_called_once()

    def test_export_eds_auto_close_exception(self):
        m_open = mock_open()
        m_open.return_value.write.side_effect = IOError("Simulated write failure")
        with patch("canopen.objectdictionary.open", m_open), self.assertRaises(IOError):
            canopen.export_od(self.od, "mock.eds")
        # File object opened at path must be closed even when an exception is raised
        m_open.return_value.__exit__.assert_called_once()
```

Note: in the original test_export_eds_auto_close_exception the assert_called_once() call is inside the assertRaises block, which means it is unreachable when no exception occurs and silently does nothing when one is raised. It must be outside.

test_load_unsupported_format: remove the canopen.import_od(object()) case. object() is outside the declared signature and the behaviour (returning "" from getattr fallback) is an implementation detail, not a contract. If a caller passes a wrong type, that is their problem.

Finally, add a positive test for pathlib.Path in import_od. The PR only adds a not-found error case. Add:
```python
    def test_load_pathlib_path(self):
        od = canopen.import_od(pathlib.Path(SAMPLE_EDS))
        self.assertTrue(len(od) > 0)
```

acolomb:
Nice idea with the nullcontext.  I tried incorporating it, but that causes some type mypy errors.  I need a bit more time to investigate, also for the tests.

acolomb:
Looking at the nullcontext variant again, I think it's not very clear, compared to the proposed PR code:

```python3
    if dest is None:
        ctx: contextlib.AbstractContextManager[TextIO] = contextlib.nullcontext(sys.stdout)
    elif isinstance(dest, (str, os.PathLike)):
        if doc_type is None:
            _, suffix = os.path.splitext(os.fspath(dest).lower())
            for t in supported_doctypes:
                if suffix == f".{t}":
                    doc_type = t
                    break
            else:
                doc_type = "eds"
        ctx = open(dest, 'w')
    else:
        ctx = contextlib.nullcontext(dest)

    with ctx as dest:
        if doc_type == "eds":
            from canopen.objectdictionary import eds
            return eds.export_eds(od, dest)
        elif doc_type == "dcf":
            from canopen.objectdictionary import eds
            return eds.export_dcf(od, dest)
```

This one is cleanly accepted by mypy.  But the type annotation alone isn't pretty nor readable.  In contrast, the current proposal doesn't even need comments, as the variable name speaks for itself.

It also goes one step further and moves opening of `sys.stdout` to the generic routine instead of passing `None` to the format-specific variant to trigger the switch there.  I think that's a cleaner approach, anyone using the format-specific functions directly can still just pass in `sys.stdout` themselves.  The generic method handles stream opening, while the lower level simply formats into a provided stream.

acolomb:
I didn't understand all the issues you noted about `mock_open()` at first, but it got me thinking.  Thank you for the hints.  In this case that function isn't really a good approach at all.  Because what we really want to test is whether a file-like object is properly closed _only if_ it was opened internally.  No matter whether `close()` is used internally, or `__exit__()`.

I took a new stab at it and now it uses `io.StringIO` everywhere to provide a realistic file-like object where we can also examine the open / closed state, not just look at what methods were called.  That's much easier and more in line with the other tests.

acolomb:
> test_load_unsupported_format: remove the canopen.import_od(object()) case.
> Finally, add a positive test for pathlib.Path in import_od.

Both suggestions added, thank you!

PR #672, opened 2026-05-29
nmt: Fix NmtSlave on a node without a heartbeat producer time

acolomb:
The NMT state transition from INITIALISING to PRE-OPERATIONAL causes the heartbeat to start automatically if the parameter is non-zero in object 0x1017:00.  This check however raises an exception if the object does not even exist.

Cover this state in tests and explicitly ignore a KeyError when asking for the object.


bizfsc (approved):
Minimal implementation to fix bug. approved

PR #673, opened 2026-06-04
od: Extend physical conversion to integer factors, fix typing

acolomb:
The `Variable.phys` property docstring mentions "Non integers will be
passed as is."  That is correct, but not reflected in the type hints
for the associated `ODVariable.encode_phys()` / `.decode_phys()`
functions.  Extend the return / argument type to all possible internal
data types, instead of plain `int`.

Try factor conversion on all `NUMBER_TYPES`, where it was previously
limited to `INTEGER_TYPES`.  Only apply rounding in `encode_phys()` if the
variable's data type is actually integer-based.

Allow integer factors in type hint for `ODVariable.factor`.  There is
nothing wrong with using integer factors to yield integers
when decoding.  For encoding, the division may still result in a
`float`, while the rounding behavior is kept unchanged for integer-typed
variables.

Thus, try to parse integer factors as integers from EDS files, not forcing
to a `float`.

Related comment first given here: https://github.com/canopen-python/canopen/pull/651#issuecomment-4386512596

### DIFF
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -354,7 +354,7 @@ class ODVariable:
         #: Physical unit
         self.unit: str = ""
         #: Factor between physical unit and integer value
-        self.factor: float = 1
+        self.factor: Union[int, float] = 1
         #: Minimum allowed value
         self.min: Optional[int] = None
         #: Maximum allowed value
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -489,7 +489,7 @@ class ODVariable:
     def decode_phys(
         self, value: Union[int, bool, float, str, bytes]
     ) -> Union[int, bool, float, str, bytes]:
-        if self.data_type in INTEGER_TYPES:
+        if self.data_type in NUMBER_TYPES:
             assert isinstance(value, (int, float))
             value *= self.factor
         return value
@@ -497,10 +497,12 @@ class ODVariable:
     def encode_phys(
         self, value: Union[int, bool, float, str, bytes]
     ) -> Union[int, bool, float, str, bytes]:
-        if self.data_type in INTEGER_TYPES:
+        if self.data_type in NUMBER_TYPES:
             assert isinstance(value, (int, float))
             if self.factor != 1:
-                value = round(value / self.factor)
+                value = value / self.factor
+            if self.data_type in INTEGER_TYPES:
+                value = round(value)
         return value
 
     def decode_desc(self, value: int) -> str:
--- a/canopen/objectdictionary/__init__.py
+++ b/canopen/objectdictionary/__init__.py
@@ -486,13 +486,19 @@ class ODVariable:
             raise TypeError(
                 f"Do not know how to encode {value!r} to data type 0x{self.data_type:X}")
 
-    def decode_phys(self, value: int) -> Union[int, bool, float, str, bytes]:
+    def decode_phys(
+        self, value: Union[int, bool, float, str, bytes]
+    ) -> Union[int, bool, float, str, bytes]:
         if self.data_type in INTEGER_TYPES:
+            assert isinstance(value, (int, float))
             value *= self.factor
         return value
 
-    def encode_phys(self, value: Union[int, bool, float, str, bytes]) -> int:
+    def encode_phys(
+        self, value: Union[int, bool, float, str, bytes]
+    ) -> Union[int, bool, float, str, bytes]:
         if self.data_type in INTEGER_TYPES:
+            assert isinstance(value, (int, float))
             if self.factor != 1:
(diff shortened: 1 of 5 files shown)


test/test_od.py
```diff
@@ -227,6 +246,25 @@ def test_phys_float_factor_decodes_to_float(self):
...
+        var = od.ODVariable("Test INTEGER16", 0x1000)
+        var.data_type = od.INTEGER16
+        var.factor = 3
+        # 10 / 3 = 3
+        encoded = var.encode_phys(10)
+        self.assertEqual(encoded, 3)
+        self.assertIsInstance(encoded, int)
+
+    def test_phys_int_factor_decodes_to_int(self):
+        """decode_phys with float factor ensures a float result."""
```
bizfsc, line 260:
```suggestion
        """decode_phys with int factor ensures a int result."""
```

bizfsc: Docstring nit

canopen/objectdictionary/eds.py
```diff
@@ -337,9 +337,12 @@ def build_variable(
...
     if eds.has_option(section, "Factor"):
         try:
-            var.factor = float(eds.get(section, "Factor"))
+            var.factor = int(eds.get(section, "Factor"))
         except ValueError:
-            pass
+            try:
+                var.factor = float(eds.get(section, "Factor"))
+            except ValueError:
+                pass
```
bizfsc:
```suggestion
                logger.warning(
                    "Could not parse Factor for %s in section [%s], ignoring",
                    var.name, section,
                )
```

Yes, we don't have this in many places, but lets get this started. Ar least in debugging the devs can see now, that the parsing failed.

canopen/objectdictionary/__init__.py
```diff
@@ -486,15 +486,23 @@ def encode_raw(self, value: Union[int, float, str, bytes, bytearray]) -> bytes:
             raise TypeError(
                 f"Do not know how to encode {value!r} to data type 0x{self.data_type:X}")
 
-    def decode_phys(self, value: int) -> Union[int, bool, float, str, bytes]:
-        if self.data_type in INTEGER_TYPES:
+    def decode_phys(
+        self, value: Union[int, bool, float, str, bytes]
+    ) -> Union[int, bool, float, str, bytes]:
+        if self.data_type in NUMBER_TYPES:
+            assert isinstance(value, (int, float))
```
bizfsc:
```suggestion
```

decode_raw ensure correct typing. no assert in production code. see bigger comment for both functions

canopen/objectdictionary/__init__.py
bizfsc, line 1:
```python
    def decode_phys(
        self, value: Union[int, bool, float, str, bytes]
    ) -> Union[int, bool, float, str, bytes]:
        if self.data_type in NUMBER_TYPES:
            numeric = cast(Union[int, float], value)
            value = numeric * self.factor
        return value

    def encode_phys(
        self, value: Union[int, bool, float, str, bytes]
    ) -> Union[int, bool, float, str, bytes]:
        if self.data_type in NUMBER_TYPES:
            numeric = cast(Union[int, float], value)
            if self.factor != 1:
                numeric = numeric / self.factor
                numeric = round(numeric)
            value = numeric
        return value
```

Used too much brain power to solve this issue for the current temperatures. But here is the dilemma:
1. don't use assert in production
2. don't change signature
3. don't ask for permission, ask for forgiveness

=> cast is noop in runtime = zero loss and the linter is happy, because the can not see that only int or float can be passed. This should be the best solution. Also numeric / self.factor will raise natural type error for wrong types

bizfsc (changes requested):
Changed to zero runtime cost and without assert.
Pylance / mypy happy.

canopen/objectdictionary/eds.py
acolomb: I have a separate PR pending which cleans up the logging / error behaviour.  Including a "strict" mode which raises instead of logging.  Just wanted to finish up smaller steps first, so let's keep it out of this one.

canopen/objectdictionary/__init__.py
acolomb: > don't use assert in production

*Why?*  I feel like it's the perfect solution.

"In production" means that one should enable optimizations I guess, so the assertions become no-ops.  A quick AI summary (without even hinting at type checks) cooked up these three use cases as OK to use assert:
* Internal consistency checks: `assert tree.root is not None`
* Conditions that indicate programmer bugs: `assert isinstance(ast_node, Expression)` (Yay!)
* Debugging assumptions: `assert cache_entry.owner is self`

I'd rather reach a consensus here, before applying suggestions to the PR which are not clearly fixes for an oversight. :wink:

canopen/objectdictionary/eds.py
friederschueler: Agreed

PR #674, opened 2026-06-24
nmt: use bytes instead of list for send_message and send_periodic

bizfsc:
This PR extracts only the nmt.py changes from #651, which covers a broader set of mypy fixes across multiple files. Once the changes here are merged, #651 can be closed, as everything but the test_od changes are covered by other PRs.

The python-can send_message and send_periodic APIs expect bytes, not list[int]. Passing a list is accepted at runtime by some backends but is incorrect according to the type signatures and triggers Pylance and mypy errors. Using bytes literals and bytes() constructor makes the intent explicit and removes the type errors without any behavioral change.

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -6,6 +6,7 @@ from collections.abc import Iterator, MutableMapping
 from typing import Callable, Final, Optional, Union
 
 import can
+from can.typechecking import CanData
 
 from canopen.lss import LssMaster
 from canopen.nmt import NmtMaster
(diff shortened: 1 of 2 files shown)


acolomb:
Why don't we just fix our type annotations on the sending functions?  Added a commit to that effect.  With the proper typing in our network module, none of these fixes are necessary.  It aligns with the parameter docstring on `send_message()`, which already stated that "anything that can be converted to bytes" is acceptable.  Since this actually works at runtime, too, the annotation was simply too narrow.

`can.typechecking.CanData` already gives us exactly the right combination of accepted types ready for use, and future-proof.

PR #675, opened 2026-06-27
eds: Style cleanup and simplification, warning fixes

acolomb:
This should not result in any functional behavior changes.  It mainly fixes some of the clumsy code in eds.py and avoids type checker errors in the process.

See individual commit messages for more details on the changes.

### DIFF
--- a/canopen/objectdictionary/eds.py
+++ b/canopen/objectdictionary/eds.py
@@ -297,12 +297,12 @@ def build_variable(
     node_id: int,
     object_type: int,
     index: int,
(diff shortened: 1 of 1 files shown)


acolomb:
I have another PR branch prepared which is based on top of this one.  It adds a "strict" mode flag where EDS importing raises exceptions instead of just logging warnings.  See [this branch](https://github.com/acolomb/canopen/tree/eds-strict-mode), which will be submitted as PR once this cleanup is done.

PR #676, opened 2026-08-02
async pt 1: Add asyncio framework and improve callback system

sveinse:
This PR is the first PR (of many) that introduce asyncio support in canopen, see #272. The PRs are split into smaller, more manageable and reviewable chunks. See PR #359 for the entire asyncio feature, which this PR is an extract from.

The PR adds the basic framework for asyncio support in `Network`. It improves the CAN notification / callback system where it does error handling in a consistent way. After this PR, user provided callbacks may now be async coroutines.

This PR does not add any CAN functional asyncio support yet.

### Features

* Implement async support in `Network`.
* Using async context manager to start in async mode
* Improve error handling in notify / CAN callbacks
* Add `Network.FILTER_ERROR` to set if error in callbacks should be logged or raised
* Add `Network.create_task()` to run a coroutine under management. Multi-threaded.
* Add `Network.dispatch_callbacks()` that runs many callbacks, handling exceptions and coroutines.
* Use `network.dispatch_callbacks()` to run user provided callbacks, which can be coroutines
* Fixups on unit tests

### DIFF
--- a/canopen/network.py
+++ b/canopen/network.py
@@ -18,12 +18,13 @@ from canopen.objectdictionary.eds import import_from_node
 from canopen.sync import SyncProducer
 from canopen.timestamp import TimeProducer
 
-# Use backported TaskGroup for Python < 3.11, as asyncio.TaskGroup was added in 3.11
-if sys.version_info < (3, 11):
-    from taskgroup import TaskGroup
-else:
+# Use backported TaskGroup and ExceptionGroup for Python < 3.11
+if sys.version_info >= (3, 11):
+    from builtins import ExceptionGroup
     from asyncio import TaskGroup
-
+else:
+    from taskgroup import TaskGroup
+    from exceptiongroup import ExceptionGroup
 
 logger = logging.getLogger(__name__)
 
@@ -63,6 +64,9 @@ class Network(MutableMapping):
         #: network is closed.
         self.exit_stack: AsyncExitStack = AsyncExitStack()
         self.loop: Optional[asyncio.AbstractEventLoop] = None
+        #: A list of exceptions that have occurred in callbacks. This is used to
+        #: ensure that exceptions are not lost
+        self.exceptions: list[BaseException] = []
         self.sync = SyncProducer(self)
         self.time = TimeProducer(self)
         self.nmt = NmtMaster(0)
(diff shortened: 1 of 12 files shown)


test/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest.fixture(scope="session", autouse=True)
+def enable_network_exceptions():
+    """Fixture to enable exceptions in the reception threads.
+
+    This makes sure exceptions are not swallowed in the reception threads,
+    which is useful for debugging and testing.
+    """
+    canopen.Network.FILTER_ERRORS = False
+    yield
```
sveinse, line 15: I was uncertain if I could use pytest fixture to set a global value or to use `self.network.FILTER_ERROR = False` everywhere `Network` is used. Aka how its done for `self.network.NOTIFIER_SHUTDOWN_TIMEOUT = 0.0` throughout the tests.

I'm open to either alternative. Please advise.

canopen/network.py
```diff
@@ -374,7 +539,7 @@ def on_message_received(self, msg):
             self.network.notify(msg.arbitration_id, msg.data, msg.timestamp)
         except Exception as e:
             # Exceptions in any callbaks should not affect CAN processing
-            logger.error(str(e))
+            self.network.on_error(e)
 
     def stop(self) -> None:
         """Override abstract base method to release any resources."""
```
sveinse, line 545: Having this `class MessageListener` have grown on me. We have talked about improving the notification system, and I've considered many options, but I think I like what we have. Here is a little pondering over this topic.

We're should not run python-can in async mode. When running python-can with loop, all notifications (i.e. `on_message_receiced()`) runs in the loop thread, even regular callbacks. Our callbacks in the stack depend on locking primitives, so that not a good thing. Furthermore error handling in async callbacks is not working properly, where they don't get handled. See https://github.com/hardbyte/python-can/issues/1912. ==> Ergo, we have to manage the async and non-async callbacks ourselves, which is what this PR propose.

As for where to process the callbacks from python-can, I have looked at two alternatives:

1. Promote `Network` to inherit `can.Listener`. This way we don't need the `MessageListener.on_message_received() -> Network.notify()` trampoline. We then have to implement `Network.on_message_received()` directly. However, I kind of like the composition pattern with `self.notifier`.

`Network.notify()` is also elegant in that it's the single point delivery of incoming CAN messages. If one wants to adopt to another CAN access layer, its very simple to plug it into `Network` with very few methods to fix.

2. Make `class MessageListener` the full manager for all subscriptions. `on_message_received` will then be what `Network.notify()` does today. The register of subscribes will be kept in this class. Of course there will need to be a proxy in `Network` for the subscription API where one needs to get the pointer to the listener instance. So I don't really think this is worth it.

3. With the introduction of async, we how need to handle callbacks in either regular function and callbacks. This is what `Network.dispatch_callbacks()` takes care of. It also handles exceptions properly, both in regular and async functions and funnels everything via `Network.on_error()`.

`Network.on_error()` on its side will either raise the error or collect it depending on class settings. The collected exceptions is added to the `Network` class so that when `Network.check()` is run, it will raise all of them.

